PR #1, opened 2019-06-27
Add _mm_test_all_zeros

marktwtn:


jserv:
You shall add entries in file `tests/impl.cpp` as new instruction(s) inside the unit test.

marktwtn:
> You shall add entries in file `tests/impl.cpp` as new instruction(s) inside the unit test.

I have added the entries and the corresponding unit testing.
It passed on the **ARM** machine.

PR #3, opened 2019-07-02
Implement _mm_shuffle_epi8 with NEON instruction

marktwtn:
Add the corresponding unit testing.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1000,33 +1000,18 @@ FORCE_INLINE __m128i _mm_shuffle_epi_3332(__m128i a)
 
 // Shuffle packed 8-bit integers in a according to shuffle control mask in the
 // corresponding 8-bit element of b, and store the results in dst.
-FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i ia, __m128i ib)
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_epi8&expand=5146
+FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
 {
-    uint8_t *a = (uint8_t *) &ia;  // input a
-    uint8_t *b = (uint8_t *) &ib;  // input b
-    int32_t r[4];
+    int8x16_t tbl = vreinterpretq_s8_m128i(a);   // input a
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -1000,33 +1000,13 @@ FORCE_INLINE __m128i _mm_shuffle_epi_3332(__m128i a)
...
 // corresponding 8-bit element of b, and store the results in dst.
-FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i ia, __m128i ib)
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_epi8&expand=5146
+FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
 {
-    uint8_t *a = (uint8_t *) &ia;  // input a
-    uint8_t *b = (uint8_t *) &ib;  // input b
-    int32_t r[4];
+    int8x16_t t = vreinterpretq_s8_m128i(a);     // input a
+    uint8x16_t idx = vreinterpretq_u8_m128i(b);  // input b
```
jserv: Can you replace `t` and `idx` with something mnemonic?

jserv:
Please append corresponding test case.

sse2neon.h
marktwtn: I just follow the input argument naming by searching the `vqtbl1q_s8` in the [Neon intrinsic page](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics).

I think the `t` variable can be changed to `tbl`, or remains the same to follow the description in the Neon intrinsic page.

jserv: `tbl` sounds better since it is a matter of table lookup operations.

marktwtn:
I have added the corresponding unit testing and modified the implementation of `_mm_shuffle_epi8`.
The modification is based on the difference behaviour of `_mm_shuffle_epi8` and `vqtbl1q_s8`.

However, the unit testing is still not passing.
The pull request should not be merged until I fix the bug.

marktwtn:
The pull request can be merged without problem now.
The unit testing has been fixed.
I wrote the wrong unit testing since I forgot the difference of the signed/unsigned integer casting.

PR #6, opened 2019-09-10
Implement '_mm_aesenc_si128' support.

mjmacleod:
This pull request implements _mm_aesenc_si128 using the armv8 crypto extensions (where available) and a neon based vectorised aes round when crypto extensions are not available.

The vectorised neon version is based on an implementation by @ardbiesheuvel (which is used in the kernel for aegis and also talked about in several articles http://www.workofard.com/2017/01/accelerated-aes-for-the-arm64-linux-kernel/ http://www.workofard.com/2017/07/ghash-for-low-end-cores/

I've reproduced it with his permission and made slight modifications to make it fit in with sse2neon, along with implementing a basic test. I've tagged @ardbiesheuvel so that he can confirm that it is okay for it to enter this project under MIT license.


ardbiesheuvel:
Acked-by: Ard Biesheuvel <ard.biesheuvel@arm.com>

jserv:
@mjmacleod, Thanks for contribution! Can you please run `clang-format` in order to enforce consistent style?

jserv:
Thank @mjmacleod and @ardbiesheuvel for the great work!

PR #8, opened 2019-09-17
Implement _mm_sub_epi64 and _mm_mul_epu32

mjmacleod:
_mm_sub_epi64 and _mm_mul_epu32 equivalents with basic unit test

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1706,12 +1706,17 @@ FORCE_INLINE __m128 _mm_mul_ps(__m128 a, __m128 b)
 // Multiply the low unsigned 32-bit integers from each packed 64-bit element in
 // a and b, and store the unsigned 64-bit results in dst.
 //
-//   r0 :=  (int32_t*)a0 * (int32_t*)b0
-//   r1 :=  (int32_t*)a3 * (int32_t*)b3
+//   r0 :=  (uint32_t*)a0 * (uint32_t*)b0
+//   r1 :=  (uint32_t*)a3 * (uint32_t*)b3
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -1693,6 +1703,17 @@ FORCE_INLINE __m128 _mm_mul_ps(__m128 a, __m128 b)
...
 
+// Multiply the low unsigned 32-bit integers from each packed 64-bit element in
+// a and b, and store the unsigned 64-bit results in dst.
+//
+//   r0 :=  (int32_t*)a0 * (int32_t*)b0
+//   r1 :=  (int32_t*)a3 * (int32_t*)b3
+FORCE_INLINE __m128i _mm_mul_epu32(__m128i a, __m128i b)
+{
+    return vreinterpretq_m128i_u32(
+        vmulq_u32(vreinterpretq_u32_m128i(a), vreinterpretq_u32_m128i(b)));
```
marktwtn: I wonder whether this is the correct instruction since `vmulq_u32()` multiplies each element in the vector?

mjmacleod: You are right, it seems to be wrong, I've re-implemented it using a mask instead, though its not clear if my implementation is the best possible way to do it.

mjmacleod:
Sorry about that, I've pushed a new version that fixes the various issues with _mm_mul_epu32 and implements a mask to match _mm_mul_epu32 semantics.

It isn't clear to me if there might be a better neon instruction that can do this without needing to involve the extra loading of a mask which probably isn't very efficient...

mjmacleod:
Sorry, upon further inspection it seems something may still not be right, going to close pull request and will open again once I'm certain.

PR #11, opened 2019-11-06
Travis CI

marktwtn:
CI: Setup Travis CI for unit testing
test: Return -1 exit code when unit test fails

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: c++
+
+arch:
+  - arm64
+
+os:
+  - linux
+
+compiler:
+  - gcc
+
+before_install:
+  - sudo apt-get install -y build-essential
+
+script:
+  - make check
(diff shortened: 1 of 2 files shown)


marktwtn:
Currently we have some failed unit testing.
They will be fixed in the future.

.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+  - arm64
+
+os:
+  - linux
+
+compiler:
+  - gcc
+
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -y build-essential; fi
```
jserv: Is it necessary to validate Linux?

marktwtn: No, I'll remove it.

marktwtn: Done.

PR #14, opened 2019-11-12
fix: Fix single-precision floating-point comparison instructions

### LINKED ISSUE #12: Fix the unit test of lower single-precision floating-point comparison command
There are four instructions which cause the unit test fail.
- _mm_comilt_ss
- _mm_comile_ss
- _mm_comieq_ss
- _mm_comineq_ss

Base on the comment, I doubt that the failure happens when one of the operand is `NaN`.
[Comment](https://github.com/DLTcollab/sse2neon/blob/master/sse2neon.h#L2189):
```
// https://msdn.microsoft.com/en-us/library/2kwe606b(v=vs.90).aspx Important
// note!! The documentation on MSDN is incorrect!  If either of the values is a
// NAN the docs say you will get a one, but in fact, it will return a zero!!
```
[Comment](https://github.com/DLTcollab/sse2neon/blob/master/tests/impl.cpp#L1366):
```
// **NOTE** The documentation on MSDN is in error!  The actual
// hardware returns a 0, not a 1 if either of the values is a
// NAN!
```

However, I do not find the document other than `MSDN` which describes the returned value of the command when one of the operand is `NaN`.

marktwtn:
The result of comparing with NaN is called unordered.
It is different from equal, greater than or less than.

The commit returns the correct value of the single-precision
floating-point comparison instructions when the comparison result
is unordered.

Close #12.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -2195,10 +2195,10 @@ FORCE_INLINE int _mm_comilt_ss(__m128 a, __m128 b)
         vceqq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(a));
     uint32x4_t b_not_nan =
         vceqq_f32(vreinterpretq_f32_m128(b), vreinterpretq_f32_m128(b));
-    uint32x4_t a_or_b_nan = vmvnq_u32(vandq_u32(a_not_nan, b_not_nan));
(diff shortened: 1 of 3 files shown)


jserv:
The subject is misleading. It should be instruction or intrinsic, not command.

PR #17, opened 2020-01-30
Finish almost all integer intrinsics

### LINKED ISSUE #15: sse2neon.h:2655:22: error: argument to '__builtin_neon_vextq_v' must be a constant integer
I build apache/impala on aarch64, I download the file https://github.com/DLTcollab/sse2neon/blob/master/sse2neon.h under impala directory, and include it in some other files, an error raised:

/home/jenkins/workspace/impala/be/src/util/sse2neon.h:2655:22: error: argument to
      '__builtin_neon_vextq_v' must be a constant integer
    return (__m128i) vextq_s8((int8x16_t) a, (int8x16_t) b, c);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jenkins/workspace/native-toolchain/build/llvm-5.0.1-p2/lib/clang/5.0.1/include/arm_neon.h:5825:23: note:
      expanded from macro 'vextq_s8'
  __ret = (int8x16_t) __builtin_neon_vextq_v((int8x16_t)__s0, (int8x16_t)__s1, __p2, 32); \

I don't know how to fix this, the third argument c is const integer type, the value of c only can be known at run-time, right? But according to compile error, it seems that the value of c should be known at compile-time.  So I have no idea is it an error of the sse2neon.h? Or how to fix this?

easyaspi314:
Rewritten to a clean branch to make the commit log cleaner.

Doesn't have all the new tests, but I am working on an easier test suite which I will add in a later PR.

Improved:
 - Fixed a few typos
 - Fixed `_mm_alignr_si128` not working on Clang (fixes #15)
 - Change `__m128i` to `int64x2_t` to make GCC vector extension overloads compatible
 - Use `__builtin_shufflevector` on Clang when possible
 - `_mm_movemask_epi8`
 - `_mm_sign_*`
 - `_mm_mul_epu32`
 - ~~Better handling of out-of-range shifts, checks can be disabled with a macro if shifts are known to be in range~~
 - ~~Use variable shifts instead of immediate shifts due to Clang's stricter constant rules for `_mm_slli_epi*`, `_mm_srli_epi*`, and `_mm_srai_epi*`~~
 - Vectorised ARMv7 fallback for `_mm_shuffle_epi8`
 - Use `vzip1` and `vzip2` on aarch64 for `_mm_unpacklo_epi*` and `_mm_unpackhi_epi*` respectively

Implemented:
 - `_mm_sign_epi8`
 - ~~`_mm_sll_epi*`/`_mm_srl_epi*`/`_mm_sra_epi*`~~
 - ~~`_mm_sllv_epi*`/`_mm_srlv_epi*`/`_mm_srav_epi*`~~
 - `_mm_extract_epi*`, `_mm_insert_epi*`
 - `_mm_cvtep*_epi*`
 - `_mm_cvtsi64_si128`
 - `_mm_blend_epi16`
 - `_mm_blendv_epi8`
 - `_mm_cmpgt_epi64` (with v7 fallback)
 - `_mm_cmpeq_epi64` (with v7 fallback)
 - `_mm_set1_epi64x`
 - `_mm_lddqu_si128` (new)
 - `_mm_mul_epi32`
 - `_mm_hadd*`, `_mm_hsub*`
 - `_mm_mulhrs_epi16`
 - `_mm_maddubs_epi16`
 - `_mm_packus_epi32`
 - `_mm_abs_epi8`
 - `_mm_clmulepi64_si128` (with `vmull_p8` fallback) **(new!)** < Tests included
 -- clhash: 5.1 GB/s with native `vmull_p64`, 920 MB/s with fallback on Snapdragon 835

Unless I am missing some, this covers all 128-bit integer intrinsics up to and including SSE4.1, with some additions from SSE4.2 and AVX-128.

Edit: Shifts are delayed to a later PR, they have some really ugly details that are missing.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -3014,6 +3014,24 @@ FORCE_INLINE __m128i _mm_packs_epi32(__m128i a, __m128i b)
                      vqmovn_s32(vreinterpretq_s32_m128i(b))));
 }
 
+// Packs the 8 unsigned 32-bit integers from a and b into unsigned 16-bit integers
+// and saturates.
+//
+//   r0 := UnsignedSaturate(a0)
+//   r1 := UnsignedSaturate(a1)
+//   r2 := UnsignedSaturate(a2)
+//   r3 := UnsignedSaturate(a3)
+//   r4 := UnsignedSaturate(b0)
+//   r5 := UnsignedSaturate(b1)
+//   r6 := UnsignedSaturate(b2)
+//   r7 := UnsignedSaturate(b3)
+FORCE_INLINE __m128i _mm_packus_epi32(__m128i a, __m128i b)
+{
+    return vreinterpretq_m128i_u16(
+        vcombine_u16(vqmovn_u32(vreinterpretq_u32_m128i(a)),
+                     vqmovn_u32(vreinterpretq_u32_m128i(b))));
+}
+
 // Interleaves the lower 8 signed or unsigned 8-bit integers in a with the lower
 // 8 signed or unsigned 8-bit integers in b.
 //
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1982,6 +1982,68 @@ FORCE_INLINE __m128i _mm_madd_epi16(__m128i a, __m128i b)
     return vreinterpretq_m128i_s32(vcombine_s32(low_sum, high_sum));
 }
 
+// Multiply packed signed 16-bit integers in a and b, producing intermediate signed
+// 32-bit integers. Shift right by 15 bits while rounding up, and store the
+// packed 16-bit integers in dst.
+//
+//   r0 := Round(((int32_t)a0 * (int32_t)b0) >> 15)
+//   r1 := Round(((int32_t)a1 * (int32_t)b1) >> 15)
+//   r2 := Round(((int32_t)a2 * (int32_t)b2) >> 15)
+//   ...
+//   r7 := Round(((int32_t)a7 * (int32_t)b7) >> 15)
+FORCE_INLINE __m128i _mm_mulhrs_epi16(__m128i a, __m128i b)
+{
+    // Has issues due to saturation
+    // return vreinterpretq_m128i_s16(vqrdmulhq_s16(a, b));
+
+    // Multiply
+    int32x4_t mul_lo = vmull_s16(vget_low_s16(vreinterpretq_s16_m128i(a)),
+                                 vget_low_s16(vreinterpretq_s16_m128i(b)));
+    int32x4_t mul_hi = vmull_s16(vget_high_s16(vreinterpretq_s16_m128i(a)),
+                                 vget_high_s16(vreinterpretq_s16_m128i(b)));
+
+    // Rounding narrowing shift right
+    // narrow = (int16_t)((mul + 16384) >> 15);
+    int16x4_t narrow_lo = vrshrn_n_s32(mul_lo, 15);
+    int16x4_t narrow_hi = vrshrn_n_s32(mul_hi, 15);
+
+    // Join together
+    return vreinterpretq_m128i_s16(vcombine_s16(narrow_lo, narrow_hi));
+}
+
+// Vertically multiply each unsigned 8-bit integer from a with the corresponding
+// signed 8-bit integer from b, producing intermediate signed 16-bit integers.
+// Horizontally add adjacent pairs of intermediate signed 16-bit integers,
+// and pack the saturated results in dst.
+//
+//   FOR j := 0 to 7
+//      i := j*16
+//      dst[i+15:i] := Saturate_To_Int16( a[i+15:i+8]*b[i+15:i+8] + a[i+7:i]*b[i+7:i] )
+//   ENDFOR
+FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
+{
+    // This would be much simpler if x86 would choose to zero extend OR sign extend,
+    // not both.
+    // This could probably be optimized better.
+    uint16x8_t a = vreinterpretq_u16_m128i(_a);
+    int16x8_t b = vreinterpretq_s16_m128i(_b);
+
+    // Zero extend a
+    int16x8_t a_odd = vreinterpretq_s16_u16(vshrq_n_u16(a, 8));
+    int16x8_t a_even = vreinterpretq_s16_u16(vbicq_u16(a, vdupq_n_u16(0xff00)));
+
+    // Sign extend by shifting left then shifting right.
+    int16x8_t b_even = vshrq_n_s16(vshlq_n_s16(b, 8), 8);
+    int16x8_t b_odd = vshrq_n_s16(b, 8);
+
+    // multiply
+    int16x8_t prod1 = vmulq_s16(a_even, b_even);
+    int16x8_t prod2 = vmulq_s16(a_odd, b_odd);
+
+    // saturated add
+    return vreinterpretq_m128i_s16(vqaddq_s16(prod1, prod2));
+}
+
 // Computes the absolute difference of the 16 unsigned 8-bit integers from a
 // and the 16 unsigned 8-bit integers from b.
 //
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1654,28 +1654,72 @@ FORCE_INLINE __m128i _mm_adds_epu16(__m128i a, __m128i b)
         vqaddq_u16(vreinterpretq_u16_m128i(a), vreinterpretq_u16_m128i(b)));
 }
 
-FORCE_INLINE __m128i _mm_sign_epi32(__m128i a, __m128i b)
+// Negate packed 16-bit integers in a when the corresponding signed
+// 16-bit integer in b is negative, and store the results in dst.
+// Element in dst are zeroed out when the corresponding element
+// in b is zero.
+//
+//   for i in 0..7
+//     if b[i] < 0
+//       r[i] := -a[i]
+//     else if b[i] == 0
+//       r[i] := 0
+//     else
+//       r[i] := a[i]
+//     fi
+//   done
+FORCE_INLINE __m128i _mm_sign_epi16(__m128i _a, __m128i _b)
 {
-    int32x4_t zer0 = vdupq_n_s32(0);
-    int32x4_t ltMask =
-        vreinterpretq_s32_u32(vcltq_s32(vreinterpretq_s32_m128i(b), zer0));
-    int32x4_t gtMask =
-        vreinterpretq_s32_u32(vcgtq_s32(vreinterpretq_s32_m128i(b), zer0));
-    int32x4_t neg = vnegq_s32(vreinterpretq_s32_m128i(a));
-    int32x4_t tmp = vandq_s32(vreinterpretq_s32_m128i(a), gtMask);
-    return vreinterpretq_m128i_s32(vorrq_s32(tmp, vandq_s32(neg, ltMask)));
+    int16x8_t a = vreinterpretq_s16_m128i(_a);
+    int16x8_t b = vreinterpretq_s16_m128i(_b);
+
+    int16x8_t zero = vdupq_n_s16(0);
+    // signed shift right: faster than vclt
+    // (b < 0) ? 0xFFFF : 0
+    uint16x8_t ltMask = vreinterpretq_u16_s16(vshrq_n_s16(b, 15));
+    // (b == 0) ? 0xFFFF : 0
+    int16x8_t zeroMask = vreinterpretq_s16_u16(vceqq_s16(b, zero));
+    // -a
+    int16x8_t neg = vnegq_s16(a);
+    // bitwise select either a or neg based on ltMask
+    int16x8_t masked = vbslq_s16(ltMask, a, neg);
+    // res = masked & (~zeroMask)
+    int16x8_t res = vbicq_s16(masked, zeroMask);
+    return vreinterpretq_m128i_s16(res);
 }
 
-FORCE_INLINE __m128i _mm_sign_epi16(__m128i a, __m128i b)
+// Negate packed 32-bit integers in a when the corresponding signed
+// 32-bit integer in b is negative, and store the results in dst.
+// Element in dst are zeroed out when the corresponding element
+// in b is zero.
+//
+//   for i in 0..3
+//     if b[i] < 0
+//       r[i] := -a[i]
+//     else if b[i] == 0
+//       r[i] := 0
+//     else
+//       r[i] := a[i]
+//     fi
+//   done
+FORCE_INLINE __m128i _mm_sign_epi32(__m128i _a, __m128i _b)
 {
-    int16x8_t zer0 = vdupq_n_s16(0);
-    int16x8_t ltMask =
-        vreinterpretq_s16_u16(vcltq_s16(vreinterpretq_s16_m128i(b), zer0));
-    int16x8_t gtMask =
-        vreinterpretq_s16_u16(vcgtq_s16(vreinterpretq_s16_m128i(b), zer0));
-    int16x8_t neg = vnegq_s16(vreinterpretq_s16_m128i(a));
-    int16x8_t tmp = vandq_s16(vreinterpretq_s16_m128i(a), gtMask);
-    return vreinterpretq_m128i_s16(vorrq_s16(tmp, vandq_s16(neg, ltMask)));
+    int32x4_t a = vreinterpretq_s32_m128i(_a);
+    int32x4_t b = vreinterpretq_s32_m128i(_b);
+
+    int32x4_t zero = vdupq_n_s32(0);
+    // signed shift right: faster than vclt
+    // (b < 0) ? 0xFFFFFFFF : 0
+    uint32x4_t ltMask = vreinterpretq_u32_s32(vshrq_n_s32(b, 31));
+    // (b == 0) ? 0xFFFFFFFF : 0
+    int32x4_t zeroMask = vreinterpretq_s32_u32(vceqq_s32(b, zero));
+    // neg = -a
+    int32x4_t neg = vnegq_s32(a);
+    // bitwise select either a or neg based on ltMask
+    int32x4_t masked = vbslq_s32(ltMask, a, neg);
+    // res = masked & (~zeroMask)
+    int32x4_t res = vbicq_s32(masked, zeroMask);
+    return vreinterpretq_m128i_s32(res);
 }
 
 // Adds the four single-precision, floating-point values of a and b.
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -706,6 +706,11 @@ FORCE_INLINE __m128i _mm_abs_epi16(__m128i a)
     return vreinterpretq_m128i_s16(vabsq_s16(vreinterpretq_s16_m128i(a)));
 }
 
+FORCE_INLINE __m128i _mm_abs_epi8(__m128i a)
+{
+    return vreinterpretq_m128i_s8(vabsq_s8(vreinterpretq_s8_m128i(a)));
+}
+
 // Takes the upper 64 bits of a and places it in the low end of the result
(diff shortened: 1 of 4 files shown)


sse2neon.h
```diff
@@ -2650,19 +3427,35 @@ FORCE_INLINE __m128i _mm_unpackhi_epi64(__m128i a, __m128i b)
...
+#define _mm_alignr_epi8(a, b, c) ((__m128i) vextq_s8((int8x16_t) (b), (int8x16_t) (a), (c)))
+
+// Extracts the selected signed or unsigned 8-bit integer from a and zero
+// extends.
+// FORCE_INLINE int _mm_extract_epi8(__m128i a, __constrange(0,16) int imm)
+#define _mm_extract_epi8(a, imm) \
+    vgetq_lane_u8(vreinterpretq_u8_m128i(a), (imm))
+
+// Inserts the least significant 8 bits of b into the selected 16-bit integer
+// of a.
```
marktwtn: Replace `16-bit` with `8-bit`.

sse2neon.h
```diff
@@ -1405,38 +1763,34 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int count)
...
-    hi = vpadd_u8(hi, hi);
-    hi = vpadd_u8(hi, hi);
-    hi = vpadd_u8(hi, hi);
-
-    return ((hi[0] << 8) | (lo[0] & 0xFF));
+    // Shift out everything but the sign bits
+    // 01 01 00 01 00 00 01 00
+    uint16x8_t high_bits = vreinterpretq_u16_u8(vshrq_n_u8(input, 7));
+    // Merge the even lanes together. The 'xx' bits are garbage.
+    // xx03 xx02 xx00 xx01
```
marktwtn: I thought it should be `xx03 xx01 xx00 xx02`?

easyaspi314: You wanted them not to be printed in little endian order, so I reversed them. Now you are more confused.

Maybe I should use this diagram:
```
89 ff 1d c0 00 10 99 33
 \  \  \  \  \  \  \  \
01 01 00 01 00 00 01 00
 |_/   |_/   |_/   |_/
/|    /|    /|    /|
\|    \|    \|    \|
03 01 02 01 00 00 01 00
03 xx 02 xx 00 xx 01 xx
 |____/      |_____/
/|          /|
\|          \|
0b 05 02 01 04 00 01 00
0b xx xx xx 04 xx xx xx
 |__________/
/|
||
4b 05 12 01 04 00 01 00
4b xx xx xx xx xx xx xx
4b
```
This shows the actual byte values of the example vector and how the shifts work in little endian. However, while it would be backwards, the algorithm is easy to see in big endian:
```
89 ff 1d c0 00 10 99 33
 \  \  \  \  \  \  \  \
01 01 00 01 00 00 01 00
  \ |   \ |   \ |   \ |
   \|    \|    \|    \|
xx 03 xx 01 xx 00 xx 02
     \___ |     \____ |
         \|      \   \|
xx xx xx 0d xx xx xx 02
           \_________ |
                     ||
xx xx xx xx xx xx xx d2
d2
```

I should use big endian as it is clear, but make a note that it is actually done in little endian.

easyaspi314:
@marktwtn how is the documentation now?

marktwtn:
> @marktwtn how is the documentation now?

It looks better now.
I think the big-endian is good for visualization and is more intuitive to illustrate the logic.

sse2neon.h
```diff
@@ -1768,28 +1768,77 @@ FORCE_INLINE __m128i _mm_sra_epi64(__m128i _a, __m128i _count)
...
+    // Last, an even wider 64-bit shift + add to get our result in the low 8 bit lanes.
+    // xx xx xx 0d xx xx xx 02
+    //            \_________ |   paired64 = (uint8x8_t)(paired32 + (paired32 >> 28))
+    //                      \|
+    // xx xx xx xx xx xx xx d2
+    //
+    // 00001101 00000010 (0d 02)
+    //     \   \___ |  |
+    //      '---.  \|  |
+    // xxxxxxxx 11011101 (xx d2)
```
marktwtn: Typo: `11010010`.

sse2neon.h
```diff
@@ -1768,28 +1768,77 @@ FORCE_INLINE __m128i _mm_sra_epi64(__m128i _a, __m128i _count)
...
+    uint8x16_t paired64 = vreinterpretq_u8_u64(vsraq_n_u64(paired32, paired32, 28));
+
+    // Extract the low 8 bits from each 64-bit lane with 2 8-bit extracts.
+    // xx xx xx xx xx xx xx d2
+    //                      ||  return paired64[0]
+    //                      d2
+    // Note: Little endian would return the correct value 4b (10111011) instead.
```
marktwtn: Ditto.

marktwtn:
Include the commit e42076a78132afa53f1999d1b6048a9741a6d8be into aa4905f2120f7dc9b1d10eb1d32a231eb37c4fde.

easyaspi314:
How? I am sorry, but I am ***seriously terrible*** at git and don't understand what you are trying to get me to do.

Git is like Japanese to me. I know a few important phrases/templates/vocab, but I can't hold or understand a casual conversation.

Since I mainly contribute to projects, I have never really needed to learn anything besides the basic forms of `clone`, `push`, `pull`, `checkout`, `commit`, `reset`, `diff`, `add`, and `rm`. 

I attempted to fix it earlier when I noticed the mistake, but that just made it so I couldn't push anymore and I had to start with another copy.

jserv:
> How? I am sorry, but I am _**seriously terrible**_ at git and don't understand what you are trying to get me to do.

Alternatively, @marktwtn can merge and rework the changes done by @easyaspi314 in a separate branch, and then we can perform final merge process after reviewing.

marktwtn:
> Alternatively, @marktwtn can merge and rework the changes done by @easyaspi314 in a separate branch, and then we can perform final merge process after reviewing.

I contacted with @easyaspi314 and gave him/her some resource for learning git.
It seems that it is easier than @easyaspi314 expected.

We will keep in touch to finish the pull request.
If it is still too hard to use git, I will follow the advice of @jserv .

marktwtn:
Rewrite the commit header 0828b479f25ff6ff54a0643de9e857ddb1c0474c as `Add Devin Hussey as a contributor` and remove the commit message.
Include the commit 578450ed392b6c52f5f0e2db17f0bc6fc8a80744 into f3acaf4a57412f64f98239a8e8d289b8b281ad16.
Include the commit 9eb12994cc777c2249a3ef4b56e16152dbd04451 into a826c6ac4a8063f1e5cc8caf6a0eb68fe8aee134.

easyaspi314:
@marktwtn Is that better?

easyaspi314:
Thanks for the help by the way.

sse2neon.h
```diff
@@ -1264,164 +1264,418 @@ FORCE_INLINE __m128i _mm_shuffle_epi32_default(__m128i a,
...
+
+// Shift packed 16-bit integers in a left by the amount specified by
+// the corresponding element in count while shifting in zeros, and
+// store the results in dst.
+FORCE_INLINE __m128i _mm_sllv_epi16(__m128i _a, __m128i _count)
+{
+    uint16x8_t a = vreinterpretq_u16_m128i(_a);
+    int16x8_t count = _mm_sllv_to_vshl_s16(_count);
+    return vreinterpretq_m128i_u16(vshlq_u16(a, count));
+}
```
marktwtn: I think it still has potential behaviour mismatching when one of the element of `_count` is `0bxxxxxxx1 00000001`.
And **vqabsq_s16(count)** seems not helping in this case?

marktwtn:
> @marktwtn Is that better?

Much better now. Thanks.

There still has some minor issues and I am still reviewing the pull request.

Typo of the commit message body dc642db7a8cf6114c4efae3d2dc361f0e7cbbbed: `Clamg` to `Clang`.

sse2neon.h
easyaspi314: Oh, yeah, if `0 > amt > -16`, that would cause a problem. Silly me.

This should work, right?
```c
FORCE_INLINE __m128i _mm_sllv_epi16(__m128i val, __m128i count)
{
    // count < 0 ? 0xffff : 0x0000
    int16x8_t mask = vshrq_n_s16(count, 15);
    // count > 0 ? (val << count) : (val >> count) 
    uint16x8_t shifted = vshlq_u16(val, count);
    // Mask out any negative shifts
    // shifted & (~mask)
    return vbicq_u16(shifted, mask);
}

FORCE_INLINE __m128i _mm_srlv_epi16(__m128i val, __m128i count)
{
    // count < 0 ? 0xffff : 0x0000
    int16x8_t mask = vshrq_n_s16(count, 15);
    // negative = shift right
    int16x8_t count_neg = vnegq_s16(count);
    // count_neg > 0 ? (val << count_neg) : (val >> count_neg) 
    uint16x8_t shifted = vshlq_u16(val, count_neg);
    // Mask out any negative shifts
    // shifted & (~mask)
    return vbicq_u16(shifted, mask);
}


FORCE_INLINE __m128i _mm_srav_epi16(__m128i val, __m128i count)
{
    // count < 0 ? 0xffff : 0x0000
    int16x8_t mask = vshrq_n_s16(count, 15);
    // negative = shift right
    int16x8_t countNeg = vnegq_s16(count);
    // count_neg > 0 ? (val << count_neg) : (val >> count_neg) 
    int16x8_t shifted = vshlq_s16(val, count_neg);
    // For large shifts, just extend the sign bit
    int16x8_t sign_mask = vshrq_n_s16(val, 15);
    // Either use the sign mask or the actual value, depending
    // on whether the shift was greater than zero
    return vbslq_s16(mask, sign_mask, shifted);
}
```

easyaspi314:
Just tell me when you are ready for me to change it, as I don't want to have to keep force pushing for every single change.

sse2neon.h
easyaspi314: And for the single ones, this seems to be pretty good:
```c
__m128i _mm_sll_epi16(__m128i val, __m128i count)
{
    // get the important part of count
    int16x8_t shifter = vdupq_lane_s16(vreinterpret_s16_u64(vget_low_u64(count)), 0);
    // Shift left with the shifter
    uint16x8_t shift = vshlq_u16(val, shifter);                                                      
    // Check if count[0] > 15
    // We use a shift to avoid moving immediates from scalar.
#ifdef __aarch64__
    // (count >> 4) == 0
    uint64x1_t shifted = vshr_n_u64(vget_low_u64(count), 4);
    uint16x8_t mask = vreinterpretq_u16_u64(vdupq_lane_u64(vceqz_u64(shifted), 0));
#else
    // no compare for uint64x1_t, improvise
    // narrow = saturate_u32(count >> 4)
    // any large values will saturate to 0xffffffff
    uint32x2_t narrow = vqshrn_n_u64(count, 4);
    // check if zero
    uint16x8_t mask = vreinterpretq_u16_u32(vdupq_lane_u32(vceq_u32(narrow, vdup_n_u32(0)), 0));
#endif
    // AND out the garbage
    return vandq_u16(shift, mask);
}
```

marktwtn: Take **_mm_sllv_epi16** as an example.

```c
    // count < 0 ? 0xffff : 0x0000
    int16x8_t mask = vshrq_n_s16(count, 15);
```

I think the result is `// count < 0 ? 0x0001 : 0x0000` since it shift right 15 bits?

```c
    // count > 0 ? (val << count) : (val >> count) 
    uint16x8_t shifted = vshlq_u16(val, count);
    // Mask out any negative shifts
    // shifted & (~mask)
    return vbicq_u16(shifted, mask);
```

The mask is not reversed in `return vbicq_u16(shifted, mask)`.

It still not handles the situation I mentioned before.
For example, what if the shift value is `0x0101` = `257`?
According to [_mm_sllv_epi16](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_sllv_epi16&expand=5326), it would shift left `257` bits.
However, the **vshlq_u16** would shift left only `1` bit since
```
This instruction takes each element in the vector of the first source SIMD&FP register,
shifts each element by a value from the least significant byte
of the corresponding element of the second source SIMD&FP register,
```

easyaspi314: It is a *signed* shift right. I'm just going to do vclt instead (aside from s64 because v7).

marktwtn: You are right, I did not notice that it is `signed shift right`.

easyaspi314: Oh, I see what you mean by the 257 thing…

It reads an int8_t, not the entire integer.

Time for yet another iteration.

easyaspi314: Okay, I think I can just do this
```c
FORCE_INLINE __m128i _mm_sllv_epi32(__m128i a, __m128i amt)
{
      uint32x4_t mask = vcltq_u32((uint32x4_t)amt, vdupq_n_u32(32));
      uint32x4_t shifted = vshlq_u32((uint32x4_t)a, (uint32x4_t)amt);
      return (__m128i)vandq_u32(shifted, mask);
}
```

marktwtn: It seems work well.
Please modify the related intrinsic functions as well.

easyaspi314:
Oof, just realized that I had used `abort()` without `<stdlib.h>`...

Should have a syntax test which simply compiles it as a C99 file.

tests/impl.cpp
```diff
@@ -2130,6 +2134,82 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+    // Karatsuba multiply
+    uint32_t a_lo = a & 0xffffffff;
+    uint32_t a_hi = a >> 32;
+    uint32_t b_lo = b & 0xffffffff;
+    uint32_t b_hi = b >> 32;
+
+    // a_pair = a_lo + b_lo;
+    uint32_t a_pair = a_lo ^ a_hi;
+    // b_pair = b_lo + b_lo;
+    uint32_t b_pair = b_lo ^ b_hi;
```
marktwtn: The comment and the source code is mismatching.

marktwtn: Why using **XOR** instead of normal addition?

tests/impl.cpp
```diff
@@ -2130,6 +2134,82 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+    // do half multiplies
+    uint64_t lower = clmul_32(a_lo, b_lo);
+    uint64_t upper = clmul_32(a_hi, b_hi);
+    uint64_t pair  = clmul_32(a_pair, b_pair);
+    // pair = pair - lower - upper;
+    pair ^= lower;
+    pair ^= upper;
```
marktwtn: Why using **XOR** instead of normal subtraction?

tests/impl.cpp
```diff
@@ -2130,6 +2134,82 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+    // ret += pair << 32;
+    lower ^= pair << 32;
+    upper ^= pair >> 32;
```
marktwtn: The comment is quite confusing compared to the source code.

tests/impl.cpp
```diff
@@ -2130,6 +2134,82 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
     return result == ret ? true : false;
 }
 
+static inline uint64_t MUL(uint32_t a, uint32_t b)
+{
+    return (uint64_t)a * (uint64_t)b;
+}
+
+// From BearSSL. Performs a 32-bit->64-bit polynomial multiply.
```
marktwtn: Can you update the link of the BearSSL algorithm resource?

tests/impl.cpp
```diff
@@ -2130,6 +2134,82 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+    z2 &= (uint64_t)0x4444444444444444;
+    z3 &= (uint64_t)0x8888888888888888;
+    return z0 | z1 | z2 | z3;
+}
+
+// 64x64->128-bit carryless multiply.
+static std::pair<uint64_t, uint64_t>
+clmul_64(uint64_t a, uint64_t b)
+{
+    // Karatsuba multiply
```
marktwtn: Add the link of `Karatsuba multiply`.
For example: https://en.wikipedia.org/wiki/Karatsuba_algorithm
Or the resource you referenced.

tests/impl.cpp
easyaspi314: It is polynomial (carry-less) arithmetic; add and subtract are both xor.

I should clarify that.

marktwtn: Please clarify that.
I don't even know it is a carry-less arithmetic.

easyaspi314: Well that is actually a typo, so I should fix it nonetheless.

easyaspi314:
Sorry about the quietness, I'm just a little busy on another project and having trouble making a clean shift implementation.

marktwtn:
> Sorry about the quietness, I'm just a little busy on another project and having trouble making a clean shift implementation.

If you still have problems, we can drop the shift intrinsic function commit and add it in the future pull request.
We've already had too many commits in this pull request.

easyaspi314:
I have the implementation, but I hate the copy-paste, and got myself lost in macromania.

I agree, let's work on the shifts later.

easyaspi314:
Ok, redocumented it and made it match the Wikipedia article, also fixed the "Clamg" typo.

easyaspi314:
The changes to shifts have also been reverted for now so I can work on trying to make it cleaner without holding back the other changes.

tests/impl.cpp
```diff
@@ -2130,6 +2134,115 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+
+// Performs a 64x64->128-bit carryless/polynomial long
+// multiply, using the above routine to calculate the
+// subproducts needed for the full-size multiply.
+//
+// This uses the Karatsuba algorithm.
+//
+// Normally, the Karatsuba algorithm isn't beneficial
+// until very large numbers due to carry tracking and
+// multiplication being relatively cheap.
```
marktwtn: `until very large number` is weird.
Change `cheap` to `low cost` and refactor the comment.

tests/impl.cpp
```diff
@@ -2130,6 +2134,115 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+//
+// However, we have no carries and multiplication is
+// definitely not cheap, so the Karatsuba algorithm is
+// a free and easy optimization.
```
marktwtn: Change `cheap` to `low cost`.
What does `free optimization` mean?

tests/impl.cpp
```diff
@@ -2130,6 +2134,115 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+//
+// https://en.m.wikipedia.org/wiki/Karatsuba_algorithm
+//
+// Note that addition and subtraction are both
+// performed with xor, since all operations are
+// carryless.
+//
+// The comments represent the actual mathematical
+// operations being performed and to mirror the linked
```
marktwtn: `mirror to`.

jserv:
Let's get this pull request ready for merge.

easyaspi314:
Sorry, I have been pretty busy/distracted on another project.

I forgot what I was doing halfway through a rebase, will try to fix everything up by tomorrow.

easyaspi314:
Oh, I was almost done.

Skipping shifts for now, as don't want to delay this any longer.

marktwtn:
I think we can merge the pull request.

easyaspi314:
Yeah, let's get it over with so I can relax a bit. 👍

jserv:
Thank @easyaspi314 for the great contributions!

easyaspi314:
No problem. Hopefully I can get my "shift" together and finish that up soon.

Sorry it took so long, I can't manage two projects at the same time, and I was busy preparing for the latest xxHash release. ¯\\_(ツ)_/¯

PR #21, opened 2020-04-14
run make check on x86

sebpop:
Make it possible to run the testsuite on both aarch64 and on x86.  When running
the tests on x86, each test executes native sse2 instructions and verifies that
the check part of the tests is correct.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,12 @@
 CXX = g++
-ARCH_CFLAGS = -march=armv8-a+fp+simd -mtune=thunderx
+processor := $(shell uname -p)
+ifeq ($(processor),aarch64)
+  ARCH_CFLAGS = -march=armv8-a+fp+simd -mtune=thunderx
+else ifeq ($(processor),$(filter $(processor),i386 x86_64))
+  ARCH_CFLAGS = -maes -mpclmul -mssse3 -msse4.2
+else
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,5 +1,15 @@
-CXX = g++
-ARCH_CFLAGS = -march=armv8-a+fp+simd -mtune=thunderx
```
jserv, line 2: It was my bad to leave `thunderx` in generic Makefile. Thanks for fixing.

tests/impl.cpp
```diff
@@ -11,7 +11,40 @@
 // This program a set of unit tests to ensure that each SSE call provide the
 // output we expect.  If this fires an assert, then something didn't match up.
 
+#ifdef __aarch64__
```
jserv, line 14: We should take both Arm and Aarch64 into considerations.

sebpop: Good point.  I will update this patch.  Thanks for your review!

sebpop: I tried to enable arm32 builds with TravisCI as well, however arm32 is not yet supported:
https://docs.travis-ci.com/user/multi-cpu-architectures
https://travis-ci.community/t/how-to-deploy-project-with-multiple-architectures-how-well-does-arm32-work/7679

PR #23, opened 2020-04-16
implement mm_sll_epi16

sebpop:
The patch passes `make check` on both aarch64 and x86_64.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1529,6 +1529,24 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int count)
         ret;                                                            \
     })
 
+// Shifts the 8 signed or unsigned 16-bit integers in a left by count bits while
+// shifting in zeros.
+//
+//   r0 := a0 << count
+//   r1 := a1 << count
+//   ...
+//   r7 := a7 << count
+//
+// https://msdn.microsoft.com/en-us/library/c79w388h(v%3dvs.90).aspx
+FORCE_INLINE __m128i _mm_sll_epi16(__m128i a, __m128i count)
+{
+    uint64_t c = ((SIMDVec *) &count)->m128_u64[0];
+    if (c > 15)
+        return _mm_setzero_si128();
+
+    int16x8_t vc = vdupq_n_s16((int16_t) c);
+    return vreinterpretq_m128i_s16(vshlq_s16(vreinterpretq_s16_m128i(a), vc));
+}
 
 // NEON does not provide a version of this function.
 // Creates a 16-bit mask from the most significant bits of the 16 signed or
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -424,6 +424,9 @@ const char *SSE2NEONTest::getInstructionTestString(InstructionTest test)
     case IT_MM_SLLI_EPI16:
         ret = "MM_SLLI_EPI16";
         break;
+    case IT_MM_SLL_EPI16:
+        ret = "IT_MM_SLL_EPI16";
+        break;
     case IT_MM_SRLI_EPI16:
         ret = "MM_SRLI_EPI16";
         break;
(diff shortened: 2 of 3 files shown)


sse2neon.h
```diff
@@ -1529,6 +1529,21 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int count)
...
+//   ...
+//   r7 := a7 << count
+//
+// https://msdn.microsoft.com/en-us/library/c79w388h(v%3dvs.90).aspx
+FORCE_INLINE __m128i _mm_sll_epi16(__m128i a, __m128i count)
+{
+    int64_t c = ((SIMDVec *) &count)->m128_i64[0];
+    int16x8_t vc = vdupq_n_s16(c);
+    return vreinterpretq_m128i_s16(vshlq_s16(vreinterpretq_s16_m128i(a), vc));
+}
```
marktwtn, line 1546: Be aware that when `c` equals to `256`, the element value of `vc` vector is `1`.
Hence the returned vector would be shifted `1` bit instead of being set to `0`.

We might need a mask `0xFFF...F00` to check whether the count is too large or too small.
If it is, set all the element to 0.
Otherwise, shift as usual.

sse2neon.h
```diff
@@ -1529,6 +1529,23 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int count)
...
+//   r1 := a1 << count
+//   ...
+//   r7 := a7 << count
+//
+// https://msdn.microsoft.com/en-us/library/c79w388h(v%3dvs.90).aspx
+FORCE_INLINE __m128i _mm_sll_epi16(__m128i a, __m128i count)
+{
+    uint64_t c = ((SIMDVec *) &count)->m128_u64[0];
+    if (c > 15)
+      return _mm_setzero_si128();
```
jserv, line 1545: Use 4 spaces for indention.

tests/impl.cpp
```diff
@@ -2773,6 +2795,9 @@ class SSE2NEONTestImpl : public SSE2NEONTest
         case IT_MM_SLLI_EPI16:
             ret = test_mm_slli_epi16((const int16_t *) mTestIntPointer1);
             break;
+        case IT_MM_SLL_EPI16:
+            ret = test_mm_sll_epi16((const int16_t *) mTestIntPointer1, i % 16);
```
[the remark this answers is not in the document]
sebpop: Fixed.

sse2neon.h
sebpop: fixed

sse2neon.h
sebpop: I think the new patch fixes this problem.

sse2neon.h
```diff
@@ -1529,6 +1529,21 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int count)
...
+//   r0 := a0 << count
+//   r1 := a1 << count
+//   ...
+//   r7 := a7 << count
+//
+// https://msdn.microsoft.com/en-us/library/c79w388h(v%3dvs.90).aspx
+FORCE_INLINE __m128i _mm_sll_epi16(__m128i a, __m128i count)
+{
+    int64_t c = ((SIMDVec *) &count)->m128_i64[0];
+    int16x8_t vc = vdupq_n_s16(c);
```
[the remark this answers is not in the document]
sebpop: fixed.

tests/impl.cpp
```diff
@@ -1678,6 +1681,25 @@ bool test_mm_slli_epi16(const int16_t *_a)
     return validateInt16(c, d0, d1, d2, d3, d4, d5, d6, d7);
 }
 
+bool test_mm_sll_epi16(const int16_t *_a, const int count)
```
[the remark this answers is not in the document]
sebpop: fixed.

PR #25, opened 2020-04-16
Implement _mm_malloc and _mm_free

jserv:
_mm_malloc is a wrapper to posix_memalign(), which is specified in
POSIX.1-2001 and POSIX.1-2008.


tests/impl.cpp
```diff
@@ -2381,6 +2384,19 @@ bool test_mm_aesenc_si128(const int32_t *a, const int32_t *b)
     return validate128(resultReference, resultIntrinsic);
 }
 
+bool test_mm_malloc(const size_t *a, const size_t *b)
+{
+    size_t size = *a % (1024 * 16) + 1;
```
marktwtn, line 2389: How do we determine the limit of `1024 * 16`?

PR #29, opened 2020-04-17
implement _mm_cmplt_epi16

sebpop:


tests/impl.h
```diff
@@ -134,6 +134,7 @@ enum InstructionTest {
     IT_MM_ADDS_EPI16,
     IT_MM_MAX_EPI16,
     IT_MM_SUBS_EPU16,
+    IT_MM_CMPLT_EPI16,
```
marktwtn, line 137: At the comment `// Unit test implemented and verified as fully working` at the end.

tests/impl.cpp
```diff
@@ -2012,6 +2015,24 @@ bool test_mm_subs_epu16(const int16_t *_a, const int16_t *_b)
...
+    uint16_t d4 = _a[4] < _b[4] ? 0xffff : 0;
+    uint16_t d5 = _a[5] < _b[5] ? 0xffff : 0;
+    uint16_t d6 = _a[6] < _b[6] ? 0xffff : 0;
+    uint16_t d7 = _a[7] < _b[7] ? 0xffff : 0;
+
+    __m128i a = test_mm_load_ps((const int32_t *) _a);
+    __m128i b = test_mm_load_ps((const int32_t *) _b);
+    __m128i c = _mm_cmplt_epi16(a, b);
+
+    return validateInt16(c, d0, d1, d2, d3, d4, d5, d6, d7);
```
marktwtn, line 2033: Change the `validateInt16()` to `validateUInt16()`.

tests/impl.h
sebpop: done

tests/impl.cpp
sebpop: done

PR #32, opened 2020-05-10
Implement _mm_set_ss

jserv:


tests/impl.cpp
```diff
@@ -764,6 +767,12 @@ bool test_mm_set_ps(float x, float y, float z, float w)
     return validateFloat(a, x, y, z, w);
 }
 
+bool test_mm_set_ss(float a)
+{
+    __m128 c = _mm_set_ss(a);
+    return validateFloatEpsilon(c, 0, 0, 0, a, 1e-8);
```
marktwtn, line 773: Why not using `validateFloat()` instead?

jserv: SSE is fully specified. There are actually a few exceptions for non-IEEE-754 operations like `rsqrtss`, where Intel never fully specified the behavior. I am not sure if we might break binary compatibility or not regarding floating validation.

marktwtn: Since `_mm_set_ss` is quite simple and fully specified without any approximation, I think using `validateFloat()` would be suitable?

PR #39, opened 2020-06-21
feat: Implement _mm_load_sd

### LINKED ISSUE #35: Implement _mm_load_sd
`_mm_load_ss` has beed implemented, what about the  `_mm_load_sd`?

Synopsis
```c
#include <emmintrin.h>
__m128d _mm_load_sd (double const* mem_addr)
```

Instruction: movsd xmm, m64
CPUID Flags: SSE2
Description
> Load a double-precision (64-bit) floating-point element from memory into the lower of dst, and zero the upper element. mem_addr does not need to be aligned on any particular boundary.

Operation
> dst[63:0] := MEM[mem_addr+63:mem_addr]
> dst[127:64] := 0

marktwtn:
The intrinsic conversion is slightly different on ARM 32-bit
architecture since it does not support float64x2_t data type.
The __m128d data type is treated as float32x4_t and the ARM 64-bit
intrinsics are avoided using on ARM 32-bit architecture.

Close #35.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -78,6 +78,14 @@
 
 typedef float32x2_t __m64;
 typedef float32x4_t __m128;
+// On ARM 32-bit architecture, the float64x2_t is not supported.
+// The data type __m128d should be repesented in a different way for related
+// intrinsic conversion.
+#if defined(__aarch64__)
+typedef float64x2_t __m128d;
+#else
+typedef float32x4_t __m128d;
+#endif
 typedef int64x1_t __m64i;
 typedef int64x2_t __m128i;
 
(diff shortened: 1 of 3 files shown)


tests/impl.cpp
```diff
@@ -723,6 +726,17 @@ bool validateSingleFloatPair(float a, float b)
                          // into account as well.
 }
 
+bool validateSingleDoublePair(double a, double b)
+{
+    const uint64_t *ia = (const uint64_t *) &a;
+    const uint64_t *ib = (const uint64_t *) &b;
+    return (*ia) == (*ib)
```
jserv, line 733: It is not necessary to place ternary operator (`?`) here.

sse2neon.h
```diff
@@ -78,6 +78,11 @@
 
 typedef float32x2_t __m64;
 typedef float32x4_t __m128;
+#if defined(__aarch64__)
```
jserv, line 81: Explain the typedefs here.

sse2neon.h
```diff
@@ -575,6 +580,21 @@ FORCE_INLINE __m128 _mm_loadu_ps(const float *p)
...
+// Loads a double-precision, floating-point value.
+// The upper double-precision, floating-point is set to zero. The address p does
+// not need to be 16-byte aligned.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/574w9fdd(v%3dvs.100)
+FORCE_INLINE __m128d _mm_load_sd(const double *p)
+{
+#if defined(__aarch64__)
+    return vsetq_lane_f64(*p, vdupq_n_f64(0), 0);
+#else
+    const float *fp = (const float *)p;
```
jserv, line 592: Use `clang-format` to ensure consistent style.

tests/impl.cpp
```diff
@@ -723,6 +726,17 @@ bool validateSingleFloatPair(float a, float b)
                          // into account as well.
 }
 
+bool validateSingleDoublePair(double a, double b)
```
jserv, line 729: I am not sure if this function is worthy as a standalone helper function.

marktwtn: As in the future we might integrate more intrinsics related to double value validation?

jserv: `validateDouble` can be generic enough to satisfy validation needs.

tests/impl.cpp
marktwtn: Removed.

sse2neon.h
marktwtn: Add the related comment.

sse2neon.h

tests/impl.cpp
marktwtn: I copied the original design of `validateFloat()` and `validateSingleFloatPair()`.
Maybe they should be integrated into a single function?
Like integrate `validateFloat()` and `validateSingleFloatPair()` into `validateFloat()`?

PR #41, opened 2020-06-24
Rewrite __builtin_shufflevector check with __has_builtin macro

jserv:
Both clang and gcc-10 support "__has_builtin", which evaluates the
existence of specified builtin function. __builtin_shufflevector might
be implemented in newer gcc releases, and this patch attempts to take
a flexible approach to detect its presence.

More about "__has_builtin":
  https://clang.llvm.org/docs/LanguageExtensions.html
GCC 10 provides "__hash_builtin" preprocessor operator:
  https://gcc.gnu.org/gcc-10/changes.html

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -62,6 +62,13 @@
 
 #include <arm_neon.h>
 
+/* "__has_builtin" can be used to query support for built-in functions
+ * provided by gcc/clang and other compilers that support it.
+ */
+#ifndef __has_builtin
+#define __has_builtin(x) 0  // Compatibility with non-{clang,gcc-10} compilers
+#endif
+
 /**
  * MACRO for shuffle parameter for _mm_shuffle_ps().
  * Argument fp3 is a digit[0123] that represents the fp from argument "b"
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -1045,7 +1052,7 @@ FORCE_INLINE __m128 _mm_shuffle_ps_default(__m128 a,
         }                                                  \
         ret;                                               \
     })
-#endif  // not clang
+#endif  // no __builtin_shufflevector
```
marktwtn, line 1055: I think the comment `no __builtin_shufflevector` should be removed since the `#endif` just means that it is the end of the `#if` and `#else` combination.

sse2neon.h
```diff
@@ -1270,7 +1277,7 @@ FORCE_INLINE __m128i _mm_shuffle_epi32_default(__m128i a,
         }                                                \
         ret;                                             \
     })
-#endif  // not clang
+#endif  // no __builtin_shufflevector
```
marktwtn, line 1280: Ditto.

PR #46, opened 2020-06-26
Add Aarch64 fastpath for _mm_movemask_ps

jserv:


jserv:
This implementation use less NEON instruction, but no performance evaluation appears yet.

PR #50, opened 2020-06-27
Add Aarch64 fastpath for _mm_movemask_epi8

jserv:


sse2neon.h
```diff
@@ -1679,6 +1679,17 @@ FORCE_INLINE __m128i _mm_sll_epi16(__m128i a, __m128i count)
...
+#if defined(__aarch64__)
+    uint8x16_t input = vreinterpretq_u8_m128i(a);
+    const int8_t ALIGN_STRUCT(16) xr[8] = {-7, -6, -5, -4, -3, -2, -1, 0};
+    const uint8x8_t mask_and = vdup_n_u8(0x80);
+    const int8x8_t mask_shift = vld1_s8(xr);
+    uint8x8_t lo = vget_low_u8(input);
+    uint8x8_t hi = vget_high_u8(input);
+
+    return vaddv_u8(vshl_u8(vand_u8(lo, mask_and), mask_shift)) +
+           (vaddv_u8(vshl_u8(vand_u8(hi, mask_and), mask_shift)) << 8);
```
marktwtn, line 1691: Can we handle the `__m128i` data at once without separate it to low and high?

Change the intrinsics and `xr` array like the following:
- `vaddv_u8` -> `vaddvq_u8`, `vshl_u8` -> `vshlq_u8`, etc.
- `xr[16] = {-15, -14, -13, -12, -11, -10, -9, -8, -7, -6, -5, -4, -3, -2, -1, 0}`

jserv:
I transfer the ownership to @marktwtn in case of my my absence on further work.

sse2neon.h
marktwtn: The original design is correct.
If it is modified as I suggested, there would have some problems.

However, I still rewrite it to make it more elegant.

marktwtn:
@jserv please review the pull request.
If the modification is acceptable, then we can merge it.

PR #56, opened 2020-07-05
Improve the performance of popcnt function

danlark1:
This remove unnecessary instruction and let's rely on the compiler. See https://gcc.godbolt.org/z/WWEq9L

__builtin_popcount is presented in all compilers where __GNUC__ is defined, also it is not recommended for all compilers to use __has_builtin for old functions, see note [here](https://clang.llvm.org/docs/LanguageExtensions.html#has-builtin)

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -3623,7 +3623,9 @@ FORCE_INLINE __m128i _mm_minpos_epu16(__m128i a)
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_popcnt_u32
 FORCE_INLINE int _mm_popcnt_u32(unsigned int a)
 {
-#if defined(__aarch64__)
+#ifdef __GNUC__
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -3624,7 +3624,11 @@ FORCE_INLINE __m128i _mm_minpos_epu16(__m128i a)
 FORCE_INLINE int _mm_popcnt_u32(unsigned int a)
 {
 #if defined(__aarch64__)
+#if defined(__GNUC__)
+    return __builtin_popcount(a);
```
jserv, line 3628: Can we assume that both gcc and clang produce the same instruction sequence with `__builtin_popcount`? If so, the check for gcc can be eliminated when target is Aarch64.

danlark1: It was written for the non-supported compilers mostly -- better portability

And yes, both gcc and clang produce at least the code written before

PR #59, opened 2020-07-06
Define macro __has_builtin for GCC prior to 10

jserv:
Clang and gcc-10 supports macro __has_builtin, and GCC <= 9 lacks of the
support. This patch attempts to make a limited list for the builtin
functions that we might make use of.


jserv:
@danlark1 discussed the use of macro `__has_builtin` in #56 as it is not recommended for all compilers to use `__has_builtin`.

PR #60, opened 2020-07-08
refactor: Unify validation functions in test suite

### LINKED ISSUE #58: Unify validation functions in test suite
The parameter order of the validation functions is not unified.
Some of them are using the reverse order.

Besides, the function naming should be consistent as well.

marktwtn:
Rename the parameters of the validation functions for better meaning.
Unify the parameter order and function naming.

Close #58.


jserv:
So, little endian is assumed for test suite. We might clarify.

PR #63, opened 2020-07-12
Provide semi-optimized _mm_aesenc_si128 for ARMv7-A

jserv:
For the sake of orthogonality, every supported intrinsic shall be
implemented for both ARMv7-A+NEON and ARMv8-A with diverse extensions.
This patch provides a semi-optimized _mm_aesenc_si128 for ARMv7-A with
NEON. There is room for performance improvements. See

  "Higher-Order Masking in Practice: A Vector Implementation of Masked AES
   for ARM NEON"
  https://link.springer.com/chapter/10.1007/978-3-319-16715-2_10

And corresponding source code:
  https://github.com/Ko-/aes-masked-neon


sse2neon.h
```diff
@@ -4006,15 +4032,64 @@ FORCE_INLINE __m128i _mm_aesenc_si128(__m128i EncBlock, __m128i RoundKey)
...
+         aes_table[2][(x2 >> 16) & 0xff] ^ aes_table[3][x3 >> 24]));
+
+    return _mm_xor_si128(out, RoundKey);
+#endif
+
+#undef SSE2NEON_AES_DATA
 }
-#elif defined(__ARM_FEATURE_CRYPTO)
+
+#elif defined(__aarch64__) && defined(__ARM_FEATURE_CRYPTO)
```
marktwtn, line 4085: Is `defined(__aarch64__)` necessary?
The intrinsics below seem to be runnable on both ARM 32-bit and 64-bit architecture.

PR #64, opened 2020-07-12
Check and set build option for Arm/Aarch64 targets

jserv:
We shall detect the configurations of clang/gcc in the presence of
NEON-based intrinsics.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -62,6 +62,27 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+/* Architecture-specific build options */
+/* FIXME: #pragma GCC push_options is only available on GCC */
+#if defined(__GNUC__)
+#if defined(__arm__) && __ARM_ARCH == 7
+/* According to ARM C Language Extensions Architecture specification,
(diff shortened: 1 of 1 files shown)


jserv:
> However, I am not sure about the benefit from `fpu=neon` and `+simd`.

My intention was to make gcc errors more verbose. Without this patch, the command `make CROSS_COMPILE=arm-linux-gnueabihf- check` raises 2 errors when `ARCH_CFLAGS = -mfpu=neon` is not set:
```
error: inlining failed in call to always_inline ‘int64x2_t vreinterpretq_s64_s32(int32x4_t)’: target specific option mismatch
 vreinterpretq_s64_s32 (int32x4_t __a)
error: inlining failed in call to always_inline ‘int32x4_t vdupq_n_s32(int32_t)’: target specific option mismatch
 vdupq_n_s32 (int32_t __a)
```

With this patch, the error would be
```
#error "You must enable NEON instructions (e.g. -mfpu=neon) to use SSE2NEON."
```

Even when you remove the above `#error` statement, you would get the following error messages:
```
error: inlining failed in call to always_inline ‘__m128i _mm_setzero_si128()’: target specific option mismatch
```

It would be better for trouble shooting.

PR #65, opened 2020-07-12
Implement precise _mm_min_ps/_mm_max_ps

### LINKED ISSUE #57: Correct _mm_min_ps, _mm_max_ps implementation
_mm_min_ps and _mm_max_ps cannot be accurately emulated with single vminq_f32/vmaxq_f32(and also vminnmq_f32 and vmaxnmq_f32) instruction.

We need special handling when both inputs are zeros and either input is NaN.

https://tavianator.com/fast-branchless-raybounding-box-intersections-part-2-nans/
https://www.felixcloutier.com/x86/minps

Here is an implementation of vmin/vmax which emulates _mm_min_ps/_mm_max_ps exactly(as far as I've tested) 
 
https://github.com/lighttransport/embree-aarch64/commit/e4a2f68d1d849de4521c495bb6fd9a5f6017ad0d

jserv:
The mapping between SSE2 _mm_min_ps/_mm_max_ps and NEON
vminq_f32/vmaxq_f32 is not accurate, that would give
inconsitent rendering (e.g. holes, NaN pixels). This patch introduces
tunable configuration and precise implementation of _mm_min_ps and
_mm_max_ps. The precise implementation is built by default.

Close #57

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -44,6 +44,16 @@
  * SOFTWARE.
  */
 
+/* Tunable configurations */
+
+/* Enable precise implementation of _mm_min_ps and _mm_max_ps
+ * This would slow down the computation a bit, but gives consistent result with
+ * x86 SSE2. (e.g. would solve a hole or NaN pixel in the rendering result)
+ */
+#ifndef SSE2NEON_PRECISE_MINMAX
+#define SSE2NEON_PRECISE_MINMAX (0)
+#endif
+
 #if defined(__GNUC__) || defined(__clang__)
 #pragma push_macro("FORCE_INLINE")
 #pragma push_macro("ALIGN_STRUCT")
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -43,6 +43,16 @@
...
  */
 
+/* Tunable configurations */
+
+/* Enable precise implementation of _mm_min_ps and _mm_max_ps
+ * This would slow down the computation a bit, but gives consistent result with
+ * x86 SSE2. (e.g. would solve a hole or NaN pixel in the rendering result)
+ */
+#ifndef SSE2NEON_PRECISE_MINMAX
+#define SSE2NEON_PRECISE_MINMAX (1)
```
danlark1, line 53: I'd rather have it off by default and if the users need the special care for NaNs, suggest turn it on

jserv:
Thank @danlark1 for the feedback. I have changed the default configuration and added notice in top-level document.

sse2neon.h
```diff
@@ -2501,31 +2511,152 @@ FORCE_INLINE __m128 _mm_rsqrt_ss(__m128 in)
...
+    float32x4_t v_special_case = vbslq_f32(v_both_are_zeros, b, v_min);
+    v_special_case = vbslq_f32(v_src0_is_nan, b, v_special_case);
+    v_special_case = vbslq_f32(v_src1_is_snan, b, v_special_case);
+
+    // Requie NaN or both zero case handling?
+    const uint32x4_t v_require_special_handling =
+        vorrq_u32(v_src1_is_snan, vorrq_u32(v_both_are_zeros, v_src0_is_nan));
+
+    // use min(a, b) when !(require special handling)
+    return vbslq_f32(v_require_special_handling, v_special_case, v_min);
```
marktwtn, line 2589: Isn't `v_special_case` the required return value?

jserv: @syoyo, Could you check the special case?

syoyo: @jserv I can't understand what @marktwtn is saying. BTW, you can omit src1 snan check. I found checking src1 snan  is redundant after I wrote the initial code.

marktwtn: I mean it seems like variable `v_special_case` and `vbslq_f32(v_require_special_handling, v_special_case, v_min)` have the same value?
If it is, then the line 2583~2589 should be removed.

marktwtn: @syoyo any idea?

syoyo: @marktwtn as I said you are better to write unit tests.

marktwtn: @jserv please remove the redundant code and return the `v_special_case`.

sse2neon.h
```diff
@@ -2501,31 +2511,152 @@ FORCE_INLINE __m128 _mm_rsqrt_ss(__m128 in)
...
+    float32x4_t v_special_case = vbslq_f32(v_both_are_zeros, b, v_max);
+    v_special_case = vbslq_f32(v_src0_is_nan, b, v_special_case);
+    v_special_case = vbslq_f32(v_src1_is_snan, b, v_special_case);
+
+    // Requie NaN or both zero case handling?
+    const uint32x4_t v_require_special_handling =
+        vorrq_u32(v_src1_is_snan, vorrq_u32(v_both_are_zeros, v_src0_is_nan));
+
+    // use max(a, b) when !(require special handling)
+    return vbslq_f32(v_require_special_handling, v_special_case, v_max);
```
marktwtn, line 2619: The redundant code here should be removed as well.

jserv:
Transfer the ownership to @marktwtn, who will think of the further test case(s) as well.

marktwtn:
The redundant code has been removed.

jserv:
> The redundant code has been removed.

It looks great. However, we still lack of the test cases which take the accuracy into considerations.

marktwtn:
I wonder the test case is for including NaN in the compared data, make sure the `SSE2NEON_PRECISE_MINMAX` would be activated and tested or both?

jserv:
> I wonder the test case is for including NaN in the compared data, make sure the `SSE2NEON_PRECISE_MINMAX` would be activated and tested or both?

The current test suite is based on randomized input. Thus, it is not ready to validate the special cases such as NaN. We can introduce them later.

marktwtn:
> > I wonder the test case is for including NaN in the compared data, make sure the `SSE2NEON_PRECISE_MINMAX` would be activated and tested or both?
> 
> The current test suite is based on randomized input. Thus, it is not ready to validate the special cases such as NaN. We can introduce them later.

To handle the NaN, the `SSE2NEON_PRECISE_MINMAX` should be turned on, which does not in our current test suite.
Since we can introduce them later, shall we open an issue for it and close this pull request?

jserv:
> To handle the NaN, the `SSE2NEON_PRECISE_MINMAX` should be turned on, which does not in our current test suite.
> Since we can introduce them later, shall we open an issue for it and close this pull request?

Yes, add some FIXME/TODO within this pull request and create new issue(s) to address.

marktwtn:
The `TODO` has been added in the `tests/impl.cpp`.

PR #69, opened 2020-07-15
refactor: Classify intrinsic functions

marktwtn:
The classification can help the developer find the unimplemented
intrinsic functions quickly.
The classification is referenced from:
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ Once the conversion is implemented, the test can be added with the following ste
 * File `tests/impl.h`
 
   Add the intrinsic in `enum InstructionTest`. The naming convention should be `IT_MM_XXX`.
+  And place it in the correct classification with the alphabetical order.
+  The classification can be referenced from [Intel Intrinsics Guide](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#).
 
 * File `tests/impl.cpp`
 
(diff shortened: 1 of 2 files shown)


marktwtn:
> Inconsistent style

It is weird that I do not modify the code which causes the inconsistent style.
And it is not detected in the previous commits.

marktwtn:
@jserv it has a weird coding style checking like this:

sse2neon.h
```
-    // AES
+// AES
 #if !defined(__arm__) && __ARM_ARCH != 7
     IT_MM_AESENC_SI128,
 #endif
```

marktwtn:
> clang-format in script ".ci/check-format.sh" should be replaced with "clang-format-6"

I am pretty sure that it would use `clang-format-6` in Travis CI without changing the script.
I'm afraid that the weird coding style checking is caused by `clang-format-6` and `.clang-format`.

jserv:
> it has a weird coding style checking like this:
> ```
> -    // AES
> +// AES
>  #if !defined(__arm__) && __ARM_ARCH != 7
>      IT_MM_AESENC_SI128,
>  #endif
> ```

I don't think it is a problem after #63 is merged.

PR #71, opened 2020-07-15
Fix coding style

marktwtn:


sse2neon.h
```diff
@@ -1214,8 +1214,8 @@ FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
     __asm__ __volatile__(
         "vtbl.8  %e[ret], {%e[tbl], %f[tbl]}, %e[idx]\n"
         "vtbl.8  %f[ret], {%e[tbl], %f[tbl]}, %f[idx]\n"
-        : [ret] "=&w"(ret)
-        : [tbl] "w"(tbl), [idx] "w"(idx_masked));
+        : [ ret ] "=&w"(ret)
+        : [ tbl ] "w"(tbl), [ idx ] "w"(idx_masked));
```
jserv, line 1218: I have no idea about the style. Is it enforced by `clang-format`?

marktwtn: Yes, they are changed by the Makefile rule `make indent`.
However, it is not treated as error in the previous commits, which really makes me confused.

jserv: The root cause is versioning of `clang-format`. Version 6 of `clang-format` generates something like `[ret] "=&w"(ret)` while version 10 generates `[ ret ] "=&w"(ret)`.

jserv: Alternatively, we can explicitly use `clang-format-6.0` for indention.

marktwtn: I'll close this pull request and create another one for assigning `clang-format-6.0` on `Travis CI`.

jserv: It is a confirmed issue: https://bugs.llvm.org/show_bug.cgi?id=45108
Recent `clang-format` has fixed the above indention problem with inline assembly.

PR #74, opened 2020-07-16
refactor: Remove unnecessary comments of test

marktwtn:
Only mark those intrinsics which have not implemented the test yet.


jserv:
How about changing "Unit test not implemented" to "Unimplemented"?

marktwtn:
> How about changing "Unit test not implemented" to "Unimplemented"?

Sounds good.

PR #75, opened 2020-07-16
Implement _mm_aeskeygenassist_si128

jserv:
Both crypto extension and NEON implementation were derived from the
original work of Michael G. Kazakov. See
    https://kazakov.life/2017/11/01/cryptocurrency-mining-on-ios-devices/


tests/impl.cpp
```diff
@@ -2850,6 +2876,20 @@ bool test_mm_aesenc_si128(const int32_t *a, const int32_t *b)
     return validate128(resultReference, resultIntrinsic);
 }
 
+// FIXME: improve the test case for AES-256 key expansion.
+// Reference:
+// https://github.com/randombit/botan/blob/master/src/lib/block/aes/aes_ni/aes_ni.cpp
+bool test_mm_aeskeygenassist_si128(const int32_t *a, const int32_t *b)
```
marktwtn, line 2882: The parameter `const int32_t *b` can be removed since it is not used.

jserv: I kept the second parameter for the sake of consistent parameter list with `_mm_aeskeygenassist_si128`.

Issue #70, opened 2020-07-15
Missing _mm_{load,loadu,store,storeu}_pd


jserv:
The suffix `pd` means the vectors contains doubles (pd stands for packed double-precision). The following intrinsics are missing:
* `_mm_load_pd`
* `_mm_loadu_pd`
* `_mm_store_pd`
* `_mm_store_pd`

For A64, the `_mm_load_pd` and `_mm_storeu_pd` can be implemented as following:
```c
// Load 128-bits (composed of 2 packed double-precision (64-bit) floating-
// point elements) from memory into dst.
// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_load_pd                                                                             
FORCE_INLINE __m128d _mm_load_pd(const double *p)
{
#if defined(__aarch64__)
    return (__m128d)(vld1q_f64(p));
#else
    /* A32 only */
#endif
}

// Stores four single-precision, floating-point values.
// https://msdn.microsoft.com/en-us/library/44e30x22(v=vs.100).aspx
FORCE_INLINE void _mm_storeu_pd(double *p, __m128d a)
{
#if defined(__aarch64__)
    vst1q_f64(p, (__m128d)(a));
#else
    /* A32 only */
#endif
}
```

marktwtn:
Take `_mm_load[u]_pd` as an example, I might use `vld1q_f32` to do the implementation and then cast it to the corresponding data type of `__m128d`(which is `float32x4_t`) in **A32**.

PR #78, opened 2020-07-17
AArch64 implementation of PS FMA and ADDSUB, and some PD load/store/convert

JishinMaster:
Hi,

I have implemented, using the AARCH64 Neon intrinsics, the following functions : 
_mm_cvtpd_ps 
_mm_cvtps_pd
 _mm_castpd_si128
 _mm_fmadd_ps,
_mm_addsub_ps
_mm_load_pd
 _mm_loadu_pd
 _mm_store_pd
 _mm_storeu_pd

I have got some more new additions which comes from my simd_utils library, but it seemed too much for one pull request.

Please let me know if you need modifications to this pull request.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1271,8 +1271,8 @@ FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
     __asm__ __volatile__(
         "vtbl.8  %e[ret], {%e[tbl], %f[tbl]}, %e[idx]\n"
         "vtbl.8  %f[ret], {%e[tbl], %f[tbl]}, %f[idx]\n"
-        : [ ret ] "=&w"(ret)
-        : [ tbl ] "w"(tbl), [ idx ] "w"(idx_masked));
+        : [ret] "=&w"(ret)
+        : [tbl] "w"(tbl), [idx] "w"(idx_masked));
     return vreinterpretq_m128i_s8(ret);
 #else
     // use this line if testing on aarch64
@@ -4298,8 +4298,8 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
 {
 #if defined(__aarch64__) && defined(__ARM_FEATURE_CRC32)
     __asm__ __volatile__("crc32cb %w[c], %w[c], %w[v]\n\t"
-                         : [ c ] "+r"(crc)
-                         : [ v ] "r"(v));
+                         : [c] "+r"(crc)
+                         : [v] "r"(v));
 #else
     crc ^= v;
     for (int bit = 0; bit < 8; bit++) {
@@ -4319,8 +4319,8 @@ FORCE_INLINE uint32_t _mm_crc32_u16(uint32_t crc, uint16_t v)
 {
 #if defined(__aarch64__) && defined(__ARM_FEATURE_CRC32)
     __asm__ __volatile__("crc32ch %w[c], %w[c], %w[v]\n\t"
-                         : [ c ] "+r"(crc)
-                         : [ v ] "r"(v));
+                         : [c] "+r"(crc)
+                         : [v] "r"(v));
 #else
     crc = _mm_crc32_u8(crc, v & 0xff);
     crc = _mm_crc32_u8(crc, (v >> 8) & 0xff);
@@ -4335,8 +4335,8 @@ FORCE_INLINE uint32_t _mm_crc32_u32(uint32_t crc, uint32_t v)
 {
 #if defined(__aarch64__) && defined(__ARM_FEATURE_CRC32)
     __asm__ __volatile__("crc32cw %w[c], %w[c], %w[v]\n\t"
-                         : [ c ] "+r"(crc)
-                         : [ v ] "r"(v));
+                         : [c] "+r"(crc)
+                         : [v] "r"(v));
 #else
     crc = _mm_crc32_u16(crc, v & 0xffff);
     crc = _mm_crc32_u16(crc, (v >> 16) & 0xffff);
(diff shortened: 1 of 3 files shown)


jserv:
@JishinMaster, Thanks for contributing. Can you use `clang-format` to indent? Also, A32 specific implementations are missing.

sse2neon.h
```diff
@@ -1219,8 +1271,8 @@ FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
     __asm__ __volatile__(
         "vtbl.8  %e[ret], {%e[tbl], %f[tbl]}, %e[idx]\n"
         "vtbl.8  %f[ret], {%e[tbl], %f[tbl]}, %f[idx]\n"
-        : [ret] "=&w"(ret)
-        : [tbl] "w"(tbl), [idx] "w"(idx_masked));
+        : [ ret ] "=&w"(ret)
```
jserv, line 1274: clang-format-10 is known to a bit weird for inline assembly. You can use [clang-format-6](https://releases.llvm.org/6.0.1/tools/clang/docs/ClangFormat.html).

JishinMaster:
I have applied clang-format (make indent) with clang-format-10, and checked with the check-format.sh script.
Travis uses clang-format-6 and still complains about indentation.

I have added the armv7 neon equivalent where it made some sense (fmadd_ps and addsub_ps).
I do no think it would be relevant to manually implement other float64 functions using the armv7 neon instructions, since it was not native.
We would be better of using vfpv4 or even plain C, or have a define flag which would turn on float64 on AArch64 CPUs. 
What do you think?

jserv:
> I have added the armv7 neon equivalent where it made some sense (fmadd_ps and addsub_ps).
> I do no think it would be relevant to manually implement other float64 functions using the armv7 neon instructions, since it was not native.
> We would be better of using vfpv4 or even plain C, or have a define flag which would turn on float64 on AArch64 CPUs.
> What do you think?

If some intrinsic is unlikely to be implemented in A32-only, C implementation can fit. We don't have to subject single pull request with full optimizations

jserv:
Related: #70

sse2neon.h
```diff
@@ -2390,6 +2442,41 @@ FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
...
+#endif
+}
+
+
+
+// Alternatively add and subtract packed single-precision (32-bit)
+// floating-point elements in a to/from packed elements in b, and store the
+// results in dst.
+//
+// The SSE3 flag is required for this function to be available on X86_64
```
jserv, line 2470: The comment about SSE3 is not necessary to SSE2NEON.

sse2neon.h
```diff
@@ -2390,6 +2442,41 @@ FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
...
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=addsub_ps&expand=2549,223
+FORCE_INLINE __m128 _mm_addsub_ps(__m128 a, __m128 b)
+{
+    __m128 mask = {-1.0f, 1.0f, -1.0f, 1.0f};
+    return _mm_fmadd_ps(b, mask, a);
+}
+
```
jserv, line 2477: Change 3 blank lines to single for consistent style.

sse2neon.h
```diff
@@ -2390,6 +2442,41 @@ FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
...
+{
+#if defined(__aarch64__)
+    return vreinterpretq_m128_f32(vfmaq_f32(vreinterpretq_f32_m128(c),
+                                            vreinterpretq_f32_m128(b),
+                                            vreinterpretq_f32_m128(a)));
+#else
+    return _mm_add_ps(_mm_mul_ps(a, b), c);
+#endif
+}
+
```
jserv, line 2463: Ditto. Style.

sse2neon.h
```diff
@@ -2390,6 +2442,41 @@ FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
     return vreinterpretq_m128i_s16(vqaddq_s16(prod1, prod2));
 }
 
+
```
jserv, line 2445: Ditto. Style.

jserv:

tests/impl.cpp
```diff
@@ -974,6 +980,18 @@ bool test_mm_load_sd(const double *p)
     return validateDouble(a, p[0], 0);
 }
 
+bool test_mm_load_pd(const double *p)
+{
+    __m128d a = _mm_load_pd(p);
+    return validateDouble(a, p[0], p[1]);
+}
+
+bool test_mm_loadu_pd(const double *p)
```
danlark1, line 989: I don't see that this function to be called in the tests

sse2neon.h
```diff
@@ -2390,6 +2439,33 @@ FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
     return vreinterpretq_m128i_s16(vqaddq_s16(prod1, prod2));
 }
 
+// Computes the fused multiple add product of 32-bit floating point numbers.
+//
+// Return Value
+// Multiplies A and B, and adds C to the temporary result before returning it.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_fmadd&expand=2549
+FORCE_INLINE __m128 _mm_fmadd_ps(__m128 a, __m128 b, __m128 c)
```
danlark1, line 2447: Technically speaking, this is not an SSE extension, this was introduced with fma extension which took place even after AVX

tests/impl.cpp
jserv: Continuous work on #81

PR #80, opened 2020-07-17
Implement _mm_testz_si128

jserv:


tests/impl.cpp
```diff
@@ -1857,6 +1859,14 @@ bool test_mm_cvtps_epi32(const float _a[4])
     return validateInt32(ret, trun[0], trun[1], trun[2], trun[3]);
 }
 
+bool test_mm_testz_si128(const int32_t *_a, const int32_t *_b)
+{
+    __m128i a = _mm_load_si128((const __m128i *) _a);
+    __m128i b = _mm_load_si128((const __m128i *) _b);
+    bool testz = validate128(a, b);
+    return _mm_testz_si128(a, b) == testz;
```
marktwtn, line 1867: I don't understand the comparison here.

Consider the following example:
||a=0b00...00, b=0b00...00|a=0b0101...01, b=0b1010...10|
|:-:|:-:|:-:|
|validate128|true|false|
|_mm_testz_si128|ZF=1|ZF=1|

tests/impl.cpp
```diff
@@ -1857,6 +1859,22 @@ bool test_mm_cvtps_epi32(const float _a[4])
...
 }
 
+bool test_mm_testz_si128(const int32_t *_a, const int32_t *_b)
+{
+    __m128i a = _mm_load_si128((const __m128i *) _a);
+    __m128i b = _mm_load_si128((const __m128i *) _b);
+    int testz = 0;
+    int i;
+    for (i = 0; i < 2; i++) {
+        if ((((SIMDVec *) &a)->m128_u64[i] & ((SIMDVec *) &b)->m128_u64[i]) ==
```
marktwtn, line 1869: Replace with macro `vreinterpretq_nth_u64_m128i`.

tests/impl.cpp
```diff
@@ -1857,6 +1859,22 @@ bool test_mm_cvtps_epi32(const float _a[4])
...
+            0) {
+            testz = 1;
+            break;
```
marktwtn, line 1872: I think it should be
```c
int testz = 1;
...
for
  if(... != 0)
    testz = 0;
    break;
```

tests/impl.cpp
jserv: Macro `vreinterpretq_nth_u64_m128i` is defined in header `sse2neon.h`. If the test suite is built for x86 target, header `sse2neon.h` will not be included. Then, this macro does not effect accordinly.

tests/impl.cpp
jserv: Even we have 10K randomized input vectors for the test case, but it is still hard to diagnose the logic errors though.

Issue #66, opened 2020-07-12
Missing intrinsics required for building Cycles


jserv:
[Cycles](https://docs.blender.org/manual/en/latest/render/cycles/introduction.html) is Blender's physically-based path tracer for production rendering. Recently, Apple Inc. sent Aarch64 patch to Blender developer as shown on [D8237: Cycles: support neon instructions for arm64 processors](https://developer.blender.org/D8237). However, file `intern/cycles/util/util_sse_to_neon.h` provided by  [D8237](https://developer.blender.org/D8237) was incomplete, and Blender developers were discussing the integration of sse2neon.

Missing intrinsics in SSE2NEON required for building [Cycles](https://docs.blender.org/manual/en/latest/render/cycles/introduction.html):
- [x] `_mm_cmpnle_ps`
- [x] `_mm_cmpnlt_ps`
- [x] `_mm_fmadd_ps` (FMA)
- [x] `_mm_fmsub_ps` (FMA)
- [ ] `_mm_fnmadd_ps` (FMA)
- [ ] `_mm_fnmsub_ps` (FMA)
- [x] `_mm_floor_ps`
- [x] `_mm_ceil_ps`
- [x] `_mm_castps_pd`
- [x] `_mm_stream_ps`
- [x] `_mm_blendv_ps`
- [x] `_mm_dp_ps`
- [x] `_mm_stream_load_si128`

[D8237](https://developer.blender.org/D8237) also defines `_mm_cmple_epi32` and `_mm_cmpge_epi32`, which are not part of SSE intrinsics.

Fortunately, the simplified implementations are available in [D8237](https://developer.blender.org/D8237).

JishinMaster:
I was waiting for my first pull request to be accepted, but I have almost all of those functions ready :)
I will try to get my pull request ready by tomorrow!

jserv:
Only one instruction is missing for building Cycles. That is `_mm_castps_pd` as described in #91.

PR #83, opened 2020-07-21
Added some of the missing intrinsics required for building Cycles

JishinMaster:
This pull request adds some of the function mentioned in #66.

`_mm_cmple_epi32` and `_mm_cmpge_epi32` are inspired from `_mm_cmpge_ps` and `_mm_cmpge_ps`.
_mm_cmpnle_ps and _mm_cmpnlt_ps are implemented as not(cmple_ps) and not(cmplt_ps), using an invert mask.
_mm_fmadd_ps was added from the previous pull request.
_mm_fmsub_ps, _mm_fnmadd_ps, and _mm_fnmsub_ps were added.

_mm_floor_ps and _mm_ceil_ps use _mm_round_ps.
_mm_round_ps uses the AARCH64 rounding functions( vrndXX_f32), or the ceilf/floorf/roundf from the compiler libm.
The _MM_SET_ROUNDING_MODE is not part of this pull request. As of now I have managed to change the rounding mode through read and write to the FPCR register. It would look something like this  (for AARCH64) : 

```c
typedef struct {
  uint16_t res0;
  uint16_t res1 : 6;
  uint8_t bit21 : 1;
  uint8_t bit22 : 1;
  uint32_t res3;

} fpcr_bitfield;

typedef union {
  fpcr_bitfield field;
  uint64_t value;
} reg;

reg r;
asm volatile("mrs %0, FPCR" : "=r"(r.value)); /* read */
 r.field.bit21 = 0;
 r.field.bit22 = 1;
asm volatile("msr FPCR, %0" ::"r"(r));        /* write */
```
Please let me know if this kind of approach is suitable.

_mm_blendv_ps has been added using the vbslq_f32 instruction (_mm_round_ps uses it for aarch32).

_mm_stream_ps and _mm_stream_load_si128 have not been added, but a quick patch would be to internally use simple load and store. From what I have benchmarked on X86 CPUs, those instructions are useful for loading and storing large chunks of data, larger than the platform cache, but they are not really faster than memcpy (tested with a recent glibc and compiler).

_mm_castps_pd and _mm_dp_ps have not been added.


jserv:
Thank @JishinMaster again for contributing. However, this pull request is a bit difficult for reviewing because it mixes several changes. Can you split into classified pull requests instead? i.e., rounding, _mm_stream_*, etc.

jserv:
FMA intrinsics are beyond the scope of SSE2NEON. The intrinsics excluding FMA in this pull request are already implemented. So, let's close.

PR #84, opened 2020-07-22
Implement _mm_cmp*_{ss,ps}

afcidk:
Implement the following intrinsics
* mm_cmplt_ss
* mm_cmpgt_ss
* mm_cmpge_ss
* mm_cmple_ss
* mm_cmpeq_ss
* mm_cmpord_ss
* mm_cmpunord_ps
* mm_cmpunord_ss
* mm_cmpneq_ss
* mm_cmpnge_ps
* mm_cmpnge_ss
* mm_cmpngt_ps
* mm_cmpngt_ss
* mm_cmpnle_ps
* mm_cmpnle_ss
* mm_cmpnlt_ps
* mm_cmpnlt_ss


sse2neon.h
```diff
@@ -3083,6 +3215,34 @@ FORCE_INLINE __m128 _mm_cmpord_ps(__m128 a, __m128 b)
...
+FORCE_INLINE __m128 _mm_cmpunord_ps(__m128 a, __m128 b)
+{
+    return vreinterpretq_m128_u32(
+        vmvnq_u32(vreinterpretq_u32_m128(_mm_cmpord_ps(a, b))));
+}
+
+// Compares for unordered.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/2as2387b(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpunord_ss(__m128 a, __m128 b)
+{
```
jserv, line 3239: Consider the following implementation:
```c
uint32x4_t f32a = vceqq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(a));
uint32x4_t f32b = vceqq_f32(vreinterpretq_f32_m128(b), vreinterpretq_f32_m128(b));
return vreinterpretq_m128_f32(vmvnq_u32(vandq_u32(f32a, f32b)));
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
         vceqq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b))));
 }
 
+// Compares for inequality.
```
jserv, line 818: We can provide `_mm_move_ss` as following:
```c

return vreinterpretq_m128_f32(
    vsetq_lane_f32(vgetq_lane_f32(vreinterpretq_f32_m128 (b), 0),
                               vreinterpretq_f32_m128 (a), 0));
```

And then, we can reimplement several intrinsics using `_mm_move_ss`.

For example:
```c
FORCE_INLINE __m128 _mm_cmpneq_ss(__m128 a, __m128 b)
{
    return _mm_move_ss(a, _mm_cmpneq_ps(a, b));
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ekya8fh4(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpneq_ss(__m128 a, __m128 b)
```
jserv, line 820: It can be reimplemented as following:
```c
FORCE_INLINE __m128 _mm_cmpneq_ss(__m128 a, __m128 b)
{
    return _mm_cmplt_ps(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+{
+    float32_t value =
+        vgetq_lane_f32(vreinterpretq_f32_m128(_mm_cmpneq_ps(a, b)), 0);
+    return vreinterpretq_m128_f32(
+        vsetq_lane_f32(value, vreinterpretq_f32_m128(a), 0));
+}
+
+// Compares for not greater than or equal.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/wsexys62(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpnge_ps(__m128 a, __m128 b)
```
jserv, line 830: Ditto. Reimplement as
```c
FORCE_INLINE __m128 _mm_cmpnge_ps(__m128 a, __m128 b)
{
    return _mm_cmplt_ps(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+{
+    return vreinterpretq_m128_u32(vmvnq_u32(
+        vcgeq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b))));
+}
+
+// Compares for not greater than or equal.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/fk2y80s8(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpnge_ss(__m128 a, __m128 b)
```
jserv, line 838: Ditto. Reimplement as
```c
FORCE_INLINE __m128 _mm_cmpnge_ss(__m128 a, __m128 b)
{
    return _mm_cmplt_ss(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+{
+    float32_t value =
+        vgetq_lane_f32(vreinterpretq_f32_m128(_mm_cmpnge_ps(a, b)), 0);
+    return vreinterpretq_m128_f32(
+        vsetq_lane_f32(value, vreinterpretq_f32_m128(a), 0));
+}
+
+// Compares for not greater than.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/d0xh7w0s(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpngt_ps(__m128 a, __m128 b)
```
jserv, line 848: Ditto. Reimplement as following:
```c
FORCE_INLINE __m128 _mm_cmpngt_ps(__m128 a, __m128 b)
 {
    return _mm_cmple_ps(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+{
+    return vreinterpretq_m128_u32(vmvnq_u32(
+        vcgtq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b))));
+}
+
+// Compares for not greater than.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/z7x9ydwh(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpngt_ss(__m128 a, __m128 b)
```
jserv, line 856: Ditto. Reimplement as
```c
FORCE_INLINE __m128 _mm_cmpngt_ss(__m128 a, __m128 b)
{
    return _mm_cmple_ss(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+        vsetq_lane_f32(value, vreinterpretq_f32_m128(a), 0));
+}
+
+// Compares for not less than or equal.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/6a330kxw(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpnle_ps(__m128 a, __m128 b)
+{
+    return vreinterpretq_m128_u32(vmvnq_u32(
+        vcleq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b))));
+}
```
jserv, line 870: Ditto. Re-implement:
```c
FORCE_INLINE __m128 _mm_cmpnle_ps(__m128 a, __m128 b)
{
    return simde_mm_cmpgt_ps(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+
+// Compares for not less than or equal.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/z7x9ydwh(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpnle_ss(__m128 a, __m128 b)
```
jserv, line 874: Ditto. Reimplement as
```c
FORCE_INLINE __m128 _mm_cmpnle_ss(__m128 a, __m128 b)
{
    return _mm_cmpgt_ss(a, b);
}
```

sse2neon.h
```diff
@@ -815,6 +815,88 @@ FORCE_INLINE __m128 _mm_cmpneq_ps(__m128 a, __m128 b)
...
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/4686bbdw(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpnlt_ps(__m128 a, __m128 b)
+{
+    return vreinterpretq_m128_u32(vmvnq_u32(
+        vcltq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b))));
+}
+
+// Compares for not less than.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/56b9z2wf(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpnlt_ss(__m128 a, __m128 b)
```
jserv, line 892: Ditto.

sse2neon.h
```diff
@@ -2868,6 +2950,16 @@ FORCE_INLINE __m128 _mm_cmplt_ps(__m128 a, __m128 b)
         vcltq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b)));
 }
 
+// Compares for less than
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/fy94wye7(v=vs.100)
+FORCE_INLINE __m128 _mm_cmplt_ss(__m128 a, __m128 b)
```
jserv, line 2955: Ditto. Use `_mm_move_ss(a, _mm_cmplt_ps(a, b))`

sse2neon.h
```diff
@@ -2882,6 +2974,16 @@ FORCE_INLINE __m128 _mm_cmpgt_ps(__m128 a, __m128 b)
         vcgtq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b)));
 }
 
+// Compares for greater than.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/1xyyyy9e(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpgt_ss(__m128 a, __m128 b)
```
jserv, line 2979: Ditto. Use `_mm_move_ss(a, _mm_cmpgt_ps(a, b))`

sse2neon.h
```diff
@@ -2890,6 +2992,16 @@ FORCE_INLINE __m128 _mm_cmpge_ps(__m128 a, __m128 b)
         vcgeq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b)));
 }
 
+// Compares for greater than or equal.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/kesh3ddc(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpge_ss(__m128 a, __m128 b)
```
jserv, line 2997: Ditto. Use `_mm_move_ss(a, _mm_cmpge_ps(a, b))`

sse2neon.h
```diff
@@ -2904,6 +3016,16 @@ FORCE_INLINE __m128 _mm_cmple_ps(__m128 a, __m128 b)
         vcleq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b)));
 }
 
+// Compares for less than or equal.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/a7x0hbhw(v=vs.100)
+FORCE_INLINE __m128 _mm_cmple_ss(__m128 a, __m128 b)
```
jserv, line 3021: Ditto. Use `_mm_move_ss(a, _mm_cmple_ps(a, b))`

sse2neon.h
```diff
@@ -2912,6 +3034,16 @@ FORCE_INLINE __m128 _mm_cmpeq_ps(__m128 a, __m128 b)
         vceqq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b)));
 }
 
+// Compares for equality.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/k423z28e(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpeq_ss(__m128 a, __m128 b)
```
jserv, line 3039: Ditto. Use `_mm_move_ss(a, _mm_cmpeq_ps(a, b))`

sse2neon.h
```diff
@@ -3083,6 +3215,34 @@ FORCE_INLINE __m128 _mm_cmpord_ps(__m128 a, __m128 b)
     return vreinterpretq_m128_u32(vandq_u32(ceqaa, ceqbb));
 }
 
+// Compares for ordered.
+// https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/343t62da(v=vs.100)
+FORCE_INLINE __m128 _mm_cmpord_ss(__m128 a, __m128 b)
```
jserv, line 3220: Ditto. Use `_mm_move_ss(a, _mm_cmpord_ps(a, b))`

sse2neon.h
marktwtn: But what if the value of `a[31:0]` is greater than `b[31:0]`?

tests/impl.cpp
```diff
@@ -3598,8 +3991,9 @@ class SSE2NEONTestImpl : public SSE2NEONTest
                 mTestFloatPointer1[2] = 1.0f / mTestFloatPointer1[2];
                 mTestFloatPointer1[3] = 1.0f / mTestFloatPointer1[3];
             }
-            if (test == IT_MM_CMPGE_PS || test == IT_MM_CMPLE_PS ||
-                test == IT_MM_CMPEQ_PS) {
+            if (test == IT_MM_CMPGE_PS || test == IT_MM_CMPGE_SS ||
+                test == IT_MM_CMPLE_PS || test == IT_MM_CMPLE_SS ||
+                test == IT_MM_CMPEQ_PS || test == IT_MM_CMPEQ_PS) {
```
marktwtn, line 3996: `IT_MM_CMPEQ_PS` -> `IT_MM_CMPEQ_SS`

marktwtn: I think we should include the `NEQ`, `NLE` and `NGE` type as well.

tests/impl.cpp
```diff
@@ -1525,6 +1576,114 @@ bool test_mm_cmplt_ps(const float *_a, const float *_b)
...
+
+    int32_t result[4];
+    result[0] = !(_a[0] < _b[0]) ? -1 : 0;
+    result[1] = !(_a[1] < _b[1]) ? -1 : 0;
+    result[2] = !(_a[2] < _b[2]) ? -1 : 0;
+    result[3] = !(_a[3] < _b[3]) ? -1 : 0;
+
+    __m128 ret = _mm_cmpnlt_ps(a, b);
+    __m128i iret = *(const __m128i *) &ret;
+    return validateInt32(iret, result[0], result[1], result[2], result[3]);
```
marktwtn, line 1607: Why don't you write the test like `test_mm_cmpnlt_ss` does?

marktwtn: I think the way you write in `test_mm_cmpnlt_ss` is better than the original test written style.

tests/impl.cpp
```diff
@@ -1525,6 +1576,114 @@ bool test_mm_cmplt_ps(const float *_a, const float *_b)
     return validateInt32(iret, result[0], result[1], result[2], result[3]);
 }
 
+bool test_mm_cmplt_ss(const float *_a, const float *_b)
+{
+    __m128 a = test_mm_load_ps(_a);
+    __m128 b = test_mm_load_ps(_b);
+
+    float result[4];
+    result[0] = _a[0] < _b[0] ? getNAN() : 0;
```
marktwtn, line 1585: @jserv do you think `getNAN()` would be a little confusing?

jserv: The function `getNAN` was intended to check the propagation of quiet/signal NaN through arithmetic operations. Thus, `get` might be confusing.

marktwtn: I think @afcidk use it in this way is probably caused by referencing the other source code such as `compord()`, which returns `getNAN()`.
We might need to replace these confusing `getNAN()` in the future,

afcidk: I need a number to represent floating number whose bits are all one (0b11111......), and saw that the `getNAN()` function actually use the number I expected. NaN representation only guarantees the exponent part to be all 1 and non-zero number in mantissa.

I think using `getNaN()` is confusing, should I introduce another function?

sse2neon.h
afcidk: I'll keep the original implementation

tests/impl.cpp
marktwtn: How about defining the macros like
```c
#define SET_ALL_BIT_1_64 0x11...11
#define SET_ALL_BIT_1_32 0x1..1
```

jserv: Something like `~0`?

tests/impl.cpp
```diff
@@ -1634,6 +1929,19 @@ float compord(float a, float b)
     return ret;
 }
 
+float compunord(float a, float b)
+{
+    float ret;
+    bool isNANA = isNAN(a);
+    bool isNANB = isNAN(b);
+    if (isNANA || isNANB) {
```
jserv, line 1937: Rewrite as
```c
ret = (isNANA || isNANB) ? getNAN() : 0.0f;
```

tests/impl.cpp
marktwtn: `~0` sounds good to me.

afcidk: We cannot assign `~0` to floating point number directly, that would be represented as -1.000 instead. I like the original implementation of `getNAN()` to cast int32_t to float. Defining it as a macro sounds good, but it seems like we still need an integer-typed variable?
```
const int32_t nan = ~0;
const int64_t nan_64 = ~0;
#define SET_ALL_BIT_1_32 *(float*)&nan;
#define SET_ALL_BIT_1_64 *(double*)&nan_64;
```

JishinMaster: The return value is clearer in NEON ISA than in SSE.
vcleq_f32 takes floating point numbers and returns uint32_x4_t (for which 0xFFFFFFFF makes sense), whereas _mm_cmple_ps returns the same value, but in a __m128 instead of a __m128i.

It think it would be less confusing to code the way you did with test_mm_cmpneq_ps, than to use getNAN :
`bool test_mm_cmpneq_ps(const float *_a, const float *_b)
{
    __m128 a = test_mm_load_ps(_a);
    __m128 b = test_mm_load_ps(_b);

    int32_t result[4];
    result[0] = _a[0] != _b[0] ? -1 : 0;
    result[1] = _a[1] != _b[1] ? -1 : 0;
    result[2] = _a[2] != _b[2] ? -1 : 0;
    result[3] = _a[3] != _b[3] ? -1 : 0;

    __m128 ret = _mm_cmpneq_ps(a, b);
    __m128i iret = *(const __m128i *) &ret;
    return validateInt32(iret, result[0], result[1], result[2], result[3]);
}`

jserv:
Hint for @afcidk: If you Interoperate the proposed changes from @JishinMaster. You can attribute a commit to more than one author by adding `Co-authored-by` trailers to the commit's message. See
https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

jserv:
Related: #66

tests/impl.cpp
afcidk: I think using macros like `SET_ALL_BIT_1_32` should be easy to understand as well once we give up the `getNAN` usage in this `0xFFFFFFFF` value.

I feel fine to both implementation, how do you think @marktwtn ?

tests/impl.cpp
```diff
@@ -1644,11 +1942,16 @@ float compord(float a, float b)
...
-    if (!isNANA && !isNANB) {
-        ret = getNAN();
-    } else {
-        ret = 0.0f;
-    }
+    ret = (!isNANA && !isNANB) ? getNAN() : 0.0f;
+    return ret;
+}
+
+float compunord(float a, float b)
```
jserv, line 1949: Can you revise the naming? It might be confusing with "compound."

tests/impl.cpp
marktwtn: I choose the macro implementation.

tests/impl.cpp
afcidk: I think we can avoid creating a new function and simply write the compare statement in `test_mm_cmpunord_ss` and `test_mm_cmpunord_ps`.

PR #85, opened 2020-07-25
added cmpnlt/e_ps and cmpl/ge_epi32 with tests

JishinMaster:
This pull request is the first split of the original #83.
I have added the required tests.


jserv:
@JishinMaster, You work might conflict with #84, which implements _mm_cmp*_{ss,ps}. Can you review #84 as well?

JishinMaster:
As stated in #66 , you need _mm_cmple_epi32 and _mm_cmpge_epi32.
I had not realized before, but they do not appear in https://software.intel.com/sites/landingpage/IntrinsicsGuide/
What compiler is Blender build with?

I will look at #84

jserv:
> As stated in #66 , you need _mm_cmple_epi32 and _mm_cmpge_epi32.
> I had not realized before, but they do not appear in https://software.intel.com/sites/landingpage/IntrinsicsGuide/
> What compiler is Blender build with?

You are right. There it no `_mm_cmple_epi32` nor `_mm_cmpge_epi32`. The two functions were taken from [D8237](https://developer.blender.org/D8237), contributed by Apple initially. I thought they were used for orthogonality at first glance. We don't have to introduce them in SSE2NEON.

jserv:
Drop this pull request in favor of #84.

PR #87, opened 2020-07-25
Added _mm_round/ceil/floor_ps with tests, plus _mm_blendv_ps

JishinMaster:
This pull request is the second split of the original #83.
I have added the required tests.


sse2neon.h
```diff
@@ -3453,6 +3461,79 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
...
+}
+
+// Round the packed single-precision (32-bit) floating-point elements in a using
+// the rounding parameter, and store the results as packed single-precision
+// floating-point elements in dst.
+// software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_round_ps&expand=4757
+FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
+{
+#if defined(__aarch64__)
+    if (rounding == (_MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC)) {
```
jserv, line 3481: How about rewriting with switch-case statements? Thus, it can benefit from lookup table optimizations of compilers.

sse2neon.h
```diff
@@ -3453,6 +3461,79 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
...
+{
+    return vreinterpretq_m128_f32(vbslq_f32(vreinterpretq_u32_m128(mask),
+                                            vreinterpretq_f32_m128(b),
+                                            vreinterpretq_f32_m128(a)));
+}
+
+// Round the packed single-precision (32-bit) floating-point elements in a using
+// the rounding parameter, and store the results as packed single-precision
+// floating-point elements in dst.
+// software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_round_ps&expand=4757
```
jserv, line 3477: You can strip `&expand=4757` from Intel url.

sse2neon.h
```diff
@@ -3453,6 +3461,79 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
...
+                      roundf(v_float[2]), roundf(v_float[3])};
+        return ret;
+    }
+#endif
+}
+
+// Round the packed single-precision (32-bit) floating-point elements in a up to
+// an integer value, and store the results as packed single-precision
+// floating-point elements in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_ceil_p&expand=4757,668
```
jserv, line 3522: Ditto. Strip `&expand=4757,668` to shorten the URL.

sse2neon.h
```diff
@@ -3453,6 +3461,79 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
...
+FORCE_INLINE __m128 _mm_ceil_ps(__m128 a)
+{
+    return _mm_round_ps(a, _MM_FROUND_TO_POS_INF | _MM_FROUND_NO_EXC);
+}
+
+// Round the packed single-precision (32-bit) floating-point elements in a down
+// to an integer value, and store the results as packed single-precision
+// floating-point elements in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_flo&expand=4757,668,2528
```
jserv, line 3531: Ditto. Shorten URL.

jserv:
Related: #66

sse2neon.h
```diff
@@ -3453,6 +3461,80 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
 }
 #endif
 
+// Blend packed single-precision (32-bit) floating-point elements from a and b
+// using mask, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=blendv_ps
```
marktwtn, line 3466: Replace `blendv_ps` with `mm_blendv_ps`

sse2neon.h
```diff
@@ -3453,6 +3461,80 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
...
+        return (__m128){roundf(v_float[0]), roundf(v_float[1]),
+                        roundf(v_float[2]), roundf(v_float[3])};
+    }
+#endif
+}
+
+// Round the packed single-precision (32-bit) floating-point elements in a up to
+// an integer value, and store the results as packed single-precision
+// floating-point elements in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_ceil_p
```
marktwtn, line 3523: `mm_ceil_p` -> `mm_ceil_ps`

sse2neon.h
```diff
@@ -3453,6 +3461,80 @@ FORCE_INLINE __m128i _mm_castpd_si128(__m128d a)
...
+FORCE_INLINE __m128 _mm_ceil_ps(__m128 a)
+{
+    return _mm_round_ps(a, _MM_FROUND_TO_POS_INF | _MM_FROUND_NO_EXC);
+}
+
+// Round the packed single-precision (32-bit) floating-point elements in a down
+// to an integer value, and store the results as packed single-precision
+// floating-point elements in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_flo
```
marktwtn, line 3532: `mm_flo` -> `mm_floor_ps`

tests/impl.h
```diff
@@ -163,6 +163,10 @@ enum InstructionTest {
     IT_MM_MULLO_EPI32,  // Unimplemented
     IT_MM_TEST_ALL_ZEROS,
     IT_MM_TESTZ_SI128,
+    IT_MM_CEIL_PS,
+    IT_MM_FLOOR_PS,
+    IT_MM_ROUND_PS,
+    IT_MM_BLENDV_PS,
```
marktwtn, line 169: Please insert them to the `SSE4.1` instruction class with alphabetical order. Thanks~

tests/impl.cpp
```diff
@@ -3571,6 +3619,18 @@ class SSE2NEONTestImpl : public SSE2NEONTest
...
+            ret = test_mm_ceil_ps(mTestFloatPointer1);
+            break;
+        case IT_MM_FLOOR_PS:
+            ret = test_mm_floor_ps(mTestFloatPointer1);
+            break;
+        case IT_MM_ROUND_PS:
+            ret = test_mm_round_ps(mTestFloatPointer1);
+            break;
+        case IT_MM_BLENDV_PS:
+            ret = true;  // test_mm_blendv_ps(mTestFloatPointer1);
```
marktwtn, line 3632: Why comment out the testing?

marktwtn: You can remove the comment and add the `// Unimplemented` right after the instruction in `tests/impl.h`.

tests/impl.h
marktwtn, line 169: Add `// Unimplemented`.

tests/impl.cpp
```diff
@@ -3000,6 +3011,43 @@ bool test_mm_crc32_u64(uint64_t crc, uint64_t v)
...
+bool test_mm_ceil_ps(const float *_a)
+{
+    float dx = ceilf(_a[0]);
+    float dy = ceilf(_a[1]);
+    float dz = ceilf(_a[2]);
+    float dw = ceilf(_a[3]);
+
+    __m128 a = test_mm_load_ps(_a);
+    __m128 c = _mm_ceil_ps(a);
+    return validateFloatEpsilon(c, dx, dy, dz, dw, 300.0f);
```
marktwtn, line 3024: I wonder what epsilon value should be set.
Wouldn't `300.0f` be too large?
@jserv any idea?

jserv: At present, `validateFloatEpsilon` appears only once. `300.0f` is definitely too rough. We can calculate via the definition of ceil operation.

tests/impl.h
```diff
@@ -163,6 +163,11 @@ enum InstructionTest {
     IT_MM_MULLO_EPI32,  // Unimplemented
     IT_MM_TEST_ALL_ZEROS,
     IT_MM_TESTZ_SI128,
+    IT_MM_BLENDV_PS, // Unimplemented
+    IT_MM_CEIL_PS,
+    IT_MM_FLOOR_PS,
+    IT_MM_ROUND_PS,
+
```
marktwtn, line 170: Remove the blank line.

tests/impl.h
```diff
@@ -163,6 +163,11 @@ enum InstructionTest {
     IT_MM_MULLO_EPI32,  // Unimplemented
     IT_MM_TEST_ALL_ZEROS,
     IT_MM_TESTZ_SI128,
+    IT_MM_BLENDV_PS, // Unimplemented
```
marktwtn, line 166: The alphabetical order shall include all the `SSE4.1` class instruction.
For example
```c
IT_MM_BLENDV_PS, // Unimplemented
IT_MM_CEIL_PS,
IT_MM_CMPEQ_EPI64,
IT_MM_FLOOR_PS,
...
```

tests/impl.cpp
JishinMaster: For floating point numbers, a == b is risky due to the way they are encoded. a == (b + epsilon) is safer.
Functions such as cos, sin, tan, etc, are implemented using Taylor expansion, hence some shortcut might give different values. The same trouble goes with fused multiple add functions, which are often less precise than separate multiply plus add.

The impl.cpp uses this as input vectors :
`        for (uint32_t i = 0; i < MAX_TEST_VALUE; i++) {
            mTestFloats[i] = ranf(-100000, 100000);
            mTestInts[i] = (int32_t) ranf(-100000, 100000);
        }`

In my opinion, floating point numbers are great for describing small values, and not so much for large ones. I tend to manipulate small values and compute the L2 error, for which something between 10^-5 and 10^-7 is a good result.

Without changing the test vector, I lowered the epsilon to 5.0f.

I did not implement any test for _mm_blendv_ps yet, but the function is used for _mm_round_ps in AArch32 mode.

jserv:

Issue #82, opened 2020-07-21
Strategy on non-SSE intrinsics


jserv:
`sse2neon` aims to support SSE, SSE2, SSE3, SSSE3, SSE4.1, SSE4.2 and AES extension, and AVX intrinsics would be excluded. Later, we might collaborate with [AvxToNeon](https://github.com/kunpengcompute/AvxToNeon) as the fundamental support on SSE. 

@danlark1 pointed out:
> Technically speaking, `_mm_fmadd_ps` is not an SSE extension, this was introduced with fma extension which took place even after AVX.

We do need to think of the strategy on non-SSE intrinsics to ease the platform transition efforts.

jasonliu--:
Greetings, I'm the person who [packaged sse2neon](https://ports.macports.org/port/sse2neon/) for the [MacPorts](https://www.macports.org/) package management system. Wouldn't it make more sense to collaborate with the [SIMDe](https://github.com/simd-everywhere/simde) project? Their `README` claims that they are already incorporating your `sse2neon.h` header [directly into their code base](https://github.com/simd-everywhere/simde/blob/master/simde/x86/sse.h), and they already also have header files that provide support for AVX, FMA, and lots of other intrinsics.

danlark1:
The only downside of SIMDe project I see is the inability to do drop-in replacements as all intrinsics are are prefixed with `simde_`, I know the users which are much more willing to update headers for all their code and dependencies rather than updating all call sites. If we can collaborate with simde on that, that would be great

jserv:
> Greetings, I'm the person who [packaged sse2neon](https://ports.macports.org/port/sse2neon/) for the [MacPorts](https://www.macports.org/) package management system. Wouldn't it make more sense to collaborate with the [SIMDe](https://github.com/simd-everywhere/simde) project? Their `README` claims that they are already incorporating your `sse2neon.h` header [directly into their code base](https://github.com/simd-everywhere/simde/blob/master/simde/x86/sse.h), and they already also have header files that provide support for AVX, FMA, and lots of other intrinsics.

`simde` already merged most parts of `SSE2NEON` efforts. See [simde #499](https://github.com/simd-everywhere/simde/issues/499) for details. Here, we focus on Arm/Aarch64 specific tweaks, which can eventually get merged into `simde`.

jasonliu--:
> The only downside of SIMDe project I see is the inability to do drop-in replacements as all intrinsics are are prefixed with `simde_`

[...]

PR #89, opened 2020-07-26
Added more FMA functions with tests

JishinMaster:
This pull request is the last split of the original #83.

I have added fmsub_ps, fmnadd_ps, fmnsub_ps, and the associated tests.
I had to lower the input vector in the test functions and use an epsilon of 0.0001f because those functions are less precise than just using mul and add, and they seem to diverge with big float numbers (I use qemu, it might be worth checking on real hardware).


jserv:
Non-SSE intrinsics are considered as the additional work to SSE2NEON as #82 stated.

sse2neon.h
```diff
@@ -1307,7 +1307,7 @@ FORCE_INLINE __m128i _mm_shuffle_epi_3332(__m128i a)
 
 // Shuffle packed 8-bit integers in a according to shuffle control mask in the
 // corresponding 8-bit element of b, and store the results in dst.
-// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_epi8
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_epi8&expand=5146
```
marktwtn, line 1310: Revert it.

sse2neon.h
```diff
@@ -2003,7 +2003,7 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
 
 // Compute the bitwise AND of 128 bits (representing integer data) in a and
 // mask, and return 1 if the result is zero, otherwise return 0.
-// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_test_all_zeros
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_test_all_zeros&expand=5871
```
marktwtn, line 2006: Ditto.

sse2neon.h
```diff
@@ -2510,11 +2510,46 @@ FORCE_INLINE __m128 _mm_fmadd_ps(__m128 a, __m128 b, __m128 c)
 #endif
 }
 
+// Computes the fused multiple add product of 32-bit floating point numbers.
+//
+// Return Value
+// Multiplies A and B, and adds C to the temporary result before returning it.
```
marktwtn, line 2516: `adds` -> `subtracts`

marktwtn, line 2516: Shall we unify these descriptions of `FMA` functions with the link of `Intel Intrinsics Guide`?
For example:
```
Multiply packed single-precision (32-bit) floating-point elements in a and b,
subtract packed elements in c from the intermediate result, and store the results in dst.
```

jserv:
Header `<immintrin.h>` is used for AVX, AVX2 and FMA instructions. We need to figure out the way to manipulate. Non-SSE intrinsics are considered as the additional work to SSE2NEON as #82 stated.

JishinMaster:
Can we use something like "ifdef USE_FMA"‚  and enable only for those who want it?
Or use a completely different "sse2neon_optional.h"  file ? Doing so we could also implement some MMX instructions from older programs which have not migrated to SSE2.

jserv:
Let's close this pull request in favor of #82.

PR #90, opened 2020-07-27
Implement _mm_stream_ps and _mm_stream_load_si128

jserv:


sse2neon.h
```diff
@@ -4591,9 +4591,17 @@ FORCE_INLINE void _mm_sfence(void)
     __sync_synchronize();
 }
 
+// Store 128-bits (composed of 4 packed single-precision (32-bit) floating-
+// point elements) from a into memory using a non-temporal memory hint.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_stream_ps
+void _mm_stream_ps(float *p, __m128 a)
+{
+    vst1q_f32(p, vreinterpretq_f32_m128(a));
```
marktwtn, line 4599: Shall we add the code like
```c
#if __has_builtin(__builtin_nontemporal_store)
    __builtin_nontemporal_store(a, (__m128 *) p);
#else 
    vst1q_f32(p, vreinterpretq_f32_m128(a));
#endif
```

sse2neon.h
```diff
@@ -4591,9 +4591,17 @@ FORCE_INLINE void _mm_sfence(void)
     __sync_synchronize();
 }
 
+// Store 128-bits (composed of 4 packed single-precision (32-bit) floating-
+// point elements) from a into memory using a non-temporal memory hint.
```
marktwtn, line 4595: I wonder why the `mem_addr must be aligned on a 16-byte boundary` related description are ignored?

jserv: My intention was to extract the first sentence from Intel/Microsoft online document. And no further sentences should be extracted for consistency.

marktwtn: Not sure which one is better.
As I check the `sse2neon.h`, it seems like most of the comments of intrinsics are following the rule like extract the first sentence, and few of them are more than one sentence.

sse2neon.h
jserv: You are right. LLVM specifies `__builtin_nontemporal_store` in intrinsics header: https://github.com/llvm/llvm-project/blob/master/clang/lib/Headers/xmmintrin.h#L2138
```c
static __inline__ void __DEFAULT_FN_ATTRS
_mm_stream_ps(float *__p, __m128 __a)
{
  __builtin_nontemporal_store((__v4sf)__a, (__v4sf*)__p);
}
```

jserv:
Related: #66

PR #92, opened 2020-07-27
Implement _mm_dp_ps

jserv:
Intrinsic "_mm_dp_ps" performs conditional multiplication for
single-precision floating-point elements, sum the four
products and conditionally store the sum. To improve the accuracy
of floating-point summation, Kahan algorithm is used for each
operation. This patch also introduced AArch64-friendly shortcuts for
certain inputs.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -2900,6 +2900,60 @@ FORCE_INLINE __m128i _mm_hsub_epi32(__m128i _a, __m128i _b)
     return vreinterpretq_m128i_s32(vsubq_s32(ab02, ab13));
 }
 
+// Kahan summation for accurate summation of floating-point numbers.
+// http://blog.zachbjornson.com/2019/08/11/fast-float-summation.html
+FORCE_INLINE void sse2neon_kadd_f32(float *sum, float *c, float y)
+{
+    y -= *c;
+    float t = *sum + y;
+    *c = (t - *sum) - y;
+    *sum = t;
+}
+
+// Conditionally multiply the packed single-precision (32-bit) floating-point
+// elements in a and b using the high 4 bits in imm8, sum the four products,
+// and conditionally store the sum in dst using the low 4 bits of imm.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_dp_ps
+FORCE_INLINE __m128 _mm_dp_ps(__m128 a, __m128 b, const int imm)
+{
+#if defined(__aarch64__)
+    /* shortcuts */
+    if (imm == 0xFF) {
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -2900,6 +2900,59 @@ FORCE_INLINE __m128i _mm_hsub_epi32(__m128i _a, __m128i _b)
...
+     * is used for each operation.
+     */
+    if (imm & (1 << 4))
+        sse2neon_kadd_f32(&s, &c, f32a[0] * f32b[0]);
+    if (imm & (1 << 5))
+        sse2neon_kadd_f32(&s, &c, f32a[1] * f32b[1]);
+    if (imm & (1 << 6))
+        sse2neon_kadd_f32(&s, &c, f32a[2] * f32b[2]);
+    if (imm & (1 << 7))
+        sse2neon_kadd_f32(&s, &c, f32a[3] * f32b[3]);
```
marktwtn, line 2945: In the reference of http://blog.zachbjornson.com/2019/08/11/fast-float-summation.html,
it would do the final sum calculation `s + c`, which is missing here?

tests/impl.cpp
```diff
@@ -1540,6 +1543,25 @@ bool test_mm_mul_ps(const float *_a, const float *_b)
...
+    const int imm = 0x3F;
+    __m128 a = test_mm_load_ps(_a);
+    __m128 b = test_mm_load_ps(_b);
+    __m128 out = _mm_dp_ps(a, b, imm);
+
+    float r[4]; /* the reference */
+    float sum = 0;
+
+    for (size_t i = 0; i < 4; i++)
+        sum += ((imm >> (i + 4)) & 1) ? _a[i] * _b[i] : 0;
```
marktwtn, line 1557: Why not using the code like `(imm) & (1 << (i + 4))`?

jserv:
Related: #66

tests/impl.cpp
```diff
@@ -1540,6 +1543,25 @@ bool test_mm_mul_ps(const float *_a, const float *_b)
     return validateFloat(c, dx, dy, dz, dw);
 }
 
+bool test_mm_dp_ps(const float *_a, const float *_b)
+{
+    const int imm = 0xFF;
```
marktwtn, line 1548: Why changing the value from `0x3F` to `0xFF`?

jserv: I wanted to validate the special case `0xFF`. Since the last parameter of `_mm_dp_ps` is `imm8`, only constants can be passed, that prevents me from writing test cases with better coverage though.

jserv: Each occurrence of `int imm` in `sse2neon.h` has the same restriction to comply with Intel SSE intrinsics.

PR #98, opened 2020-07-28
Drop incorrect _mm_srai_epi32 implementation

jserv:
There were two _mm_srai_epi32 implementations: one was the inline function; another was the macro. According to Intel manual, the operation of intrinsic _mm_srai_epi32 is illustrated as following:
    
    __m128i _mm_srai_epi32 (__m128i a, int imm8)
    FOR j := 0 to 3
            i := j*32
            IF imm8[7:0] > 31
                    dst[i+31:i] := (a[i+31] ? 0xFFFFFFFF : 0x0)
            ELSE
                    dst[i+31:i] := SignExtend32(a[i+31:i] >> imm8[7:0])
            FI
    ENDFOR
    
The original inlined function would produce different behavior, and this patch drops it for the sake of correctness.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1592,17 +1592,6 @@ FORCE_INLINE __m128i _mm_blendv_epi8(__m128i _a, __m128i _b, __m128i _mask)
 
 /* Shifts */
 
-// Shifts the 4 signed 32-bit integers in a right by count bits while shifting
-// in the sign bit.
-//
-//   r0 := a0 >> count
-//   r1 := a1 >> count
-//   r2 := a2 >> count
-//   r3 := a3 >> count immediate
-FORCE_INLINE __m128i _mm_srai_epi32(__m128i a, int count)
-{
-    return (__m128i) vshlq_s32((int32x4_t) a, vdupq_n_s32(-count));
-}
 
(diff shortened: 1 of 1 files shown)


marktwtn:
I wonder the reason of removing the function instead of macro.
Could you give me more explanation?

jserv:
> I wonder the reason of removing the function instead of macro.
> Could you give me more explanation?

The subject of this pull request was misleading. The operation of intrinsic [_mm_srai_epi32](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_srai_epi32) is illustrated as following:
```c
__m128i _mm_srai_epi32 (__m128i a, int imm8)
FOR j := 0 to 3
	i := j*32
	IF imm8[7:0] > 31
		dst[i+31:i] := (a[i+31] ? 0xFFFFFFFF : 0x0)
	ELSE
		dst[i+31:i] := SignExtend32(a[i+31:i] >> imm8[7:0])
	FI
ENDFOR
```

The original inlined function would produce different actions and behavior.
```c
FORCE_INLINE __m128i _mm_srai_epi32(__m128i a, int count)                                                                                                     
{
    return (__m128i) vshlq_s32((int32x4_t) a, vdupq_n_s32(-count));
}
```

I will clarify the commit message.

PR #100, opened 2020-08-08
WIP: Add unimplemented result for tests

afcidk:
Some test cases that are unimplemented now is indicated by comments in the source code. I modified the return value from bool to result_t to accept three states: TEST_SUCCESS(1), TEST_FAIL(0), and TEST_UNIMPL(-1).
In this way `main.cpp` can recognize whether this test case is implemented or not.

I've also made some changes that might make the tests look cleaner.
* Some common functions from `impl` files to `common` files.
* The instructionString and enums are also wrapped by macros (it was
separated before this commit). We can have better control of the
implemented instrinsics in this way.


tests/impl.h
```diff
@@ -1,5 +1,273 @@
 #ifndef SSE2NEONTEST_H
 #define SSE2NEONTEST_H
+#include "common.h"
+#define ENUM(c) it_##c,
+#define STR(c) #c,
+#define CASE(c)                   \
+    case it_##c:                  \
+        ret = test_##c(*this, i); \
+        break;
+#define TYPE_FOREACH(TYPE)            \
```
jserv, line 10: Can we classify the following instructions as @marktwtn did for SSE, SSE2, SSE3, etc.?

marktwtn:
Commit message: `recognize3` -> `recognize`

Makefile
```diff
@@ -39,7 +39,8 @@ LDFLAGS	+= -lm
 OBJS = \
     tests/binding.o \
     tests/impl.o \
-    tests/main.o
+    tests/main.o \
+    tests/common.o
```
marktwtn, line 43: Sort with alphabetical order.

tests/common.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+result_t validateDouble(__m128d a, double d0, double d1)
+{
+    const double *t = (const double *) &a;
+    ASSERT_RETURN(validateSingleDoublePair(t[0], d0));
+    ASSERT_RETURN(validateSingleDoublePair(t[1], d1));
+    return TEST_SUCCESS;
+}
+
+}  // namespace SSE2NEON
```
marktwtn, line 229: Fix `No new line at end of file`.

tests/common.h
```diff
@@ -0,0 +1,136 @@
...
+result_t validateFloatEpsilon(__m128 a,
+                              float f0,
+                              float f1,
+                              float f2,
+                              float f3,
+                              float epsilon);
+result_t validateDouble(__m128d a, double d0, double d1);
+}  // namespace SSE2NEON
+
+#endif
```
marktwtn, line 136: Ditto.

tests/main.cpp
```diff
@@ -7,26 +7,26 @@ int main(int /*argc*/, const char ** /*argv*/)
     SSE2NEON::SSE2NEONTest *test = SSE2NEON::SSE2NEONTest::create();
     uint32_t passCount = 0;
     uint32_t failedCount = 0;
+    uint32_t ignoreCount = 0;
     for (uint32_t i = 0; i < SSE2NEON::IT_LAST; i++) {
         SSE2NEON::InstructionTest it = SSE2NEON::InstructionTest(i);
-        printf("Running Test %s\n",
-               SSE2NEON::SSE2NEONTest::getInstructionTestString(it));
-        bool ok = test->runTest(it);
+        SSE2NEON::result_t ok = test->runTest(it);
```
marktwtn, line 13: I think we can change the variable naming from `ok` to `ret` or `res`.

tests/impl.cpp
```diff
@@ -948,134 +263,159 @@ bool test_mm_set_epi8(const int8_t *_a)
...
     ASSERT_RETURN(p[0] == w);
     ASSERT_RETURN(p[1] == z);
     ASSERT_RETURN(p[2] == y);
     ASSERT_RETURN(p[3] == x);
-    return true;
+    return TEST_SUCCESS;
 }
 
-bool test_mm_store_ps(int32_t *p, int32_t x, int32_t y, int32_t z, int32_t w)
+result_t do_mm_store_ps(int32_t *p, int32_t x, int32_t y, int32_t z, int32_t w)
```
marktwtn, line 301: These tests are not modified because of?

tests/impl.cpp
```diff
@@ -3075,1199 +2601,1106 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+}
 
 SSE2NEONTest *SSE2NEONTest::create(void)
 {
     SSE2NEONTestImpl *st = new SSE2NEONTestImpl;
     return static_cast<SSE2NEONTest *>(st);
 }
-
-}  // namespace SSE2NEON
+}  // namespace SSE2NEON
```
marktwtn, line 3706: Fix `No new line at end of file`.

tests/impl.h
```diff
@@ -10,222 +293,23 @@ namespace SSE2NEON
...
+    // Passed: TEST_SUCCESS (1)
+    // Failed: TEST_FAIL (0)
+    // Unimplemented: TEST_UNIMPL (-1)
+    virtual result_t runTest(InstructionTest test) = 0;
     virtual void release(void) = 0;
 };
 
 }  // namespace SSE2NEON
 
+
```
marktwtn, line 314: Remove the blank line.

tests/common.cpp
```diff
@@ -0,0 +1,229 @@
+#include "common.h"
+#include <cmath>
+#include <cstdint>
+
+namespace SSE2NEON
+{
+int32_t nan = ~0;
+int64_t nan64 = ~0;
+
+result_t validate128(__m128i a, __m128i b)
```
marktwtn, line 10: I am not sure about returning the `result_t` is a good idea since there is no way that it would return `TEST_UNIMPL`.

afcidk: I changed the type to `result_t` since we could return the `validateXX` function directly in test functions.

tests/impl.cpp
afcidk: That was my fault :b. 
My intention to change some `test_xxx` functions to `do_xxx` functions is because the return value of these functions are not `result_t`, which should not be executed as test cases. (e.g. `__m128 do_mm_load_ps`). For these functions, we should another `test_xxx` functions that returns `result_t`

afcidk: Apparently, these two functions should keep the `test` prefix rather than the `do` prefix.

Makefile
```diff
@@ -38,8 +38,9 @@ CXXFLAGS += -Wall -Wcast-qual -I. $(ARCH_CFLAGS) -std=gnu++14
 LDFLAGS	+= -lm
 OBJS = \
     tests/binding.o \
+    tests/common.o \
     tests/impl.o \
-    tests/main.o
+    tests/main.o
```
marktwtn, line 43: Remove the space after `main.o`.

tests/impl.cpp
```diff
@@ -98,835 +190,56 @@ static inline float bankersRounding(float val)
...
     return validateFloat(a, w, w, w, w);
 }
 
-bool test_mm_set_ps(float x, float y, float z, float w)
+result_t test_mm_set_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+    float x = impl.mTestFloats[0];
+    float y = impl.mTestFloats[1];
+    float z = impl.mTestFloats[2];
+    float w = impl.mTestFloats[3];
```
marktwtn, line 228: I think this should be replaced with `impl.mTestFloats[i]` or `impl.mTestFloatPointer1[0]`.
Otherwise, it would keep using the same value.

tests/impl.cpp
```diff
@@ -950,134 +263,163 @@ bool test_mm_set_epi8(const int8_t *_a)
...
-__m128i test_mm_set_epi32(int32_t x, int32_t y, int32_t z, int32_t w)
+__m128i do_mm_set_epi32(int32_t x, int32_t y, int32_t z, int32_t w)
 {
     __m128i a = _mm_set_epi32(x, y, z, w);
     validateInt32(a, w, z, y, x);
     return a;
 }
 
-bool test_mm_store_ps(float *p, float x, float y, float z, float w)
+result_t test_mm_store_ps(float *p, float x, float y, float z, float w)
```
marktwtn, line 290: Unify the parameters as the others.

marktwtn:
We still have a lot of testing functions with `do` prefix in `test/impl.cpp`.
What is the rule for using `do` or `test` prefix?
I think you should clarify in the comment and make sure they are unified.

afcidk:
> We still have a lot of testing functions with `do` prefix in `test/impl.cpp`.
> What is the rule for using `do` or `test` prefix?
> I think you should clarify in the comment and make sure they are unified.
> 
> For exmaple, we have three different `test_mm_store_ps` and we have two `do_mm_load_ps` without any `test_mm_load_ps`.

Functions with `test` prefix will be called in `runSingleTest` while the `do` prefixed functions do not.
I think I have to rename two `test_mm_store_ps` to `do_mm_store_ps` and leave only one `test_mm_store_ps`.

We have `test_mm_load_ps` in `impl.cpp:2641`, but I left it to be unimplemented in this commit.

tests/impl.h
```diff
@@ -1,5 +1,288 @@
 #ifndef SSE2NEONTEST_H
 #define SSE2NEONTEST_H
+#include "common.h"
+#define ENUM(c) it_##c,
+#define STR(c) #c,
+#define CASE(c)                   \
+    case it_##c:                  \
+        ret = test_##c(*this, i); \
+        break;
+#define TYPE_FOREACH(TYPE)            \
```
jserv, line 10: I am not sure if `TYPE` is a meaningful name which refers to intrinsics. @marktwtn, how do you think?

marktwtn: How about `INTRINSICS`?

jserv: I think `INTRIN` is reasonable as an abbreviation.

marktwtn:
> Functions with `test` prefix will be called in `runSingleTest` while the `do` prefixed functions do not.

Describe it with the comment in the source code.

afcidk:
> > Functions with `test` prefix will be called in `runSingleTest` while the `do` prefixed functions do not.
> 
> Describe it with the comment in the source code.

I've mentioned [here](https://github.com/DLTcollab/sse2neon/blob/6bab7e8969c35e40b51336eb6908388428084ddc/tests/impl.cpp#L17) already, should it be commented before every functions?

marktwtn:
> I've mentioned [here](https://github.com/DLTcollab/sse2neon/blob/6bab7e8969c35e40b51336eb6908388428084ddc/tests/impl.cpp#L17) already, should it be commented before every functions?

I did not notice that.
However, I think it might be better to place it right before the first `do_xx` function?

tests/impl.cpp
```diff
@@ -98,835 +192,110 @@ static inline float bankersRounding(float val)
...
+{
+    __m128 a = _mm_load_ps((const float *) p);
+    __m128i ia = *(const __m128i *) &a;
+    validateInt32(ia, p[0], p[1], p[2], p[3]);
+    return ia;
+}
 
-static float ranf(void)
+// This function is not called from `runSingleTest`, but for other intrinsic
+// tests that might need to call `_mm_load_ps`.
```
marktwtn, line 235: `_mm_store_ps`

tests/impl.cpp
```diff
@@ -950,134 +319,115 @@ bool test_mm_set_epi8(const int8_t *_a)
...
+    return validateInt32(a, x, x, x, x);
 }
 
-bool testret_mm_set_epi32(int32_t x, int32_t y, int32_t z, int32_t w)
+result_t test_mm_set_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+    int32_t x = impl.mTestInts[0];
+    int32_t y = impl.mTestInts[1];
+    int32_t z = impl.mTestInts[2];
+    int32_t w = impl.mTestInts[3];
```
marktwtn, line 334: Use `i` to avoid unchanged value.

README.md
```diff
@@ -93,42 +93,23 @@ Once the conversion is implemented, the test can be added with the following ste
 
 * File `tests/impl.h`
 
-  Add the intrinsic in `enum InstructionTest`. The naming convention should be `IT_MM_XXX`.
-  And place it in the correct classification with the alphabetical order.
+  Add the intrinsic under `#define TYPE_FOREACH(TYPE)` macro. The naming convention
```
jserv, line 96: Adjust the wording around `TYPE_FOREACH`.

tests/common.cpp
```diff
@@ -0,0 +1,229 @@
+#include "common.h"
+#include <cmath>
+#include <cstdint>
+
+namespace SSE2NEON
+{
+int32_t nan = ~0;
```
jserv, line 7: How about renaming `nan` to `NaN`? Thus, it can avoid potential symbol conflicts with [nan(3)](https://man7.org/linux/man-pages/man3/nan.3.html).

tests/common.cpp
```diff
@@ -0,0 +1,229 @@
...
+int64_t nan64 = ~0;
```
jserv, line 8: Ditto. `NaN64`

tests/impl.cpp
```diff
@@ -3090,1196 +2627,1097 @@ bool test_mm_test_all_zeros(const int32_t *_a, const int32_t *_mask)
...
+}
+
+result_t SSE2NEONTestImpl::runSingleTest(InstructionTest test, uint32_t i)
+{
+    result_t ret = TEST_SUCCESS;
+
+    switch (test) {
+        INTRIN_FOREACH(CASE)
+    case IT_LAST: /* should not happend */
+        break;
```
marktwtn, line 3716: We should add the indention here?

tests/impl.h
```diff
@@ -1,5 +1,288 @@
...
+    TYPE(mm_popcnt_u32)               \
+    TYPE(mm_popcnt_u64)               \
+    TYPE(mm_shuffle_epi32_splat)      \
+    TYPE(mm_shuffle_epi32)            \
+    TYPE(mm_shuffle_epi32_default)    \
+    TYPE(mm_shufflehi_epi16)          \
+    TYPE(mm_shufflehi_epi16_function) \
+    TYPE(mm_shufflelo_epi16)          \
+    TYPE(mm_shufflelo_epi16_function)
+
```
marktwtn, line 285: Remove the blank line.

PR #101, opened 2020-08-12
feat: Implement _mm_cvt_ss2si

### LINKED ISSUE #76: Missing _mm_cvt_ss2si
Hi the following instruction is missing: _mm_cvt_ss2si

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=4979,3330,3863,4979,5197,3853,3330,3863,5197,3407,3863,5610,1382&text=_mm_cvt_ss2si

Thanks,
Alex

marktwtn:
Close #76.


tests/impl.cpp
```diff
@@ -2242,6 +2245,15 @@ bool test_mm_hadd_epi16(const int16_t *_a, const int16_t *_b)
     return validateInt16(ret, d0, d1, d2, d3, d4, d5, d6, d7);
 }
 
+bool test_mm_cvt_ss2si(const float *_a)
+{
+    int32_t d0 = (int32_t) _a[0];
+
+    __m128 a = test_mm_load_ps(_a);
+    int32_t ret = _mm_cvt_ss2si(a);
+    return abs(ret - d0) <= 1;
```
jserv, line 2254: I am not sure wether if `<= 1` makes sense or not. Can you explain?

marktwtn: Take floating point `12.5` as an example.

When calling the converted `_mm_cvt_ss2si`, it would be following the rule `Round to nearest, ties away from zero`, which would make `12.5` become `13`.

However, the C type casting has different behaviour on `ARM` and `x86` architecture.
We do not know which rule they are following.
Hence, the worst case would be converting the `12.5` to `12` with C type casting.

Or we can use `floor()` or `ceil()` to avoid using  `abs(ret - d0) <= 1`.

jserv: Let's use rounding functions such as `floor()` and `ceil()`.

marktwtn: It turns out that there are some behaviour I do not understand.
When running the test suite on my computer, there are several results which are out of my expectation.
Take `_mm_cvt_ss2si` as an example:
`73339.843750` -> `73339` (expected: `73340`)
`9362.796875` -> `9362` (expected: `9363`)
`-75634.765625` -> `75634` (expected: `75635`)
`-98614.500000  ` -> `-98614` (expected: `-98615`)

I am not sure whether this is caused by the `QEMU` or not.

jserv: How about A1 (aarch64) instance provided by AWS EC2?

PR #102, opened 2020-08-16
Add test for _mm_abs_epi32

HowJMay:


tests/impl.cpp
```diff
@@ -1381,6 +1383,19 @@ bool test_mm_sub_epi64(const int64_t *_a, const int64_t *_b)
     return validateInt64(c, d0, d1);
 }
 
+bool test_mm_abs_epi32(const int32_t *_a)
+{
+    __m128i a = test_mm_load_ps((const int32_t *) _a);
```
marktwtn, line 1388: The casting is not necessary.

tests/impl.h
```diff
@@ -12,6 +12,7 @@ namespace SSE2NEON
 // are compared to the NEON approximate version.
 enum InstructionTest {
     // SSE
+    IT_MM_ABS_EPI32,
```
marktwtn, line 15: The function should be placed under the `SSSE3` class.

tests/impl.h
```diff
@@ -175,6 +175,7 @@ enum InstructionTest {
     IT_MM_HADD_EPI16,
     IT_MM_MADDUBS_EPI16,
     IT_MM_SHUFFLE_EPI8,
+    IT_MM_ABS_EPI32,
```
marktwtn, line 178: Place it with alphabetical order.

PR #115, opened 2020-08-25
chore: Add cross-compiling tools into gitignore

HowJMay:
Bash script '.ci/cross-tool.sh' would download cross-compiling tools
for testing. However, these tools are not included in gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.o
 tests/*.d
 tests/main
+gcc-arm-*


HowJMay:
@marktwtn How about this way?

.gitignore
```diff
@@ -1,3 +1,5 @@
 *.o
 tests/*.d
 tests/main
+gcc-arm-*-x86_64-aarch64-none-linux-gnu/
+gcc-arm-*-x86_64-arm-none-linux-gnueabihf/
```
marktwtn, line 5: I think we can just use `gcc-arm-*`?

HowJMay: sure

jserv (approved):

marktwtn (approved):

PR #116, opened 2020-08-26
feat: Implement _mm_test_all_ones

HowJMay:


sse2neon.h
```diff
@@ -2023,6 +2023,23 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
...
 }
 
+// Compute the bitwise NOT of a and then AND with a 128-bit vector containing
+// all 1's, and return 1 if the result is zero, otherwise return 0.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_test_all_ones
+FORCE_INLINE int _mm_test_all_ones(__m128i a)
+{
+#define ONES32 0xffffffff
+    uint32_t ALIGN_STRUCT(16) data[4] = {ONES32, ONES32, ONES32, ONES32};
+    int32x4_t a_and_complement =
```
jserv, line 2033: You can reimplement with `vgetq_lane_s64`. i.e, something like
```c
return (vgetq_lane_s64(a, 0) & vgetq_lane_s64(a, 1)) == ~(uint64_t) 0;
```

sse2neon.h
```diff
@@ -2023,6 +2023,23 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
...
+        vbicq_s32(vld1q_s32((int32_t *) data), vreinterpretq_s32_m128i(a));
```
marktwtn, line 2034: You can replace data with `vdup_n_s32`.

sse2neon.h
```diff
@@ -2023,6 +2023,23 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
 #endif
 }
 
+// Compute the bitwise NOT of a and then AND with a 128-bit vector containing
+// all 1's, and return 1 if the result is zero, otherwise return 0.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_test_all_ones
+FORCE_INLINE int _mm_test_all_ones(__m128i a)
+{
+#define ONES32 0xffffffff
```
marktwtn, line 2031: I am not sure defining a macro here is a good idea or not.

HowJMay: I am going to use ` ~(uint64_t) 0` to replace the macro

tests/impl.cpp
```diff
@@ -2661,6 +2661,23 @@ result_t test_mm_minpos_epu16(const SSE2NEONTestImpl &impl, uint32_t i)
     return validateUInt16(ret, d0, d1, d2, d3, d4, d5, d6, d7);
 }
 
+result_t test_mm_test_all_ones(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+#define ONES32 0xffffffff
```
marktwtn, line 2666: I think this macro should be removed.

sse2neon.h
```diff
@@ -2023,6 +2023,15 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
 #endif
 }
 
+// Check if input 'a' is all 1's, and return 0 if the 'a' is all 1's, otherwise
+// return 1.
```
marktwtn, line 2027: You should provide the description of [_mm_test_all_ones](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_test_all_ones).

sse2neon.h
```diff
@@ -2023,6 +2023,15 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
...
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_test_all_ones
+FORCE_INLINE int _mm_test_all_ones(__m128i a)
+{
+    return (uint64_t)(vgetq_lane_s64(a, 0) & vgetq_lane_s64(a, 1)) ==
```
marktwtn, line 2031: Why doing the `(uint64_t)` casting?

sse2neon.h
HowJMay: I will replace the current one with the description in Intel Intrinsics page

sse2neon.h
```diff
@@ -2023,6 +2023,15 @@ FORCE_INLINE int _mm_movemask_ps(__m128 a)
 #endif
 }
 
+// Check if input 'a' is all 1's, and return 0 if the 'a' is all 1's, otherwise
```
jserv, line 2026: Quote from Intel Intrinsic website:
> Compute the bitwise NOT of a and then AND with a 128-bit vector containing all 1's, and return 1 if the result is zero, otherwise return 0.

marktwtn (approved):

PR #119, opened 2020-08-28
Implement _mm_hadd_pi32

HowJMay:
Add overloaded function, 'validateInt32' for testing datatype '__m64'.

### DIFF
--- a/tests/common.cpp
+++ b/tests/common.cpp
@@ -103,6 +103,14 @@ result_t validateUInt16(__m128i a,
     return TEST_SUCCESS;
 }
 
+result_t validateInt32(__m64 a, int32_t u0, int32_t u1)
+{
+    const int32_t *t = (const int32_t *) &a;
+    ASSERT_RETURN(t[0] == u0);
+    ASSERT_RETURN(t[1] == u1);
+    return TEST_SUCCESS;
+}
+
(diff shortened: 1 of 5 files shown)


tests/common.cpp
```diff
@@ -103,6 +103,14 @@ result_t validateUInt16(__m128i a,
     return TEST_SUCCESS;
 }
 
+result_t validateInt32(__m64 a, int32_t u0, int32_t u1)
```
jserv, line 106: Can you check line 38? There is already one `validateInt32`.

HowJMay: But `validateInt32` at line 38 takes `__m128i` as input argument. The new `validateInt32` implemented here takes `__m64` as input argument.

jserv: It was intended to add an overloading function. You shall mention in git commit message(s) as well.

jserv (changes requested):
Change tests/impl.h for this intrinsic.

jserv:
Function `result_t validateInt32(__m64 a, int32_t u0, int32_t u1)` is an "overloaded" function instead of "reload."
See https://en.wikipedia.org/wiki/Overloading

marktwtn (approved):

Issue #81, opened 2020-07-21
Missing A32 implementations for _mm_store_pd, _mm_storeu_pd, _mm_cvtpd_ps, _mm_cvtps_pd, _mm_castpd_si128


jserv:
The following intrinsics have Aarch64 implementations without A32:
* `_mm_store_pd`
* `_mm_storeu_pd`
* `_mm_cvtpd_ps`
* `_mm_cvtps_pd`
* `_mm_castpd_si128`

We expect to provide corresponding A32 implementations as well.

PR #120, opened 2020-08-28
Implement A32 counterpart of _mm_store_pd

marktwtn:
Related #81.


jserv:
Let's always use the term "A32" instead of "Arm 32-bit," which sounds a bit long-winded.

PR #121, opened 2020-08-28
feat: Implement mm_andnot_pd

HowJMay:


sse2neon.h
```diff
@@ -896,6 +899,20 @@ FORCE_INLINE __m128 _mm_andnot_ps(__m128 a, __m128 b)
...
+//
+//   r0 := ~a0 & b0
+//   r1 := ~a1 & b1
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_andnot_pd
+FORCE_INLINE __m128d _mm_andnot_pd(__m128d a, __m128d b)
+{
+    return vreinterpretq_m128d_s64(
+        vbicq_s64(vreinterpretq_s64_m128d(b),
+                  vreinterpretq_s64_m128d(a)));  // *NOTE* argument swap
```
jserv, line 913: Move the comment before `return` statement.

jserv (approved):

sse2neon.h
```diff
@@ -239,6 +239,9 @@ typedef int64x2_t __m128i; /* 128-bit vector containing integers */
 #define vreinterpretq_f64_m128d(x) (x)
 #else
 #define vreinterpretq_m128d_f32(x) (x)
+#define vreinterpretq_m128d_s64(x) vreinterpretq_f32_s64(x)
+
+#define vreinterpretq_s64_m128d(x) vreinterpretq_s64_f32(x)
```
marktwtn, line 244: Put this line right after `#define vreinterpretq_f32_m128d(x) (x)`.

tests/impl.h
```diff
@@ -13,6 +13,7 @@
     TYPE(mm_add_ss)                   \
     TYPE(mm_and_ps)                   \
     TYPE(mm_andnot_ps)                \
+    TYPE(mm_andnot_pd)                \
```
marktwtn, line 16: It is class `SSE2`.

tests/impl.cpp
```diff
@@ -469,6 +469,27 @@ result_t test_mm_andnot_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    __m128d b = _mm_load_pd(_b);
+    __m128d c = _mm_andnot_pd(a, b);
+    // now for the assertion...
+    const uint64_t *ia = (const uint64_t *) &a;
+    const uint64_t *ib = (const uint64_t *) &b;
+    uint64_t r0 = ~ia[0] & ib[0];
+    uint64_t r1 = ~ia[1] & ib[1];
+    r = validateUInt64(*(const __m128i *) &c, r0, r1);
+
+    return r;
```
marktwtn, line 490: Replace with `return vlalidateXXXX;`.

tests/impl.cpp
```diff
@@ -469,6 +469,27 @@ result_t test_mm_andnot_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+result_t test_mm_andnot_pd(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    result_t r = TEST_FAIL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+
+    __m128d a = _mm_load_pd(_a);
+    __m128d b = _mm_load_pd(_b);
+    __m128d c = _mm_andnot_pd(a, b);
+    // now for the assertion...
```
marktwtn, line 483: The commit message should be more detailed.
As other developers might not be aware that the assertion here means the assertion in `validateXXXX`.

I think you should describe why you do the bitwise operation on `uint64_t` instead of `double` since bitwise operation does not accept `float/double` data type.

sse2neon.h
```diff
@@ -896,6 +899,20 @@ FORCE_INLINE __m128 _mm_andnot_ps(__m128 a, __m128 b)
                   vreinterpretq_s32_m128(a)));  // *NOTE* argument swap
 }
 
+// Compute the bitwise NOT of packed double-precision (64-bit) floating-point
+// elements in a and then AND with b, and store the results in dst.
+//
+//   r0 := ~a0 & b0
+//   r1 := ~a1 & b1
```
marktwtn, line 906: Shall we replace it with the **Operation** on [_mm_andnot_pd](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_andnot_pd)?

@jserv what do you think?

tests/impl.cpp
```diff
@@ -469,6 +469,27 @@ result_t test_mm_andnot_ps(const SSE2NEONTestImpl &impl, uint32_t i)
     return r;
 }
 
+// r0 := ~a0 & b0
+// r1 := ~a1 & b1
```
marktwtn, line 473: I think the comments can be removed.

sse2neon.h
HowJMay: why?

tests/impl.cpp
```diff
@@ -469,6 +469,24 @@ result_t test_mm_andnot_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+{
+    result_t r = TEST_FAIL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+
+    __m128d a = _mm_load_pd(_a);
+    __m128d b = _mm_load_pd(_b);
+    __m128d c = _mm_andnot_pd(a, b);
+
+    // Bitwise operations are not allowed on float/double datatype
```
jserv, line 482: This comment is not exact on bitwise operations. Can you reword?

tests/impl.cpp
```diff
@@ -469,6 +469,24 @@ result_t test_mm_andnot_ps(const SSE2NEONTestImpl &impl, uint32_t i)
     return r;
 }
 
+result_t test_mm_andnot_pd(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    result_t r = TEST_FAIL;
```
marktwtn, line 474: Remove it.

sse2neon.h
marktwtn: Other macros which are the same class are put together.
And you have extra blank line at `245`.

HowJMay: I wonder that can I remove the blank line, line 243, since I think `vreinterpretq_s64_m128d` and `vreinterpretq_m128d_s64` are the same class. However, `vreinterpretq_m128d_f32` is not in the same class of they do.

marktwtn: The class is based on how the other macros do.
It does not mean which one is correct or wrong.
What I ask for is the consistency.

HowJMay: Oh I get the point how do you sort each macro. I thought you sort them depends on the name of macro. However, these macros are sorted depends on the implementation instead.

marktwtn: My previous comments might not consider well.
After consideration, here are two ways I suggest to do:

1. Follow the macros which are defined when A64(aarch64) is used
```c
#define vreinterpretq_m128d_s64(x) vreinterpretq_f32_s64(x)
#define vreinterpretq_m128d_f32(x) (x)

#define vreinterpretq_s64_m128d(x) vreinterpretq_s64_f32(x)
#define vreinterpretq_f32_m128d(x) (x)
```

2. Based on the classification of the other macros.
```c
#define vreinterpretq_m128d_s64(x) vreinterpretq_f32_s64(x)

#define vreinterpretq_m128d_f32(x) (x)

#define vreinterpretq_s64_m128d(x) vreinterpretq_s64_f32(x)

#define vreinterpretq_f32_m128d(x) (x)
```
However, if you change the code this way, you should change the macros like `vreinterpretq_m128d_s64` as well for consistency.

HowJMay: I choose the second way

tests/impl.cpp
```diff
@@ -469,6 +469,25 @@ result_t test_mm_andnot_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+result_t test_mm_andnot_pd(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+
+    __m128d a = _mm_load_pd(_a);
+    __m128d b = _mm_load_pd(_b);
+    __m128d c = _mm_andnot_pd(a, b);
+
+    // Take AND operation an complement of 'a' and 'b'. Bitwise operations are
```
marktwtn, line 481: `an`?

sse2neon.h
jserv: I agree. Let's copy the description on operations from Intel website.

jserv: Quote from Intel:
```
FOR j := 0 to 1
	i := j*64
	dst[i+63:i] := ((NOT a[i+63:i]) AND b[i+63:i])
ENDFOR
```

jserv (changes requested):
Use operation descriptions from Intel website

marktwtn (approved):

PR #129, opened 2020-08-31
refactor: Allow non-constant input _mm_slli_epi32

HowJMay:
_mm_slli_epi32 was not able to take non-constant value for imm.
Test case has been added in this PR, too.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1693,19 +1693,15 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
 // shifting in zeros. :
 // https://msdn.microsoft.com/en-us/library/z2k3bbtb%28v=vs.90%29.aspx
 // FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, __constrange(0,255) int imm)
-#define _mm_slli_epi32(a, imm)                                   \
-    __extension__({                                              \
-        __m128i ret;                                             \
-        if ((imm) <= 0) {                                        \
-            ret = a;                                             \
-        } else if ((imm) > 31) {                                 \
-            ret = _mm_setzero_si128();                           \
-        } else {                                                 \
-            ret = vreinterpretq_m128i_s32(                       \
-                vshlq_n_s32(vreinterpretq_s32_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, int imm)
+{
+    if (imm <= 0) /* TODO: add constant range macro: [0, 255] */
+        return a;
+    if (imm > 31) /* TODO: add unlikely macro */
+        return _mm_setzero_si128();
+    return vreinterpretq_m128i_s32(
+        vshlq_s32(vreinterpretq_s32_m128i(a), vdupq_n_s32(imm)));
+}
 
 // Shift packed 64-bit integers in a left by imm8 while shifting in zeros, and
 // store the results in dst.
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -3117,7 +3117,27 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int32_t *_b = (const int32_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    // Clang compiler does not allow the second argument of _mm_slli_epi32() to
+    // be greater than 31.
+    int count = (uint32_t) _b[0] % 32;
+#else
+    int count = (uint32_t) _b[0] % 64;
+    // The value for doing the modulo should be greater
+    // than 32. Using 64 would provide more equal
+    // distribution for both under 32 and above 32 input value.
+#endif
+
+    int32_t d0 = (count > 31) ? 0 : _a[0] << count;
+    int32_t d1 = (count > 31) ? 0 : _a[1] << count;
+    int32_t d2 = (count > 31) ? 0 : _a[2] << count;
+    int32_t d3 = (count > 31) ? 0 : _a[3] << count;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    __m128i c = _mm_slli_epi32(a, count);
+    return validateInt32(c, d0, d1, d2, d3);
 }
 
 result_t test_mm_srai_epi32(const SSE2NEONTestImpl &impl, uint32_t i)


tests/impl.cpp
```diff
@@ -3117,7 +3117,17 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int count = 3;
```
jserv, line 3121: Here `count` is `imm8`. You can do another test for the condition `imm8 > 31`.

marktwtn: Can this be set to random value?

marktwtn: Any progress?

HowJMay: I choose to use fix value

marktwtn: But the other situations would not be tested?

HowJMay: Or let me test all the cases

marktwtn: I think you can use the pass-in argument `i` and do the modulo calculation on it.
Like `int count = i % x`, x can be a value which is larger than 32.

tests/impl.cpp
```diff
@@ -3117,7 +3117,22 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int32_t *_b = (const int32_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    int count = (uint64_t) _b[0] % 31;
+#else
+    int count = (uint32_t) _b[0] % 63;  // Other value greater than 63
```
marktwtn, line 3125: Fix the comment.

tests/impl.cpp
```diff
@@ -3117,7 +3117,22 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int32_t *_b = (const int32_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    int count = (uint64_t) _b[0] % 32;
+#else
+    int count = (uint32_t) _b[0] % 32;  // Other value greater than 31
```
marktwtn, line 3125: The value should not be `32` since the result would be 0~31.
The comment should be `Other value greater than 32`.

HowJMay: it should be greater equal than 32 or greater than 31

marktwtn: Nope.
Greater than `31` means `32` is acceptable, which is not.

I think we have different definition of the comment.
For your definition, it should be `// For generating other value greater than 31`, which means the count variable.
For my definition, it should be `// The value for doing the modulo should be greater than 32`, which means the constant for doing the modulo.

Pick a comment and give a more detailed description.

tests/impl.cpp
```diff
@@ -3117,7 +3117,22 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int32_t *_b = (const int32_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    int count = (uint32_t) _b[0] % 32;
```
marktwtn, line 3123: Why don't you use `i` instead of `_b[0]`?
I am just curious.

HowJMay: I think `_b[0]` is more random value and `i` is a simple increasing `int`. Therefore, I choose the radnom one.

marktwtn (approved):

sse2neon.h
```diff
@@ -1692,20 +1692,16 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
...
-            ret = a;                                             \
-        } else if ((imm) > 31) {                                 \
-            ret = _mm_setzero_si128();                           \
-        } else {                                                 \
-            ret = vreinterpretq_m128i_s32(                       \
-                vshlq_n_s32(vreinterpretq_s32_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, int imm)
```
jserv, line 1695: Should be `const int imm`.

tests/impl.cpp
```diff
@@ -3117,7 +3117,25 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
...
+#else
+    int count = (uint32_t) _b[0] %
+                64;  // The value for doing the modulo should be greater
```
jserv, line 3126: Move `64` to the previous line.

sse2neon.h
marktwtn: Is it necessary to be `const int imm`?
Since based on the [_mm_slli_epi32](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_slli_epi32), it does not contain `const`?

sse2neon.h
```diff
@@ -1692,20 +1692,16 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
...
-            ret = a;                                             \
-        } else if ((imm) > 31) {                                 \
-            ret = _mm_setzero_si128();                           \
-        } else {                                                 \
-            ret = vreinterpretq_m128i_s32(                       \
-                vshlq_n_s32(vreinterpretq_s32_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, const int imm)
```
marktwtn, line 1695: Remove the `const`.
After that, I think the pull request can be merged.

tests/impl.cpp
```diff
@@ -3117,7 +3117,25 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int32_t *_b = (const int32_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    int count = (uint32_t) _b[0] % 32;
```
marktwtn, line 3123: Could you add the comment describing the reason for using `32` with the `Clang` compiler?
I remember that it is the compiler error which indicates that the value larger than `32` would out of the shift range.

marktwtn (changes requested):

sse2neon.h
marktwtn: @jserv I saw you previous comment that the `const` should be added.
However, based on the [_mm_slli_epi32](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_slli_epi32), it does not contain `const`.

jserv: Yes, there is no `const` qualifier according to Intel Intrinsics. However, it is indeed an `imm8`, which should be mapped to 8-bit width.

HowJMay: So should I conduct any change?

tests/impl.cpp
```diff
@@ -3117,7 +3117,27 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    const int32_t *_b = (const int32_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    // Clang compiler doesn't allow `_mm_slli_epi32()` take the second arg,
+    // count, greater than 31.
```
marktwtn, line 3124: Change to `Clang compiler does not allow the second argument of _mm_slli_epi32() to be greater than 31.`

marktwtn (changes requested):

marktwtn (approved):

sse2neon.h
```diff
@@ -1692,20 +1692,16 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
...
-            ret = vreinterpretq_m128i_s32(                       \
-                vshlq_n_s32(vreinterpretq_s32_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, int imm)
+{
+    if (imm <= 0) {
+        return a;
+    } else if (imm > 31) {
```
jserv, line 1699: Rewrite `} else if (imm > 31) {` as `if (imm > 31)`.

sse2neon.h
```diff
@@ -1692,20 +1692,16 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
 // Shifts the 4 signed or unsigned 32-bit integers in a left by count bits while
 // shifting in zeros. :
 // https://msdn.microsoft.com/en-us/library/z2k3bbtb%28v=vs.90%29.aspx
-// FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, __constrange(0,255) int imm)
```
jserv, line 1695: Let's move `__constrange(0,255)` back.  Someday, we can introduce something like [SIMDE_REQUIRE_CONSTANT_RANGE](https://github.com/simd-everywhere/simde/blob/master/simde/simde-common.h#L186) to validate imm8 at build time.

sse2neon.h
```diff
@@ -1693,19 +1693,17 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
...
-        } else if ((imm) > 31) {                                 \
-            ret = _mm_setzero_si128();                           \
-        } else {                                                 \
-            ret = vreinterpretq_m128i_s32(                       \
-                vshlq_n_s32(vreinterpretq_s32_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, int imm)
+{
```
jserv, line 1697: Rewrite as the following:
```c
    return a;
if (imm > 31) /* TODO: add unlikely macro */
    return _mm_setzero_si128();
return vreinterpretq_m128i_s32(
        vshlq_s32(vreinterpretq_s32_m128i(a), vdupq_n_s32(imm)));
```

PR #130, opened 2020-08-31
refactor: Allow non-constant input _mm_slli_epi64

HowJMay:
_mm_slli_epi64 was not able to take non-constant value for imm.
Test case has been added in this PR, too.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1781,19 +1781,15 @@ FORCE_INLINE __m128i _mm_slli_epi32(__m128i a, int imm)
 
 // Shift packed 64-bit integers in a left by imm8 while shifting in zeros, and
 // store the results in dst.
-#define _mm_slli_epi64(a, imm)                                   \
-    __extension__({                                              \
-        __m128i ret;                                             \
-        if ((imm) <= 0) {                                        \
-            ret = a;                                             \
-        } else if ((imm) > 63) {                                 \
-            ret = _mm_setzero_si128();                           \
-        } else {                                                 \
-            ret = vreinterpretq_m128i_s64(                       \
-                vshlq_n_s64(vreinterpretq_s64_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi64(__m128i a, int imm)
+{
+    if (imm <= 0) /* TODO: add constant range macro: [0, 255] */
+        return a;
+    if (imm > 63) /* TODO: add unlikely macro */
+        return _mm_setzero_si128();
+    return vreinterpretq_m128i_s64(
+        vshlq_s64(vreinterpretq_s64_m128i(a), vdupq_n_s64(imm)));
+}
 
 // Shifts the 8 signed or unsigned 16-bit integers in a right by count bits
 // while shifting in zeros.
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -4192,7 +4192,25 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int64_t *_b = (const int64_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    // Clang compiler does not allow the second argument of `_mm_slli_epi64()`
+    // to be greater than 63.
+    int count = (uint64_t) _b[0] % 64;
+#else
+    int count =
+        (uint64_t) _b[0] %
+        128;  // The value for doing the modulo should be greater
+              // than 64. Using 128 would provide more equal
+              // distribution for both under 64 and above 64 input value.
+#endif
+    int64_t d0 = (count > 63) ? 0 : _a[0] << count;
+    int64_t d1 = (count > 63) ? 0 : _a[1] << count;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    __m128i c = _mm_slli_epi64(a, count);
+    return validateInt64(c, d0, d1);
 }
 
 result_t test_mm_undefined_ps(const SSE2NEONTestImpl &impl, uint32_t i)


tests/impl.cpp
```diff
@@ -3429,7 +3429,15 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int count = 3;
```
jserv, line 3433: The last parameter of `_mm_slli_epi64` is `imm8`, and you took one numeric constant `3`.  You can add comments such as `const int count = 3; /* imm8 <= 63 */`.
Even better, you can do another test for `imm8 > 63`.

tests/impl.cpp
```diff
@@ -3429,7 +3429,30 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    result_t result1, result2;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    int count = 3; /* imm8 <= 63 */
+
+    int64_t d0 = _a[0] << count;
+    int64_t d1 = _a[1] << count;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    __m128i c = _mm_slli_epi64(a, count);
+    result1 = validateInt64(c, d0, d1);
```
jserv, line 3441: If `result1` was wrong, then we can return `TEST_FAIL`.

jserv (changes requested):
Fix CI regression

tests/impl.cpp
HowJMay: Actually I met a problem. What is the expected result when count is greater than 63?

jserv: Quote Intel Intrinsic Guide:
```
FOR j := 0 to 1
	i := j*64
	IF imm8[7:0] > 63
		dst[i+63:i] := 0
	ELSE
		dst[i+63:i] := ZeroExtend64(a[i+63:i] << imm8[7:0])
	FI
ENDFOR
```

You shall check the lowest byte (i.e. `count & 0xFF`)

jserv: You can check the value of `result1`. If it is false, then return `TEST_FAIL`. Thus, we don't need `if ((result1 == TEST_SUCCESS) && (result2 == TEST_SUCCESS))` statement anymore.

tests/impl.cpp
```diff
@@ -3502,7 +3502,29 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int count1 = 3; /* imm8 <= 63 */
```
marktwtn, line 3506: I think it can be simplified as
```c
#if define(__clang__)
int count = i % 64;
#else 
int count = i % 70; // or other value greater than 64
#endif 
```
and the extra code below can be removed.

tests/impl.cpp
```diff
@@ -3502,7 +3502,29 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
...
+
+    int64_t d0 = _a[0] << count1;
+    int64_t d1 = _a[1] << count1;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    __m128i c = _mm_slli_epi64(a, count1);
+#if !defined(__clang__)
```
jserv, line 3513: Why did you regard clang as a special case?

HowJMay: Clang prevents `_mm_slli_epi64` from taking an imm8 which is more than 63, so we have to exclude clang from the special case

tests/impl.cpp
```diff
@@ -3502,7 +3502,19 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int64_t *_b = (const int64_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
+    int count = (uint64_t) _b[0] % 64;
+#else
+    int count = (uint64_t) _b[0] % 128;  // Other value greater than 63
```
marktwtn, line 3510: Change the comment.

marktwtn (approved):

tests/impl.cpp
```diff
@@ -3502,7 +3502,23 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_slli_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int64_t *_b = (const int64_t *) impl.mTestIntPointer2;
+#if defined(__clang__)
```
marktwtn, line 3507: Please add the comment based on [this](https://github.com/DLTcollab/sse2neon/pull/129#discussion_r502309433).

marktwtn (changes requested):

tests/impl.cpp
marktwtn: After that, the pull request should be merged.

sse2neon.h
```diff
@@ -1709,19 +1709,16 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
...
-            ret = vreinterpretq_m128i_s64(                       \
-                vshlq_n_s64(vreinterpretq_s64_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi64(__m128i a, int imm)
+{
+    if (imm <= 0) {
+        return a;
+    } else if (imm > 63) {
```
jserv, line 1716: Change `} else if (imm > 63) {` to `if (imm > 63) {`.

HowJMay: I am curious about the reason for such a change.

jserv: Shorten code and make it cleaner.

HowJMay: got it

jserv: Later, we can introduce macros such as `likely` and `unlikely` as the hint to the compiler to emit instructions that will cause branch prediction to favour the "likely" side of a jump instruction. 
See https://stackoverflow.com/questions/109710/how-do-the-likely-unlikely-macros-in-the-linux-kernel-work-and-what-is-their-ben

HowJMay: Let me open another new issue for this requirement

tests/impl.cpp
```diff
@@ -3754,7 +3754,25 @@ result_t test_mm_srli_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    // Clang compiler doesn't allow `_mm_slli_epi64()` take the second arg,
+    // count, greater than 63.
```
marktwtn, line 3761: Rewrite the comment.

marktwtn (approved):

sse2neon.h
```diff
@@ -1711,19 +1711,17 @@ FORCE_INLINE __m128i _mm_srai_epi16(__m128i a, int imm)
...
-            ret = _mm_setzero_si128();                           \
-        } else {                                                 \
-            ret = vreinterpretq_m128i_s64(                       \
-                vshlq_n_s64(vreinterpretq_s64_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
-    })
+FORCE_INLINE __m128i _mm_slli_epi64(__m128i a, int imm)
+{
+    if (imm <= 0) {
```
jserv, line 1716: Rewrite as the following:
```c
    return a;
if (imm > 63) /* TODO: add unlikely macro */
    return _mm_setzero_si128()
return vreinterpretq_m128i_s64(
        vshlq_s64(vreinterpretq_s64_m128i(a), vdupq_n_s64(imm)));
```

marktwtn: @HowJMay any progress?

HowJMay: done

PR #134, opened 2020-09-01
test: Add test for _mm_insert_epi64

HowJMay:


tests/impl.cpp
```diff
@@ -3450,7 +3450,19 @@ result_t test_mm_extract_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_insert_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int imm8 = 1;
```
marktwtn, line 3454: I think this should be a random value which would be mod by `2`.

tests/impl.cpp
```diff
@@ -3450,7 +3450,19 @@ result_t test_mm_extract_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    const int64_t insert = 0xBBBBBBBBBBBBBBBB;
```
marktwtn, line 3455: You can use `impl.mTestIntPointer2` to assign it a value.

tests/impl.cpp
HowJMay: imm8 should take constant value at compiling time.
Using mod operation for a random number would cause error

tests/impl.cpp
```diff
@@ -3450,7 +3450,19 @@ result_t test_mm_extract_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_insert_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t *_a = (const int64_t *) impl.mTestIntPointer1;
+    const int64_t insert = (const int64_t ) *impl.mTestIntPointer2;
+    const int imm8 = insert & 1; // insert mod 2
+
+    int64_t d[2];
+    for (int i = 0; i < 2; i++) {
```
jserv, line 3458: Eliminate the loop since there are only two elements. Instead, use two assignments such as `d[0] = ...; d[1] = ...;`.

tests/impl.cpp
HowJMay: I will use fixed name temporarily

marktwtn: Let's keep the original design.

tests/impl.cpp
marktwtn: Replace the value.

HowJMay: It is macro. A variable is not allowed

marktwtn: I do not see any macro here.
Can't it be used like
```c
const int64_t *insert = (const int64_t *) impl.mTestIntPointer2;
```
and use `*insert` in the following code?

HowJMay: my fault

marktwtn (approved):

PR #151, opened 2020-09-11
Implement A64 fastpath for _mm_maddubs_epi16

jserv:


marktwtn:
Although the implementation is correct, I am not sure whether it would be faster or not in A64.

PR #153, opened 2020-09-12
refactor: Tweak and add test case for _mm_sign_epi8

HowJMay:
The original conversion of _mm_sign_epi8 was opposite to what document
mentioned. Here, the bug of the conversion has been fixed, and we add
A64 operation and corresponding test case.

Close #149

PR #155, opened 2020-09-14
fix: Fix wrong behavior of _mm_sign_epi32

HowJMay:
The behavior of _mm_sign_epi16 is opposite. The wrong behavior has been
fixed in this PR. A64 operation and test case have been appended in
this PR, too.


marktwtn (changes requested):
Rewrite the commit message like `The original conversion of _mm_sign_epi32 was opposite to what document mentioned`.

marktwtn (changes requested):
Rewrite the commit message like what you did in https://github.com/DLTcollab/sse2neon/pull/153.

marktwtn (approved):

PR #166, opened 2020-10-18
feat: Implement _mm_cvtpi32x2_ps

marktwtn:


sse2neon.h
```diff
@@ -3944,6 +3944,25 @@ FORCE_INLINE __m128 _mm_cvtpi32_ps(__m128 a, __m64 b)
...
+//   dst[31:0] := Convert_Int32_To_FP32(a[31:0])
+//   dst[63:32] := Convert_Int32_To_FP32(a[63:32])
+//   dst[95:64] := Convert_Int32_To_FP32(b[31:0])
+//   dst[127:96] := Convert_Int32_To_FP32(b[63:32])
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtpi32x2_ps
+FORCE_INLINE __m128 _mm_cvtpi32x2_ps(__m64 a, __m64 b)
+{
+    return vreinterpretq_m128_f32(
+        vcombine_f32(vcvt_f32_s32(vreinterpret_s32_m64(a)),
```
jserv, line 3962: How about `vcvtq_f32_s32(vcombine_s32(vreinterpret_s32_m64(a), vreinterpret_s32_m64(b)))`?
Which one is faster?

marktwtn: Although I have not done the test, I prefer the `vcvtq_f32_s32(vcombine_s32(vreinterpret_s32_m64(a), vreinterpret_s32_m64(b)))` implementation.

marktwtn: I did a simple time test and found out that most of the time difference of these two implementations would be less than `100 nsec`.

PR #171, opened 2020-10-19
feat: Implement _m_pmulhuw as macro and rewrite _mm_mulhi_pu16

marktwtn:


sse2neon.h
```diff
@@ -2734,6 +2734,23 @@ FORCE_INLINE __m128i _mm_mullo_epi32(__m128i a, __m128i b)
...
+// integers in dst.
+//
+//   FOR j := 0 to 3
+//      i := j*16
+//      tmp[31:0] := a[i+15:i] * b[i+15:i]
+//      dst[i+15:i] := tmp[31:16]
+//   ENDFOR
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_m_pmulhuw
+FORCE_INLINE __m64 _m_pmulhuw(__m64 a, __m64 b)
```
jserv, line 2748: Don't implement duplicated function.
Instead, we can simply use macro:
```c
#define _m_pmulhuw(a, b) _mm_mulhi_pu16(a, b)
```

tests/impl.cpp
```diff
@@ -4211,6 +4211,23 @@ result_t test_m_pavgw(const SSE2NEONTestImpl &impl, uint32_t i)
     return validateUInt16(c, d0, d1, d2, d3);
 }
 
+result_t test_m_pmulhuw(const SSE2NEONTestImpl &impl, uint32_t i)
```
jserv, line 4214: This test item is not required since we should focus on `_mm_mulhi_pu16`.

sse2neon.h
marktwtn: I did not notice that there is another duplicated function.
I think I will rewrite it like the duplicated function `_m_pavgb`
```c
FORCE_INLINE __m64 _m_pavgb(__m64 a, __m64 b)
{
    return _mm_avg_pu8(a, b);
}
```
I am not sure which one is better.
Using macro would be simpler but using function would be more consistent.

Besides, I will change the implementation of `_mm_mulhi_pu16` from
`vmovn_u32 + vshrq_n_u32` to `vshrn_n_u32`.

tests/impl.cpp
marktwtn: I think I will keep it to make sure that the developer would know the function conversion has been implemented.

sse2neon.h
jserv: We can implement inline functions starting with `_mm_`, and `_m_pavgb` can be reimplemented as a macro.

marktwtn: Got it.
I will do it in the future pull request.

tests/impl.cpp
jserv: We only need the test case for `_mm_mulhi_pu16`.

PR #174, opened 2020-10-20
feat: Implement _m_psadbw as macro

marktwtn:


tests/impl.cpp
```diff
@@ -1136,6 +1136,21 @@ result_t test_mm_dp_ps(const SSE2NEONTestImpl &impl, uint32_t i)
     return validateFloatEpsilon(out, r[0], r[1], r[2], r[3], 2050.0f);
 }
 
+result_t test_m_psadbw(const SSE2NEONTestImpl &impl, uint32_t i)
```
jserv, line 1139: It is not worthy to validate the straightforward macro.

marktwtn: It has been removed.

PR #177, opened 2020-10-21
refactor: Reorder the intrinsic functions with class and alphabet

marktwtn:
The order is following the alphabetical order of each intrinsic class
Ref: https://software.intel.com/sites/landingpage/IntrinsicsGuide/


marktwtn:
The pull request should be modified after the #130 is merged.

marktwtn:
@jserv although the pull request includes a lot of modification, it just reorders the instruction.
I think it can be merged without any problem.

jserv:
> @jserv although the pull request includes a lot of modification, it just reorders the instruction.
> I think it can be merged without any problem.

I agree. Later, we can prepare some scripts to automate the reordering.

PR #186, opened 2020-10-22
refactor: Move the _mm_mul_epi32 test to the correct classification

marktwtn:


jserv:
The subject was misleading. The real item you moved is the test case rather than `_mm_mul_epi32` implementation.

jserv (changes requested):
Instead of "right position," I think the patch changed to correct classification.

PR #218, opened 2020-10-29
fix: Fix shift-related conversion and implement the tests

### LINKED ISSUE #20: Implement the rest of the shift intrinsic functions
The following shift intrinsic functions are not implemented yet.
- _mm_sll_epi*
- _mm_srl_epi*
- _mm_sra_epi*
- _mm_sllv_epi*
- _mm_srlv_epi*
- _mm_srav_epi*

However, the related shift intrinsic functions of **x86** and **ARM** have different behaviour when the shift value is too large .
Please check the discussion in https://github.com/DLTcollab/sse2neon/pull/17#discussion_r374810457.

marktwtn:
Close #20.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1834,82 +1834,115 @@ FORCE_INLINE __m128i _mm_slli_epi64(__m128i a, int imm)
         vshlq_s64(vreinterpretq_s64_m128i(a), vdupq_n_s64(imm)));
 }
 
-// Shifts the 8 signed or unsigned 16-bit integers in a right by count bits
-// while shifting in zeros.
+// Shift packed 16-bit integers in a right by imm8 while shifting in zeros, and
+// store the results in dst.
 //
-//   r0 := srl(a0, count)
-//   r1 := srl(a1, count)
-//   ...
-//   r7 := srl(a7, count)
+//   FOR j := 0 to 7
+//     i := j*16
+//     IF imm8[7:0] > 15
+//       dst[i+15:i] := 0
+//     ELSE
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -1834,82 +1834,115 @@ FORCE_INLINE __m128i _mm_slli_epi64(__m128i a, int imm)
...
-        } else {                                                 \
-            ret = vreinterpretq_m128i_u16(                       \
-                vshrq_n_u16(vreinterpretq_u16_m128i(a), (imm))); \
-        }                                                        \
-        ret;                                                     \
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_srli_epi16
+#define _mm_srli_epi16(a, imm)                                             \
+    __extension__({                                                        \
+        __m128i ret;                                                       \
+        if ((imm) == 0) {                                                  \
```
jserv, line 1853: How about the following changes?
```c
if ((imm) <= 0) {
    ret = a;
} else if ((imm) > 15) {
    ret = _mm_setzero_si128();
} else {
    ret = vreinterpretq_m128i_u16(vshrq_n_u16(vshlq_u16(vreinterpretq_u16_m128i(a),
         (((imm) & 15) | (((imm) & 15) == 0))));
}
```

marktwtn: As I executed the `_mm_srli_epi16` on X86, it turns out that when the `imm` is less than 0 or greater than 15, the vector element would become 0.

The reason of not using `vshrq_n_u16` is that the second argument of it should be the constant immediate value.
However, the `_mm_srli_epi16` does not.

PR #220, opened 2020-11-10
test: Add test for _mm_store_ss

HowJMay:


tests/impl.cpp
```diff
@@ -1991,7 +1991,13 @@ result_t test_mm_store_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_store_ss(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    float *p = (float *) impl.mTestFloatPointer1;
+    float x = impl.mTestFloats[i + 4];
```
marktwtn, line 1995:
```
float x = impl.mTestFloats[i];
float p[4];
```

I think we should not store the value into input data.

marktwtn (changes requested):

tests/impl.cpp
HowJMay: I refered to `test_mm_store_ps()`
`impl.mTestFloats` is taking a random float as testing data

marktwtn: You mean referred from `test_mm_store_ps()`?
We still have random float in this modification.

HowJMay: Is there any suggested way to get a random testing number

marktwtn: We already have it.
The `impl.mTestFloats` array is generated with random float values.

HowJMay: so I can just simply use `impl.mTestFloats[0]`?

marktwtn: Use `impl.mTestFloats[i]`.

HowJMay: got it

tests/impl.cpp
```diff
@@ -1991,7 +1991,13 @@ result_t test_mm_store_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_store_ss(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    float *p = (float *) impl.mTestFloatPointer1;
```
marktwtn, line 1994: Rewrite it as `float p[4];`.

marktwtn (changes requested):

marktwtn (approved):

PR #221, opened 2020-11-14
feat: Implement _mm_mulhi_epu16

HowJMay:


sse2neon.h
```diff
@@ -3602,6 +3602,25 @@ FORCE_INLINE __m128i _mm_mulhi_epi16(__m128i a, __m128i b)
...
+// intermediate 32-bit integers, and store the high 16 bits of the intermediate
+// integers in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_mulhi_epu16
+FORCE_INLINE __m128i _mm_mulhi_epu16(__m128i a, __m128i b)
+{
+    /* FIXME: issue with large values because of result saturation as
+     * _mm_mulhi_epi16 */
+    uint16x4_t a3210 = vget_low_u16(vreinterpretq_u16_m128i(a));
+    uint16x4_t b3210 = vget_low_u16(vreinterpretq_u16_m128i(b));
+    uint32x4_t ab3210 = vmull_u16(a3210, b3210); /* 3333222211110000 */
```
jserv, line 3615: Provide Aarch64 fastpath right after this line.
```c
#if defined(__aarch64__)
    uint32x4_t ab7654 = vmull_high_u16(vreinterpretq_u16_m128i (a), vreinterpretq_u16_m128i (b));
    uint16x8x2_t r = vuzp2q_u16(vreinterpretq_u16_u32(ab3210), vreinterpretq_u16_u32(ab7654));
    return vreinterpretq_m128i_u16(r);
#endif
```

marktwtn, line 3615: The comment is not really meaningful.
Remove it or rewrite it.

sse2neon.h
```diff
@@ -3602,6 +3602,25 @@ FORCE_INLINE __m128i _mm_mulhi_epi16(__m128i a, __m128i b)
...
+    uint16x4_t a7654 = vget_high_u16(vreinterpretq_u16_m128i(a));
+    uint16x4_t b7654 = vget_high_u16(vreinterpretq_u16_m128i(b));
+    uint32x4_t ab7654 = vmull_u16(a7654, b7654); /* 7777666655554444 */
```
marktwtn, line 3618: Ditto.

sse2neon.h
```diff
@@ -3602,6 +3602,33 @@ FORCE_INLINE __m128i _mm_mulhi_epi16(__m128i a, __m128i b)
     return vreinterpretq_m128i_u16(r.val[1]);
 }
 
+// Multiply the packed unsigned 16-bit integers in a and b, producing
+// intermediate 32-bit integers, and store the high 16 bits of the intermediate
+// integers in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_mulhi_epu16
+FORCE_INLINE __m128i _mm_mulhi_epu16(__m128i a, __m128i b)
+{
+    /* FIXME: issue with large values because of result saturation as
```
jserv, line 3611: Can you provide more precise comments which address the issue? That is, the range and/or condition to cause saturation,

HowJMay: After checking out the document of both [Neon](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vmull_u16) and [SSE](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_mulhi_epu16), I think the saturation problem doesn't exist as 
[_mm_mulhi_epi16](https://github.com/DLTcollab/sse2neon/blob/master/sse2neon.h#L3724) does. Therefore, I have removed the comment

marktwtn (approved):

PR #222, opened 2020-11-17
test: Add test of _mm_storeu_ps

HowJMay:


tests/impl.cpp
```diff
@@ -2026,7 +2026,16 @@ result_t test_mm_storel_pi(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_storeu_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    float *f = impl.mTestFloatPointer1;
+    float _a[4] = {1.0, 2.0, 3.0, 4.0};
```
jserv, line 2030: Why did you pick up the sequence?

HowJMay: let me use `impl.mTestFloats` instead

marktwtn: You just need to rewrite it as `float _a[4];`

tests/impl.cpp
```diff
@@ -2026,7 +2026,16 @@ result_t test_mm_storel_pi(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    __m128 a = _mm_load_ps(_a);
+
+    _mm_storeu_ps(f, a);
```
marktwtn, line 2033: You are storing the value of `a` to the memory address `f`.
`f` can be declared as an uninitialized array.
`a` should contain random value.

Rewrite as
```c
    float *_a = impl.mTestFloatPointer1;
    float f[4];
```

marktwtn (changes requested):

tests/impl.cpp
```diff
@@ -2026,7 +2026,16 @@ result_t test_mm_storel_pi(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    float *_a = impl.mTestFloatPointer1;
+    float f[4];
+    __m128 a = _mm_load_ps(_a);
+
+    _mm_storeu_ps(f, a);
+    ASSERT_RETURN(f[0] == _a[0]);
+    ASSERT_RETURN(f[1] == _a[1]);
+    ASSERT_RETURN(f[2] == _a[2]);
+    ASSERT_RETURN(f[3] == _a[3]);
+    return TEST_SUCCESS;
```
marktwtn, line 2038: How about rewrite it as `validateFloat(a, f[0], f[1], f[2], f[3]);`?

tests/impl.cpp
```diff
@@ -2026,7 +2026,16 @@ result_t test_mm_storel_pi(const SSE2NEONTestImpl &impl, uint32_t i)
...
-    return TEST_UNIMPL;
+    float *_a = impl.mTestFloatPointer1;
+    float f[4];
+    __m128 a = _mm_load_ps(_a);
+
+    _mm_storeu_ps(f, a);
+    ASSERT_RETURN(f[0] == _a[0]);
+    ASSERT_RETURN(f[1] == _a[1]);
+    ASSERT_RETURN(f[2] == _a[2]);
+    ASSERT_RETURN(f[3] == _a[3]);
```
marktwtn, line 2037: These shall be removed.

marktwtn (changes requested):

marktwtn (approved):

PR #223, opened 2020-11-19
feat: Implement _mm_cvt_ps2pi

marktwtn:


sse2neon.h
```diff
@@ -4296,6 +4296,38 @@ FORCE_INLINE __m128 _mm_cvt_pi2ps(__m128 a, __m64 b)
...
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvt_ps2pi
+FORCE_INLINE __m64 _mm_cvt_ps2pi(__m128 a)
+{
+#if defined(__aarch64__)
+    return vreinterpret_m64_s32(
+        vget_low_s32(vcvtnq_s32_f32(vreinterpretq_f32_m128(a))));
+#else
+    int32_t res[2];
+    float32_t *data = (float32_t *) &a;
+    for (int i = 0; i < 2; i++) {
```
HowJMay, line 4316: Is it better to unfold it?

marktwtn: For performance, it is.
I'll unfold it.

sse2neon.h
```diff
@@ -4296,6 +4296,45 @@ FORCE_INLINE __m128 _mm_cvt_pi2ps(__m128 a, __m64 b)
...
+//   ENDFOR
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvt_ps2pi
+FORCE_INLINE __m64 _mm_cvt_ps2pi(__m128 a)
+{
+#if defined(__aarch64__)
+    return vreinterpret_m64_s32(
+        vget_low_s32(vcvtnq_s32_f32(vreinterpretq_f32_m128(a))));
+#else
+    int32_t res[2];
```
jserv, line 4314: How about moving the following A32 specific implementation to `_mm_round_ps`? Thus, we can reimplement as following:
```c
a = _mm_round_ps(a, _MM_FROUND_CUR_DIRECTION);
return vcvt_s32_f32(vget_low_f32(a));
```

PR #224, opened 2020-11-19
test: Add test for _mm_extract_epi32

HowJMay:


tests/impl.cpp
```diff
@@ -5238,7 +5238,14 @@ result_t test_mm_dp_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
 result_t test_mm_extract_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    int32_t *_a = (int32_t *) impl.mTestIntPointer1;
+    const int imm = 1;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    int c = _mm_extract_epi32(a, imm);
+
+    ASSERT_RETURN(c == *(_a + imm));
```
marktwtn, line 5247: I think `*(_a + imm)` is only suitable when `imm` is between 0 ~ 3?

HowJMay: I tried an `imm` which is greater than 3. It is an illegal case for GCC

marktwtn (approved):

PR #228, opened 2020-11-22
feat: Add macro unlikely for branch prediction

HowJMay:
Add GCC specific macro unlikely to enhance branch prediction.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -61,6 +61,8 @@
 #pragma push_macro("ALIGN_STRUCT")
 #define FORCE_INLINE static inline __attribute__((always_inline))
 #define ALIGN_STRUCT(x) __attribute__((aligned(x)))
+#define likely(x) __builtin_expect(!!(x), 1)
+#define unlikely(x) __builtin_expect(!!(x), 0)
 #else
 #error "Macro name collisions may happen with unsupported compiler."
 #ifdef FORCE_INLINE
@@ -70,6 +72,10 @@
 #ifndef ALIGN_STRUCT
 #define ALIGN_STRUCT(x) __declspec(align(x))
 #endif
+#endif
+#ifndef likely
+#define likely(x) (x)
+#define unlikely(x) (x)
 #endif
 
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


jserv:
Can you ensure the effectiveness reported by [godbolt](https://godbolt.org/)? That is, verify the generated code with `__builtin_expect`.

marktwtn:
@HowJMay any progress?

HowJMay:
@marktwtn it is ok for merging

marktwtn:
> 
> 
> Can you ensure the effectiveness reported by [godbolt](https://godbolt.org/)? That is, verify the generated code with `__builtin_expect`.

It seems like the question of @jserv is not answered?

HowJMay:
the following is the comparison between the 2 situations
## with unlikely
```c
int _mm_slli_epi32(int a, int imm)
{
        return a;
    if (unlikely(imm > 31))
        return 0;
    return 1;
}
```

assembly
```assembly
_mm_slli_epi32(int, int):
        testl   %esi, %esi
        jle     .L3
        cmpl    $31, %esi
        setg    %al
        movzbl  %al, %eax
        testq   %rax, %rax
        je      .L4
        movl    $0, %eax
        ret
.L3:
        movl    %edi, %eax
        ret
.L4:
        movl    $1, %eax
        ret
```

HowJMay:
It does show different, so this PR can be merged.

marktwtn (approved):

jserv (changes requested):
Figure out more occurrences that fit branch prediction hints.

sse2neon.h
```diff
@@ -70,6 +72,8 @@
 #ifndef ALIGN_STRUCT
 #define ALIGN_STRUCT(x) __declspec(align(x))
 #endif
+#define likely(x) (x)
```
jserv, line 75: Add guard like this:
```c
#ifndef likely
#define likely(x) (x)
#endif
```
For compiler lacking of `push_macro` feature, it is necessary to avoid the pollution of macros.

PR #234, opened 2020-12-04
feat: Implement _mm_cvtss_si[32,64]

marktwtn:


sse2neon.h
```diff
@@ -419,6 +419,39 @@ FORCE_INLINE float _mm_cvtss_f32(__m128 a)
...
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtss_si64
+FORCE_INLINE int _mm_cvtss_si64(__m128 a)
+{
+#if defined(__aarch64__)
+    return vgetq_lane_s64(
+        vreinterpretq_s64_s32(vcvtnq_s32_f32(vreinterpretq_f32_m128(a))), 0);
+#else
+    float32_t data = vgetq_lane_f32(vreinterpretq_f32_m128(a), 0);
+    float32_t diff = data - floor(data);
```
jserv, line 443: It is feasible to replace the following code snip with round implementations:
* `__builtin_roundf`
* `roundf`

Then, use `ROUND_IMPL(vgetq_lane_f32(vreinterpretq_f32_m128(a), 0))`.

marktwtn: We can not replace it since the `roundf` and `__buildin_roundf` use the rounding mode `round to nearest, where ties round away from zero`.
The x86 intrinsics use the rounding mode `round to nearest, where ties round to the nearest even digit in the required position`.

PR #235, opened 2020-12-12
Update readme with Minimap2 as an example software that adopts sse2neon

hasindu2008:
Add Minimap2 as software that adopts sse2neon


jserv:
Thank @hasindu2008 for contributing!

PR #236, opened 2020-12-12
Added _mm_storeu_si<16-64>

RemiArnaud:
How does this look?

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -737,29 +737,35 @@ FORCE_INLINE void _mm_store_si128(__m128i *p, __m128i a)
     vst1q_s32((int32_t *) p, vreinterpretq_s32_m128i(a));
 }
 
-// Stores four 32-bit integer values as (as a __m128i value) at the address p.
-// https://msdn.microsoft.com/en-us/library/vstudio/edk11s13(v=vs.100).aspx
+// Stores 128-bits of integer data a at the address p.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si128
 FORCE_INLINE void _mm_storeu_si128(__m128i *p, __m128i a)
 {
     vst1q_s32((int32_t *) p, vreinterpretq_s32_m128i(a));
 }
 
+// Stores 64-bits of integer data a at the address p.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si64
 FORCE_INLINE void _mm_storeu_si64(void *p, __m128i a)
 {
-    int64x1_t lo =  vget_low_s64(vreinterpretq_s64_m128i(a));
-	* (int64x1_t *) p = lo;
+    int64x1_t lo = vget_low_s64(vreinterpretq_s64_m128i(a));
+    *(int64x1_t *) p = lo;
 }
 
+// Stores 32-bits of integer data a at the address p.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si32
 FORCE_INLINE void _mm_storeu_si32(void *p, __m128i a)
 {
-	int32x2_t lo = vget_low_s32(vreinterpretq_s32_m128i(a));
-	* (uint32_t*) p = lo[0];
+    int32x2_t lo = vget_low_s32(vreinterpretq_s32_m128i(a));
+    *(uint32_t *) p = lo[0];
 }
 
+// Stores 16-bits of integer data a at the address p.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si16
 FORCE_INLINE void _mm_storeu_si16(void *p, __m128i a)
 {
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
Please execute "make indent" to ensure the consistent style.

sse2neon.h
```diff
@@ -743,6 +744,24 @@ FORCE_INLINE void _mm_storeu_si128(__m128i *p, __m128i a)
     vst1q_s32((int32_t *) p, vreinterpretq_s32_m128i(a));
 }
 
+FORCE_INLINE void _mm_storeu_si64(void *p, __m128i a)
+{
+    int64x1_t lo =  vget_low_s64(vreinterpretq_s64_m128i(a));
+	* (int64x1_t *) p = lo;
```
marktwtn, line 750: I think it can be replaced with
```
vst1q_lane_s64((int64_t *) p, vreinterpretq_s64_m128i(a), 0);
```

tests/impl.cpp
```diff
@@ -4107,6 +4107,36 @@ result_t test_mm_storeu_si128(const SSE2NEONTestImpl &impl, uint32_t i)
     return validateInt32(a, _b[0], _b[1], _b[2], _b[3]);
 }
 
+result_t test_mm_storeu_si64(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    __m128i b;
+    __m128i a = _mm_loadu_si128((const __m128i *) _a);
```
marktwtn, line 4114: Replace with
```
__m128i a = do_mm_load_ps(_a);
```

marktwtn (changes requested):
First, edit and add the corresponding macros in the `tests/impl.h` file.
Please reference: https://github.com/DLTcollab/sse2neon#add-more-test-items

Second, please change the test function position in `tests/impl.cpp`.
The order should be like `tests/impl.h`, which is classified with different `SSE` type and follow the alphabetical order.

sse2neon.h
RemiArnaud: This would write a 0 into the second 64bits, whereas the intent is to only write into the first 64 bits. So I don't think the suggested replacement would work

tests/impl.cpp
RemiArnaud: done

RemiArnaud:
@jserv @marktwtn - any idea why the build fails for one of the configuration? how can I repro?

sse2neon.h
marktwtn: As I replaced it with `vst1q_lane_s64`, it does not behave like you described.
Have you tried it before?

tests/impl.cpp
```diff
@@ -257,6 +257,28 @@ result_t do_mm_store_ps(int32_t *p, int32_t x, int32_t y, int32_t z, int32_t w)
...
 
+result_t test_mm_storeu_si16(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    const int32_t *_a = (const int32_t *) impl.mTestIntPointer1;
+    __m128i b;
+    __m128i a = do_mm_load_ps(_a);
+    _mm_storeu_si16(&b, a);
+    int16_t *_b = (int16_t *) &b;
+    int16_t *_c = (int16_t *) &a;
+    return validateInt16(a, _b[0], _c[1], _c[2], _c[3], _c[4], _c[5], _c[6],
```
marktwtn, line 268: Although the result is correct, the logic is kind of weird.
It should be `validateInt16(b, _a[0], b[1], ......)`.
We care about the data stored into the memory instead of the input data for testing.

marktwtn (changes requested):

marktwtn:
> @jserv @marktwtn - any idea why the build fails for one of the configuration? how can I repro?

Execute the command `$ make check` and the failure would be reproduced.
I believe it is caused by the GCC that is does not implement ` _mm_storeu_si16` and ` _mm_storeu_si32`.

jserv:
> Execute the command `$ make check` and the failure would be reproduced.
> I believe it is caused by the GCC which does not implement ` _mm_storeu_si16` and ` _mm_storeu_si32`.

Recent GCC lacks of the following intrinsics:
* `_mm_storeu_si16` (SSE)
* `_mm_loadu_si16` (SSE)
* `_mm_loadu_si32` (SSE2)
* `_mm_storeu_si32` (SSE2)

The commit on Oct 14 this year finally implemented the above. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95483

marktwtn:
@RemiArnaud any progress?

jserv (changes requested):
Rebase the latest master branch

jserv:
@RemiArnaud, can you rework this patch according to the above discussions?

marktwtn:
Should I take care of this pull request?

jserv:
> Should I take care of this pull request?

Please do.

jserv:
supersede by commit 3294d69d9c706eb20eafb00bc628778dcbd9af19

PR #240, opened 2020-12-22
refactor: Wrap _mm_load_pd for unification

marktwtn:


jserv (changes requested):
Improve the commit message. You should explain why you would make such change.

marktwtn:
> Improve the commit message. You should explain why you would make such change.

The reason to do such change is for unification and wrapping.
For writing the tests, loading `__mm128`, `__mm128i`, `__mm128d` or `__m64` data from address is necessary.
We already have `do_mm_load_ps` for `__mm128` and `__mm128i`.
However, we do not have the wrapped functions for `__mm128d` and `__m64`.
Therefore, the commit is for the wrapped function for `__m128d`.

The wrapped function naming might be changed and the wrapped function for `__m64` might be added in the future.

PR #246, opened 2020-12-28
feat: Implement _mm_extract_pi16

HowJMay:


HowJMay:
https://travis-ci.com/github/DLTcollab/sse2neon/jobs/465364435 the CI failed, since the gcc is in version 7.
Should we update it to version 10?

marktwtn:
> https://travis-ci.com/github/DLTcollab/sse2neon/jobs/465364435 the CI failed, since the gcc is in version 7.
> Should we update it to version 10?

Is it caused by the GCC version? Is there any reference?
From the log, I think the tests are compiled with `g++-10`.

HowJMay:
> > https://travis-ci.com/github/DLTcollab/sse2neon/jobs/465364435 the CI failed, since the gcc is in version 7.
> > Should we update it to version 10?
> 
> Is it caused by the GCC version? Is there any reference?
> From the log, I think the tests are compiled with `g++-10`.

I saw the wrong one. I will dig the reason that cause this problem

tests/impl.cpp
```diff
@@ -3095,6 +3095,17 @@ result_t test_mm_extract_epi16(const SSE2NEONTestImpl &impl, uint32_t i)
...
 }
 
+result_t test_mm_extract_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    uint16_t *_a = (uint16_t *) impl.mTestIntPointer1;
+    const int imm = 1;
+
+    const __m64 *a = ((const __m64 *) _a);
+    int c = _mm_extract_pi16(*a, imm);
+    ASSERT_RETURN((uint16_t) c == *(_a + imm));
```
marktwtn, line 3105: The casting should be `uint16_t` to `int`.
The return value of `_mm_extract_pi16` is `int`, and we should focus on it.

Be aware that the test of `mm_extract_epi16` might have the same problem.

marktwtn (changes requested):

tests/impl.cpp
HowJMay: But `_mm_extract_pi16` takes arithmetic shift operation, which means expected value 65535 will return -1 by `_mm_extract_pi16`

marktwtn: Then I think the comparison should include two parts, which are `c[31:16]` and `c[15:0]`.

HowJMay: sorry I don't understand what you expect

marktwtn:
```
dst[15:0] := (a[63:0] >> (imm8[1:0] * 16))[15:0]
dst[31:16] := 0
```
The test should verify this.

HowJMay: I found the behavior is not actually the same as the pseudocode you showed
with `make check`  and the following test I got such response

```
int16_t *_a = (int16_t *) impl.mTestIntPointer1;
    const int imm = 1;

    const __m64 *a = ((const __m64 *) _a);
    int c = _mm_extract_pi16(*a, imm);

    printf("_a= %d, %d, %d, %d\n", _a[0], _a[1], _a[2], _a[3]);
    printf("c= %d, (c & 0xFFFF0000) = %d\n",c,  (c & 0xFFFF0000));
```

this will generate this response 
```
_a= 13943, -1, -28555, 0
c= -1, (c & 0xFFFF0000) = -65536
```

this response shows intrinsics, `_mm_extract_pi16` doesn't set the leading 16 bits as zeros, instead it just takes one of the  16 bits integer from a 64 bits long integer

HowJMay: so keeping the current test may be a better idea I think

marktwtn: The reason of the mismatching behaviour should be found.
I wonder whether this is caused by the compiler or not.

I will try to investigate.

marktwtn: The document of [PEXTRW - Extract Word](https://www.felixcloutier.com/x86/pextrw) says that the upper bits should be set to 0.
And the reported [bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98495) of GCC Bugzilla says that `X86 _mm_extract_pi16 incorrectly sign extends result`.

Please modify the test as I suggested and implement the test for `Clang` only.
The `GCC` part should be left unimplemented and add the corresponding comment that indicates it is caused by a bug and might be fixed in `GCC-11`.

sse2neon.h
```diff
@@ -6368,6 +6368,11 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
 #define _mm_extract_epi16(a, imm) \
     vgetq_lane_u16(vreinterpretq_u16_m128i(a), (imm))
 
+// Extract a 16-bit integer from a, selected with imm8, and store the result in
+// the lower element of dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_extract_pi16
+#define _mm_extract_pi16(a, imm) vget_lane_u16(vreinterpret_u16_m64(a), (imm))
```
jserv, line 6374: Use inline function instead. Thus, checks on parameters can be applied.

tests/impl.cpp
```diff
@@ -1636,7 +1636,18 @@ result_t test_mm_div_ss(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_extract_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+#if defined(__clang__)
```
jserv, line 1639: You have to explain why the test case is dedicated to clang.

tests/impl.cpp
```diff
@@ -1636,7 +1636,20 @@ result_t test_mm_div_ss(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    uint64_t *_a = (uint64_t *) impl.mTestIntPointer1;
+    const int imm = 1;
+
+    const __m64 *a = ((const __m64 *) _a);
+    int c = _mm_extract_pi16(*a, imm);
+    ASSERT_RETURN((uint64_t) c == ((*_a >> ((imm & 0x3) * 16)) & 0xFFFF));
+    ASSERT_RETURN(0 == ((uint64_t) c & 0xFFFF0000));
+    return TEST_SUCCESS;
+#else
+    // FIXME GCC has bug on `_mm_extract_pi16` intrinsics. We will enable this
```
jserv, line 1649: Move this comment to the beginning of function body.

sse2neon.h
HowJMay: I met an error 
```
./sse2neon.h:6376:12: error: argument to '__builtin_neon_vget_lane_i16' must be a constant integer
    return vget_lane_u16(vreinterpret_u16_m64(a), imm);
```
So I change it back to macro

jserv: Use multi-line form of macro. See `_mm_srai_epi32` as an example.
The single-line form of macro is used for equivalent expression.

tests/impl.cpp
```diff
@@ -1636,7 +1636,20 @@ result_t test_mm_div_ss(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_extract_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+    // FIXME GCC has bug on `_mm_extract_pi16` intrinsics. We will enable this
```
jserv, line 1639: While you were mentioning GCC bug, can you give some reference links? such as GCC bugzilla and/or explicit issues.

sse2neon.h
```diff
@@ -6389,6 +6389,11 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
```
marktwtn, line 6395: Shall we do the type casting?
Since it returns the type `uint16_t` but the returned type of `_mm_extract_pi16(a, imm)` is `int32_t`.

marktwtn:
@HowJMay after this pull request is merged, please help change the testing of `test_mm_extract_epi16` as well.

sse2neon.h
HowJMay: Should I use `int` or `int32_t`? It is `int` on official website
```
int _mm_extract_pi16 (__m64 a, int imm8)
```

marktwtn: As the description of the document, the result would be 32-bit.
I think we can use `int32_t` without any problem.

marktwtn (approved):

PR #254, opened 2020-12-30
CI: Add the yaml file for Github Actions

### LINKED ISSUE #253: CI: Migrate from Travis CI to Github Actions
Travis CI introduced the new pricing structure that each running test would consume the credits until it is empty.
Since we have run out of the credits, we decide to move to another CI system, which is Github Actions.

The CI should include running the test of different platforms and compilers
Platform:
- x86
- arm

Compiler:
- GCC
- Clang

And it should check the coding convention as well.

marktwtn:
check-format.sh:
The clang-format should use version 6 instead of 10.
Discussion: #71 (comment)
Makefile:
The returned processor type of Github Actions arm 32-bit platform is armv7l,
hence it should be included in Makefile detection.
github_actions.yml:
The non-x86 CPU architecture is offered by https://github.com/uraimo/run-on-arch-action.
The related YAML should be written by following the document rule.

Close #253.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -1,0 +1,56 @@
+name: Github Actions
+
+on: [push, pull_request]
+
+jobs:
+  test_x86:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        cxx_compiler: [g++-10, clang++-10]
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+      - name: test
+        env:
+          CXX: ${{ matrix.cxx_compiler }}
+        run: |
+          sh .ci/cross-tool.sh
+          make check
+          sh .ci/cross-check.sh
+
+  test_arm:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        arch: [armv7, aarch64]
+        cxx_compiler: [g++-10, clang++-10]
+    steps:
+      - name: checkout code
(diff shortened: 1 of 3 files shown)


marktwtn:
Do not merge this pull request until the Github Actions is passed.

marktwtn:
@jserv the pull request is ready for being reviewed.

.github/workflows/github_actions.yml
```diff
@@ -0,0 +1,57 @@
+name: Github Actions
+
+on: [push, pull_request]
+
+jobs:
+  test_x86:
+    runs-on: ubuntu-latest
```
jserv, line 7: Shall we specify the version for Ubuntu Linux? That is, 20.04-LTS in particular.

marktwtn: The `ubuntu-latest` would be `18.04` now.
Please check:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

Besides, the Github Actions page has warned us that:
```
Ubuntu-latest workflows will use Ubuntu-20.04 soon.
For more details, see https://github.com/actions/virtual-environments/issues/1816 
```
I think we should replace it with `20.04`.
Based on the https://github.com/actions/virtual-environments/issues/1816, it would use `Clang` version 10 as default, which is the compiler version we use.
Soon the `ubuntu-latest` would become `20.04`.

PR #256, opened 2021-01-02
fix: Complete mm_blendv_ps with signed shift right 

HowJMay:
Mapping `_mm_blendv_ps` intrinsics directly with `vbslq_f32` is not an
accurate implementaion. We should take signed shift right then call
`vbslq_f32`.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -5321,7 +5321,29 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_blendv_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const float *_a = impl.mTestFloatPointer1;
+    const float *_b = impl.mTestFloatPointer2;
+    const float _mask[] = {impl.mTestFloats[i], impl.mTestFloats[i+1],
+                           impl.mTestFloats[i+2], impl.mTestFloats[i+3]};
+
+    float _c[4];
+    for (int i = 0; i < 4; i++) {
+        // signed shift right would return a result which is either all 1's from
+        // negative numbers or all 0's from positive numbers
+        if ((*(const int32_t *) (_mask + i)) >> 31) {
+            _c[i] = _b[i];
+        } else {
+            _c[i] = _a[i];
+        }
+    }
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 b = do_mm_load_ps(_b);
+    __m128 mask = do_mm_load_ps(_mask);
+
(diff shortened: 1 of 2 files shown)


tests/impl.cpp
```diff
@@ -5265,7 +5265,29 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_blendv_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const float *_a = impl.mTestFloatPointer1;
+    const float *_b = impl.mTestFloatPointer2;
+    const float _mask[] = {impl.mTestFloats[0], impl.mTestFloats[1],
+                           impl.mTestFloats[2], impl.mTestFloats[3]};
```
marktwtn, line 5271: Replace with `i`, `i+1`, `i+2`, `i+3`.

tests/impl.cpp
```diff
@@ -5265,7 +5265,29 @@ result_t test_mm_blendv_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
...
+        if ((*(const int32_t *) (_mask + i)) >> 31) {
+            _c[i] = _b[i];
+        } else {
+            _c[i] = _a[i];
+        }
+    }
+
+    __m128 a = _mm_load_ps(_a);
+    __m128 b = _mm_load_ps(_b);
+    __m128 mask = _mm_load_ps(_mask);
```
marktwtn, line 5286: Replace with `do_mm_load_ps`.

marktwtn (changes requested):
Please rewrite the commit message. It is quite confusing.
I do not see anything that related to the accuracy.
I believe that the original implementation is wrong, so you might need to change the `refactor` to `fix`.

What would happen if you run the tests with the original implementation?

HowJMay:
> Please rewrite the commit message. It is quite confusing.
> I do not see anything that related to the accuracy.
> I believe that the original implementation is wrong, so you might need to change the `refactor` to `fix`.
> 
> What would happen if you run the tests with the original implementation?

In the original implementation, the last a few digits are wrong, so I thought the problem that the original implementation has is relating to accuracy. But, yeah, I should change the git commit message to fix instead of refactor

jserv:
> In the original implementation, the last a few digits are wrong, so I thought the problem that the original implementation has is relating to accuracy. But, yeah, I should change the git commit message to fix instead of refactor

Change the subject of this pull request accordingly.

marktwtn:
The request changes are not handled.

jserv (approved):
The subject is misleading. Revise it.

marktwtn (approved):

PR #257, opened 2021-01-02
feat: Add intrinsics _mm_blendv_pd

HowJMay:


tests/impl.h
```diff
@@ -294,6 +294,7 @@
     TYPE(mm_blend_epi16)              \
     TYPE(mm_blendv_epi8)              \
     TYPE(mm_blendv_ps)                \
+    TYPE(mm_blendv_pd)                \
```
marktwtn, line 297: The order is wrong.

tests/impl.cpp
```diff
@@ -5268,6 +5268,33 @@ result_t test_mm_blendv_ps(const SSE2NEONTestImpl &impl, uint32_t i)
     return TEST_UNIMPL;
 }
 
+result_t test_mm_blendv_pd(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+    const double _mask[] = {(double) impl.mTestFloats[0],
+                            (double) impl.mTestFloats[1]};
```
marktwtn, line 5276: Replace `0` and `1` with `i` and `i+1`.

tests/impl.cpp
```diff
@@ -5268,6 +5268,33 @@ result_t test_mm_blendv_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+        if ((*(const int64_t *) (_mask + i)) >> 63) {
+            _c[i] = _b[i];
+        } else {
+            _c[i] = _a[i];
+        }
+    }
+
+    __m128d a = do_mm_load_pd(_a);
+    __m128d b = do_mm_load_pd(_b);
+    __m128d mask = do_mm_load_pd((const double *) _mask);
```
marktwtn, line 5291: Is the casting necessary?

sse2neon.h
```diff
@@ -5078,6 +5081,26 @@ FORCE_INLINE __m128 _mm_blendv_ps(__m128 a, __m128 b, __m128 mask)
...
 }
 
+// Blend packed double-precision (64-bit) floating-point elements from a and b
+// using mask, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_blendv_pd
+FORCE_INLINE __m128d _mm_blendv_pd(__m128d _a, __m128d _b, __m128d _mask)
+{
+#if defined(__aarch64__)
+    uint64x2_t mask =
+        vreinterpretq_u64_s64(vshrq_n_s64(vreinterpretq_s64_m128d(_mask), 63));
```
marktwtn, line 5091: The code is the same as arm 32-bit.
Extract them.

marktwtn (changes requested):

sse2neon.h
HowJMay: The 64-bit implementation uses `vbslq_f64`
The 32-bit implementation uses `vbslq_u64`
Should I unify them to `vbslq_u64`?

marktwtn: Using `vbslq_f64` is fine.

HowJMay: I can extract this line only
```
uint64x2_t mask =
        vreinterpretq_u64_s64(vshrq_n_s64(vreinterpretq_s64_m128d(_mask), 63));
```

marktwtn: That is what I meant.

marktwtn (approved):

PR #260, opened 2021-01-03
feat: Add intrinsics _mm_mul_sd

HowJMay:


tests/impl.h
```diff
@@ -92,6 +92,7 @@
     TYPE(mm_movelh_ps)                \
     TYPE(mm_movemask_ps)              \
     TYPE(mm_mul_ps)                   \
+    TYPE(mm_mul_sd)                   \
```
marktwtn, line 95: The order and the instruction type are wrong again.
Please be careful.

marktwtn (changes requested):

tests/impl.cpp
```diff
@@ -1894,6 +1894,19 @@ result_t test_mm_mul_ps(const SSE2NEONTestImpl &impl, uint32_t i)
     return validateFloat(c, dx, dy, dz, dw);
 }
 
+result_t test_mm_mul_sd(const SSE2NEONTestImpl &impl, uint32_t i)
```
marktwtn, line 1897: Place it with the correct order like `tests/impl.h`.

marktwtn: The function definition is duplicated.

marktwtn (changes requested):

tests/impl.cpp
```diff
@@ -6680,7 +6680,15 @@ result_t test_mm_movemask_pd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+    double dx = _a[0] * _b[0];
+    double dy = _a[1];
+
+    __m128d a = do_mm_load_pd(_a);
+    __m128d b = do_mm_load_pd(_b);
+    __m128d c = _mm_mul_sd(a, b);
+    return validateDouble(c, dx, dy);
 }
```
marktwtn, line 6692: Place in the correct order and type.

marktwtn (changes requested):

marktwtn (approved):

PR #263, opened 2021-01-06
feat: Declare unimplemented intrinsics

HowJMay:
Declare all the intrinsics in impl.h and add the corresponding tests
in impl.cpp with returning `TEST_UNIMPL`. The order of declarations in
impl.h is based on the order on official guide.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -6149,6 +6149,7 @@ result_t test_mm_popcnt_u64(const SSE2NEONTestImpl &impl, uint32_t i)
     return TEST_SUCCESS;
 }
 
+/* Unimplemented */
 result_t test_mm_shuffle_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
     return TEST_UNIMPL;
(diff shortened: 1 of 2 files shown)


tests/impl.cpp
```diff
@@ -6197,6 +6197,853 @@ result_t test_mm_shufflelo_epi16_function(const SSE2NEONTestImpl &impl,
     return TEST_UNIMPL;
 }
```
jserv, line 6199: Add the corresponding classifications here. e.g. SSE2, SSE3, etc.

jserv:
After this change is merged, can we generate the reports of intrinsics coversge?

tests/impl.cpp
marktwtn: Currently, the classification comments have been added in `tests/impl.cpp`.
It is just like what we did in `tests/impl.h`.
@HowJMay you need to reorder it like you did in `test.impl.h`.

HowJMay: Reordering it is quite time-consuming task. I prefer doing that only in header file tho.

HowJMay:
> After this change is merged, can we generate the reports of intrinsics coversge?

I think we can present coverage report after merging this PR

tests/impl.cpp
```diff
@@ -6197,6 +6197,853 @@ result_t test_mm_shufflelo_epi16_function(const SSE2NEONTestImpl &impl,
...
+    return TEST_UNIMPL;
+}
+
+result_t test_mm_cmpistrz(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    return TEST_UNIMPL;
+}
+
+
+
```
marktwtn, line 7046: Remove the lines.

tests/impl.cpp
marktwtn: Nonetheless, it will be done in the future.
At least add the comment `// Unimplement`.

marktwtn (approved):

tests/impl.cpp
```diff
@@ -6149,6 +6149,7 @@ result_t test_mm_popcnt_u64(const SSE2NEONTestImpl &impl, uint32_t i)
     return TEST_SUCCESS;
 }
 
+/* Unimplement */
```
jserv, line 6152: Should be "Unimplemented"

PR #264, opened 2021-01-06
Add intrinsics _mm_cvtps_pi16

JishinMaster:
Dear all,

I have implemented _mm_cvtps_pi16.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -4973,6 +4973,18 @@ FORCE_INLINE __m128i _mm_cvtps_epi32(__m128 a)
 #endif
 }
 
+// Convert packed single-precision (32-bit) floating-point elements in a to
+// packed 16-bit integers, and store the results in dst. Note: this intrinsic
+// will generate 0x7FFF, rather than 0x8000, for input values between 0x7FFF and
+// 0x7FFFFFFF.
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtps_pi16
+FORCE_INLINE __m64 _mm_cvtps_pi16(__m128 a)
+{
+    return vreinterpret_m64_s16(
+        vmovn_s32(vreinterpretq_s32_m128i(_mm_cvtps_epi32(a))));
+}
+
 // Copy the lower 32-bit integer in a to dst.
 //
 //   dst[31:0] := a[31:0]
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -1362,7 +1362,23 @@ result_t test_mm_cvtpi8_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtps_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const float *_a = impl.mTestFloatPointer1;
+    float _b[4];
+    int16_t trun[4];
+
+    // Beyond int16_t range _mm_cvtps_pi16 function (both native and arm)
+    // do not behave the same as BankersRounding.
+    // Forcing the float input values to be in the int16_t range
+    // Dividing by 10.0f ensures (with the current data set) it,
+    // without forcing a saturation.
+    for (int j = 0; j < 4; j++) {
+        _b[j] = fabsf(_a[j]) > 32767.0f ? _a[j] / 10.0f : _a[j];
+        trun[j] = (int16_t)(bankersRounding(_b[j]));
+    }
+
+    __m128 b = do_mm_load_ps(_b);
+    __m64 ret = _mm_cvtps_pi16(b);
+    return validateInt16(ret, trun[0], trun[1], trun[2], trun[3]);
 }
 
 result_t test_mm_cvtps_pi32(const SSE2NEONTestImpl &impl, uint32_t i)


JishinMaster:
the x86 test fail with clang-10/gcc-10.
It is related to what was observed for _mm_cvtps_epi32.
I used vcvtq_s32_f32 which rounds toward zero. The correct way, for AArch64, would be to use vcvtnq_s32_f32, which is not available for Armv-7.

sse2neon.h
```diff
@@ -4884,6 +4884,21 @@ FORCE_INLINE __m128i _mm_cvtps_epi32(__m128 a)
...
+//
+//   r0 := (short) a0
+//   r1 := (short) a1
+//   r2 := (short) a2
+//   r3 := (short) a3
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtps_pi16
+FORCE_INLINE __m64 _mm_cvtps_pi16(__m128 a)
+{
+    return vreinterpret_m64_s16(
```
jserv, line 4898: For Aarch64, would it be better to use `vmovn_s32(vcvtq_s32_f32(vrndiq_f32(a)))`?

JishinMaster: In my lastest commit (868a0472abb7cf15eb85b69cbca858a6bd6ea4f2), I have improved things and I use _mm_cvtps_epi32 as the base for float to integer conversion. This function uses round to nearest directly (no need for 2 instructions).

jserv (changes requested):
Rebase the latest master branch.

sse2neon.h
```diff
@@ -4893,6 +4893,21 @@ FORCE_INLINE __m128i _mm_cvtps_epi32(__m128 a)
 #endif
 }
 
+// Converts the four single-precision, floating-point values of a to signed
```
jserv, line 4896: Take the description from https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtps_pi16
That is,
> Convert packed single-precision (32-bit) floating-point elements in a to packed 16-bit integers, and store the results in dst. Note: this intrinsic will generate 0x7FFF, rather than 0x8000, for input values between 0x7FFF and 0x7FFFFFFF.

jserv (approved):

tests/impl.cpp
```diff
@@ -6235,7 +6235,18 @@ result_t test_mm_shufflelo_epi16_function(const SSE2NEONTestImpl &impl,
 
 result_t test_mm_cvtps_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const float *_a = impl.mTestFloatPointer1;
+    float _b[4];
+    int16_t trun[4];
+    
+    for (int j = 0; j < 4; j++) {
+        _b[j] = fabsf(_a[j]) > 32767.0f ? _a[j] / 10.0f : _a[j];
```
marktwtn, line 6243: What is `_a[j] / 10.0f` for?

JishinMaster:
Neither the NEON version nor the original SSE behave well beyond the range of an int16_t.
It might be related to "Note: this intrinsic will generate 0x7FFF, rather than 0x8000, for input values between 0x7FFF and 0x7FFFFFFF.".

The test might be improved later. I did not want to just saturate the values, so I  divided the values by 10 since the test set does not seem to go that high.

tests/impl.cpp
marktwtn: @JishinMaster let's discuss under this conversation.

First, I would like to know the behaviour of the SSE when it is beyond the range of an `int16_t`.
That is, please give more description about `Neither the NEON version nor the original SSE behave well beyond the range of an int16_t.` .

I prefer the test should take the situation of `Note: this intrinsic will generate 0x7FFF, rather than 0x8000, for input values between 0x7FFF and 0x7FFFFFFF` into consideration.

Since divide with the value `10.0` is not intuitive, at least add the comment to describe the reason for doing this.

JishinMaster: If I remove the /10.0f patch, here is what I have on my Atom CPU :
In -68298.343750 Out -32768 Ref -2762
In 44244.390625 Out 32767 Ref -21292
In -61859.132812 Out -32768 Ref 3677
In 8441.164062 Out 8441 Ref 8441
In -68298.343750 Out -32768 Ref -2762
In 44244.390625 Out 32767 Ref -21292
In -61859.132812 Out -32768 Ref 3677
In 8441.164062 Out 8441 Ref 8441

The "Out" is the value after _mm_cvtps_pi16, and "Ref" is using (int16_t)bankersRounding.
For input values above the range of int16_t ( > 32767.0f or < -32768.0f), _mm_cvtps_pi16 saturates at either 32767 or 32768 (depending on the sign).
The aarch64 sse2neon port (using Qemu) gives the same results as the Atom CPU as long as the input value stay in the range of int16_t. Outside this range values may differ.

I will add a comment so it is clearer.

marktwtn (approved):

jserv:
@JishinMaster, can you rebase the latest master branch since recent commit already tweaked the order of test items?

JishinMaster:
Hi,
I rebased.

Could you please mention my simd_utils library in the Adoptions section?

Regards

jserv:
> Could you please mention my simd_utils library in the Adoptions section? 

DONE! It is my pleasure to collaborate with you.

PR #265, opened 2021-01-06
Add intrinsics _mm_sub_pd

JishinMaster:


sse2neon.h
```diff
@@ -2651,6 +2651,27 @@ FORCE_INLINE __m128i _mm_subs_epi16(__m128i a, __m128i b)
         vqsubq_s16(vreinterpretq_s16_m128i(a), vreinterpretq_s16_m128i(b)));
 }
 
+// Subtracts the two double-precision, floating-point values of a and b.
+//
+//   r0 := a0 - b0
+//   r1 := a1 - b1
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=sub_pd
```
marktwtn, line 2659: Replace with
```
// Subtract packed double-precision (64-bit) floating-point elements in b from packed double-precision (64-bit) floating-point elements in a, and store the results in dst.
// 
//   FOR j := 0 to 1
//     i := j*64
//     dst[i+63:i] := a[i+63:i] - b[i+63:i]
//   ENDFOR
//  https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=mm_sub_pd

```

tests/impl.h
```diff
@@ -121,6 +121,7 @@
     TYPE(mm_stream_ps)                \
     TYPE(mm_sub_ps)                   \
     TYPE(mm_sub_ss)                   \
+    TYPE(mm_sub_pd)                   \
```
marktwtn, line 124: The intrinsic type is `SSE2`.
Please place it in the correct classification.

marktwtn: Reference: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#techs=SSE2

tests/impl.cpp
```diff
@@ -2261,6 +2261,19 @@ result_t test_mm_sub_ss(const SSE2NEONTestImpl &impl, uint32_t i)
     return validateFloat(c, dx, dy, dz, dw);
 }
 
+result_t test_mm_sub_pd(const SSE2NEONTestImpl &impl, uint32_t i)
```
marktwtn, line 2264: Place it in the correct classification.

marktwtn (changes requested):
Please rebase the commits to a single one.

marktwtn (approved):

jserv:

PR #268, opened 2021-01-11
feat: Add intrinsics _mm_cmpeq_pd

HowJMay:


sse2neon.h
```diff
@@ -4277,6 +4284,24 @@ FORCE_INLINE __m128i _mm_cmpeq_epi8(__m128i a, __m128i b)
...
+// Compare packed double-precision (64-bit) floating-point elements in a and b
+// for equality, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpeq_pd
+FORCE_INLINE __m128d _mm_cmpeq_pd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    return vreinterpretq_m128d_u64(
+        vceqq_f64(vreinterpretq_f64_m128d(a), vreinterpretq_f64_m128d(b)));
+#else
+    // ARMv7 lacks vceqq_f64
```
jserv, line 4296: You don't have to mention `ARMv7 lacks vceqq_f64`.

tests/impl.cpp
```diff
@@ -2858,6 +2858,19 @@ result_t test_mm_cmpeq_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
...
+{
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+    uint64_t d0 = (_a[0] == _b[0]) ? 0xffffffffffffffff : 0x00;
+    uint64_t d1 = (_a[1] == _b[1]) ? 0xffffffffffffffff : 0x00;
+
+    __m128d a = do_mm_load_pd(_a);
+    __m128d b = do_mm_load_pd(_b);
+    __m128d c = _mm_cmpeq_pd(a, b);
+    return validateDouble(c, (double) d0, (double) d1);
```
marktwtn, line 2871: I wonder the casting is appropriate or not.
For example, if `d0` is `1`, then `(double) d0` would be `1.0`.
I think `1.0` would not be `0x0000000000000001`.

tests/impl.cpp
```diff
@@ -2858,6 +2858,19 @@ result_t test_mm_cmpeq_epi8(const SSE2NEONTestImpl &impl, uint32_t i)
                         d12, d13, d14, d15);
 }
 
+result_t test_mm_cmpeq_pd(const SSE2NEONTestImpl &impl, uint32_t i)
+{
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+    uint64_t d0 = (_a[0] == _b[0]) ? 0xffffffffffffffff : 0x00;
+    uint64_t d1 = (_a[1] == _b[1]) ? 0xffffffffffffffff : 0x00;
```
marktwtn, line 2866: Replace `0x00` with `0`.

PR #273, opened 2021-01-14
CI: Make sure the coding convention is checked

marktwtn:
The .ci/check-format script does not return error when
the clang-format-6.0 is not installed.
Currently, the commit makes sure the required software is installed.

The fixing of the script will be done in the future.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -52,5 +52,7 @@ jobs:
       - name: checkout code
         uses: actions/checkout@v2
       - name: coding convention
-        run: sh .ci/check-format.sh
+        run: |
+            sudo apt-get install -y clang-format-6.0
+            sh .ci/check-format.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/github_actions.yml
```diff
@@ -52,5 +52,7 @@ jobs:
       - name: checkout code
         uses: actions/checkout@v2
       - name: coding convention
-        run: sh .ci/check-format.sh
+        run: |
```
jserv, line 55: I am no expert of YAML. What is the pipe `I` used for?

marktwtn: In `Github Actions`, it means the start of the multiple commands.

If `|` is not used, we can only use a single command or wrap multiple commands into a single script and then invoke it.

jserv: Learnt!

PR #269, opened 2021-01-12
refactor: Rearrange impl.cpp as the same order as impl.h

HowJMay:
Rearrange impl.cpp as the same order as impl.h.

close #267

PR #274, opened 2021-01-17
feat: Implement intrinsics _mm_div_sd

HowJMay:


HowJMay:
depends on #269

sse2neon.h
```diff
@@ -3496,6 +3496,16 @@ FORCE_INLINE __m128d _mm_div_pd(__m128d a, __m128d b)
...
 }
 
+// Divide the lower double-precision (64-bit) floating-point element in a by the
+// lower double-precision (64-bit) floating-point element in b, store the result
+// in the lower element of dst, and copy the upper element from a to the upper
+// element of dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_div_sd
+FORCE_INLINE __m128d _mm_div_sd(__m128d a, __m128d b)
+{
+    return _mm_move_sd(a, _mm_div_pd(a, b));
```
jserv, line 3506: This implementation might be slightly heavier. For Aarch64, you can reimplement `_mm_div_sd` as following:
```c
float64x2_t tmp = vdivq_f64(vreinterpretq_f64_m128d(a), vreinterpretq_f64_m128d(b));
return vreinterpretq_m128d_f64(vsetq_lane_f64(vgetq_lane(vreinterpretq_f64_m128d(a), 1), tmp, 1));
```

Less instructions.

jserv (approved):

sse2neon.h
marktwtn: @jserv I wonder what if I change the implementation of `_mm_move_sd` from `vcombine + vget_low/high_f32` to `vext_f32`?
Then @HowJMay can use `_mm_move_sd` without modification.

jserv: Agree. Let's close this pull request and concentrate on `_mm_move_sd` changes.

marktwtn: I tried to implement `_mm_move_sd` with `vext_f32` and found out that it needs at least two `vext_f32` to do the same thing.
I think we should leave `_mm_move_sd` unchanged.

marktwtn (approved):

PR #276, opened 2021-01-24
implemented _MM_SET_ROUNDING_MODE

JishinMaster:
Dear All, 
I have implemented _MM_SET_ROUNDING_MODE using the information found on the [Arm Armv8-A Architecture Registers](https://developer.arm.com/docs/ddi0595/e/aarch32-system-registers/fpscr).
At the moment the test only fails for _MM_ROUND_TOWARD_ZERO on aarch32, the other ones seem to behave correctly.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -141,6 +141,10 @@
 #define _MM_FROUND_TO_ZERO 0x03
 #define _MM_FROUND_CUR_DIRECTION 0x04
 #define _MM_FROUND_NO_EXC 0x08
+#define _MM_ROUND_NEAREST 0x0000
+#define _MM_ROUND_DOWN 0x2000
+#define _MM_ROUND_UP 0x4000
+#define _MM_ROUND_TOWARD_ZERO 0x6000
 
 /* indicate immediate constant argument in a given range */
 #define __constrange(a, b) const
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -5247,6 +5251,71 @@ FORCE_INLINE __m128d _mm_blendv_pd(__m128d _a, __m128d _b, __m128d _mask)
 #endif
 }
 
+
+#if defined(__aarch64__)
+typedef struct {
```
jserv, line 5256: It seem that the structure definitions between Aarch64 and ARM32 had some overlapping. Can you rewrite as following?
```c
typedef struct {
    uint16_t res0;
    uint8_t res1 : 6;
    uint8_t bit22 : 1;
    uint8_t bit23 : 1;
    uint8_t res2;
#if defined(__aarch64__)
    uint32_t res3;
#endif
} fpcr_bitfield;

typedef union {
    fpcr_bitfield field;
#if defined(__aarch64__)
    uint64_t value;
#else
    uint32_t value;
#endif
} reg;

```

sse2neon.h
```diff
@@ -5247,6 +5251,71 @@ FORCE_INLINE __m128d _mm_blendv_pd(__m128d _a, __m128d _b, __m128d _mask)
...
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_MM_SET_ROUNDING_MODE
+FORCE_INLINE void _MM_SET_ROUNDING_MODE(int rounding)
+{
+    reg r;
+#if defined(__aarch64__)
+    asm volatile("mrs %0, FPCR" : "=r"(r.value)); /* read */
+#else
+    asm volatile("vmrs %0, FPSCR" : "=r"(r.value)); /* read */
+#endif
+    if (rounding == _MM_ROUND_TOWARD_ZERO) {
```
jserv, line 5299: Rewrite with switch-case.

JishinMaster: Done!

I don't see what's wrong with CI.

tests/impl.h
```diff
@@ -158,6 +158,7 @@
     TYPE(mm_unpackhi_ps)              \
     TYPE(mm_unpacklo_ps)              \
     TYPE(mm_xor_ps)                   \
+    TYPE(mm_set_rounding_mode)        \
```
marktwtn, line 161: Place it right after the intrinsic `_mm_set_ps1`.

tests/impl.cpp
```diff
@@ -2502,6 +2502,49 @@ result_t test_mm_xor_ps(const SSE2NEONTestImpl &impl, uint32_t i)
                          *((float *) &d2), *((float *) &d3));
 }
 
+result_t test_mm_set_rounding_mode(const SSE2NEONTestImpl &impl, uint32_t i)
```
marktwtn, line 2505: Place it right after `test_mm_set_ps1`.

sse2neon.h
```diff
@@ -5247,6 +5251,68 @@ FORCE_INLINE __m128d _mm_blendv_pd(__m128d _a, __m128d _b, __m128d _mask)
...
+        break;
+    default:  //_MM_ROUND_NEAREST
+        r.field.bit22 = 0;
+        r.field.bit23 = 0;
+    }
+
+#if defined(__aarch64__)
+    asm volatile("msr FPCR, %0" ::"r"(r)); /* write */
+#else
+    asm volatile("vmsr FPSCR, %0" ::"r"(r));        /* write */
```
marktwtn, line 5312: Remove the extra blank between the assembly and the comment.

tests/impl.cpp
```diff
@@ -2502,6 +2502,49 @@ result_t test_mm_xor_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    res_toward_zero =
+        validateFloatEpsilon(c, ((float *) &b)[0], ((float *) &b)[1],
+                             ((float *) &b)[2], ((float *) &b)[3], 5.0f);
+
+    _MM_SET_ROUNDING_MODE(_MM_ROUND_DOWN);
+    b = _mm_round_ps(a, _MM_FROUND_CUR_DIRECTION);
+    c = _mm_round_ps(a, _MM_FROUND_TO_NEG_INF | _MM_FROUND_NO_EXC);
+    res_to_neg_inf =
+        validateFloatEpsilon(c, ((float *) &b)[0], ((float *) &b)[1],
+                             ((float *) &b)[2], ((float *) &b)[3], 5.0f);
```
marktwtn, line 2524: What is the reason of setting the value to `5.0f`?

JishinMaster: Not in particular, I took the same value than the test_mm_round_ps

JishinMaster:
The test on X86 failed but it came from the aarch32 _MM_ROUND_TOWARD_ZERO.
I have commented it out at the moment.

Would it be clearer to comment out the test and the port in sse2neon.h? (disable only the _MM_ROUND_TOWARD_ZERO so users can still use the function on aarch32 with the other rounding mode)

jserv (changes requested):
Since _MM_ROUND_TOWARD_ZERO has not the expected behavior on aarch32, the git commit message should address the limitation.

sse2neon.h
jserv: Can you tweak the struct definitions?

jserv (changes requested):
The subject of git commit message is a bit longer. Limit the subject line to 60 characters and 
wrap the body at 72 characters.

JishinMaster:
What commit message do you want me to use?

jserv:
> What commit message do you want me to use?

I meant the subject of GIT commit message. It was a bit longer.

sse2neon.h
jserv: It was my fault. I was not aware of the updated code.

jserv:

PR #282, opened 2021-01-27
feat: Implement _mm_storer_ps

marktwtn:


sse2neon.h
```diff
@@ -834,6 +834,23 @@ FORCE_INLINE void _mm_store_ps1(float *p, __m128 a)
...
+//   MEM[mem_addr+31:mem_addr] := a[127:96]
+//   MEM[mem_addr+63:mem_addr+32] := a[95:64]
+//   MEM[mem_addr+95:mem_addr+64] := a[63:32]
+//   MEM[mem_addr+127:mem_addr+96] := a[31:0]
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storer_ps
+FORCE_INLINE void _mm_storer_ps(float *p, __m128 a)
+{
+    float32x4_t tmp = vrev64q_f32(vreinterpretq_f32_m128(a));
+    float32x4_t rev = vcombine_f32(vget_high_f32(tmp), vget_low_f32(tmp));
```
jserv, line 850: You can reimplement as following:
```c
rev = vextq_f32(tmp, tmp, 2);
```

marktwtn: That is much better!

PR #286, opened 2021-02-01
feat: Implement intrinsics _mm_set_pd1

HowJMay:


tests/impl.cpp
```diff
@@ -4411,7 +4411,14 @@ result_t test_mm_set_pd(const SSE2NEONTestImpl &impl, uint32_t i)
...
 {
-    return TEST_UNIMPL;
+    const double _a = impl.mTestFloats[i];
+
+    double f0 = _a;
+    double f1 = _a;
+
+    __m128d a = _mm_set_pd1(_a);
+
+    return validateDouble(a, f0, f1);
```
marktwtn, line 4421: I think we can just write it as `validateDouble(a, _a, _a)`?

sse2neon.h
```diff
@@ -805,6 +805,14 @@ FORCE_INLINE __m128d _mm_set_pd(double e1, double e0)
...
 }
 
+// Broadcast double-precision (64-bit) floating-point value a to all elements of
+// dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_set_pd1
+FORCE_INLINE __m128d _mm_set_pd1(double a)
+{
+    return _mm_set_pd(a, a);
+}
+
```
marktwtn, line 815: We have already implemented `_mm_set1_pd`.
I think you can define the macro like
```
#define _mm_set_pd1 _mm_set1_pd
```

marktwtn (approved):

PR #288, opened 2021-02-01
test: Call tests with the same implementation

HowJMay:
_mm_ucomieq_ss, _mm_ucomige_ss, _mm_ucomigt_ss,_mm_ucomile_ss
,_mm_ucomilt_ss, _mm_ucomineq_ss share the same implementation of
_mm_comieq_ss, _mm_comige_ss, _mm_comigt_ss,_mm_comile_ss, _mm_comilt_ss
, _mm_comineq_ss.

Therefore, we can call the tests of _mm_comieq_ss, _mm_comige_ss,
_mm_comigt_ss,_mm_comile_ss, _mm_comilt_ss, _mm_comineq_ss directly.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -4699,11 +4699,11 @@ FORCE_INLINE int _mm_comineq_ss(__m128 a, __m128 b)
 
 // according to the documentation, these intrinsics behave the same as the
 // non-'u' versions.  We'll just alias them here.
-#define _mm_ucomilt_ss _mm_comilt_ss
-#define _mm_ucomile_ss _mm_comile_ss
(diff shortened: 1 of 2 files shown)


marktwtn:
Based on the document, it still has a little difference.
The intrinsics come with `u` says `This instruction will not signal an exception for QNaNs.`

@jserv I think we can merge the pull request without any problem since we do not handle the exception yet.
However, in the future we might need to separate the tests.

jserv:
> Based on the document, it still has a little difference.
> The intrinsics come with `u` says `This instruction will not signal an exception for QNaNs.`

The git commit messages should address the exceptions.

HowJMay:
I have revised the commit

PR #289, opened 2021-02-01
feat: Implement _mm_blend_ps with test

BarabashkaD:
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_blend_ps
tested on NVIDIA XAVIAR
test for x86 disable due immediate 4 bit restriction

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -6175,6 +6175,9 @@ result_t test_mm_blend_pd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_blend_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+// gcc and clang can't compile call to _mm_blend_ps with 3rd argument as integer type due 4 bit constatnt size limitation
+// and test framework doesn't support compile time constant
+// but on target platform its complie , run and passed    
 #if __x86_64__
     return TEST_UNIMPL;
 #else
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -6175,12 +6175,6 @@ result_t test_mm_blend_pd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_blend_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-// gcc and clang can't compile call to _mm_blend_ps with 3rd argument as integer type due 4 bit constatnt size limitation
-// and test framework doesn't support compile time constant
-// but on target platform its complie , run and passed    
-#if __x86_64__
-    return TEST_UNIMPL;
-#else
     const float *_a = impl.mTestFloatPointer1;
     const float *_b = impl.mTestFloatPointer2;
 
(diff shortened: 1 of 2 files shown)


marktwtn:
Where is the `immediate version of _mm_blendv_ps`?

tests/impl.cpp
```diff
@@ -6175,7 +6175,30 @@ result_t test_mm_blend_pd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_blend_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+#if __x86_64__
```
jserv, line 6178: Why did you keep `UNIMPL`?

BarabashkaD:
@marktwtn 
_mm_blend_ps is a immediate version of _mm_blendv_ps
__m128 _mm_blend_ps (__m128 a, __m128 b, const int imm8)
vs
__m128 _mm_blendv_ps (__m128 a, __m128 b, __m128 mask)
difference in 3rd argument that in first version should be 4 bit constant value
and in second version vector register

tests/impl.cpp
BarabashkaD: gcc and clang can't compile  call to _mm_blend_ps with 3rd argument as integer type due 4 bit size limitation
and test framework doesn't support compile time constant 
but on target platform its complie , run and passed

marktwtn:
> @marktwtn
> _mm_blend_ps is a immediate version of _mm_blendv_ps
> __m128 _mm_blend_ps (__m128 a, __m128 b, const int imm8)
> vs
> __m128 _mm_blendv_ps (__m128 a, __m128 b, __m128 mask)
> difference in 3rd argument that in first version should be 4 bit constant value
> and in second version vector register

I think you can simplify the commit header with `feat: Implement _mm_blend_ps with test`

tests/impl.cpp
jserv: @marktwtn, is it feasible to extend the existing testing framework for the above considerations?

BarabashkaD:
GUys whats up ? ;-) can your approve it ?

tests/impl.cpp
jserv: Add the above explanation of excluding x86_64 in the comment of this test case. Later, we might overcome it.

BarabashkaD: comment as is its ok ?

BarabashkaD: BTW 
due its only 4 bits constant 
I can add test with switch case for all possible permutation
if you  prefer it

jserv: Adding all possible permutations sounds great to me.

BarabashkaD: Done

jserv:
Thank @BarabashkaD for contributing!

tests/impl.cpp
marktwtn: The constant problem is left to be fixed in the future.
However, switch case is a good way to do this.
I will apply the same technique in other tests instead of using fixed constant.

PR #290, opened 2021-02-03
fix: Fix _mm_alignr_epi8 implementation and add test

marktwtn:
The original version does not accept the third argument of
_mm_alignr_epi8 larger than 15 with GCC.

However, the fixed version still has the problem on ARM platform with
Clang-10.
Clang-10 would check the third argument of vextq_u8 in different if-else
condition and generate error message "argument value is outside the
valid range [0, 15]".

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1522,6 +1522,37 @@ FORCE_INLINE __m64 _mm_abs_pi8(__m64 a)
     return vreinterpret_m64_s8(vabs_s8(vreinterpret_s8_m64(a)));
 }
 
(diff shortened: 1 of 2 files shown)


jserv:
> However, the fixed version still has the problem on ARM platform with
> Clang.

You should mention the explicit clang version you have tested with.

marktwtn:
> > However, the fixed version still has the problem on ARM platform with
> > Clang.
> 
> You should mention the explicit clang version you have tested with.

The version has been updated.

PR #292, opened 2021-02-05
avoid double declaration for vld1q_u8_x4

sebpop:
The missing intrinsic has been added to gcc 7 on Amazon Linux 2, to gcc-8.5, and to gcc-9.4.
Using these compilers to compile sse2neon.h will start failing with multiple definitions of `vld1q_u8_x4`.
I see this error on Amazon Linux 2 with gcc-7.3.1-12:

```
g++ -o tests/common.o -Wall -Wcast-qual -I. -march=armv8-a+fp+simd+crc -std=gnu++14 -c -MMD -MF tests/common.o.d tests/common.cpp
In file included from tests/common.h:5:0,
                 from tests/common.cpp:1:
./sse2neon.h: In function ‘uint8x16x4_t vld1q_u8_x4(const uint8_t*)’:
./sse2neon.h:322:27: error: ‘uint8x16x4_t vld1q_u8_x4(const uint8_t*)’ redeclared inline without ‘gnu_inline’ attribute
 FORCE_INLINE uint8x16x4_t vld1q_u8_x4(const uint8_t *p)
                           ^~~~~~~~~~~
In file included from ./sse2neon.h:105:0,
                 from tests/common.h:5,
                 from tests/common.cpp:1:
/usr/lib/gcc/aarch64-redhat-linux/7/include/arm_neon.h:18033:1: note: ‘uint8x16x4_t vld1q_u8_x4(const uint8_t*)’ previously defined here
 vld1q_u8_x4 (const uint8_t *__a)
 ^~~~~~~~~~~
```

I tested this patch with gcc-7.3.1-12 on Amazon Linux 2, and on Ubuntu 20.04 with gcc-9.3 and gcc-10.2.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -317,9 +317,11 @@ typedef union ALIGN_STRUCT(16) SIMDVec {
 /* Backwards compatibility for compilers with lack of specific type support */
 
 // Older gcc does not define vld1q_u8_x4 type
-#if defined(__GNUC__) && !defined(__clang__)
-#if __GNUC__ <= 9
-FORCE_INLINE uint8x16x4_t vld1q_u8_x4(const uint8_t *p)
+#if defined(__GNUC__) && !defined(__clang__) &&   \
+    ((__GNUC__ == 10 && (__GNUC_MINOR__ <= 1)) || \
+     (__GNUC__ == 9 && (__GNUC_MINOR__ <= 3)) ||  \
+     (__GNUC__ == 8 && (__GNUC_MINOR__ <= 4)) || __GNUC__ <= 7)
+FORCE_INLINE uint8x16x4_t sse2neon_vld1q_u8_x4(const uint8_t *p)
 {
     uint8x16x4_t ret;
     ret.val[0] = vld1q_u8(p + 0);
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -317,9 +317,11 @@ typedef union ALIGN_STRUCT(16) SIMDVec {
...
 
 // Older gcc does not define vld1q_u8_x4 type
-#if defined(__GNUC__) && !defined(__clang__)
-#if __GNUC__ <= 9
-FORCE_INLINE uint8x16x4_t vld1q_u8_x4(const uint8_t *p)
+#if defined(__GNUC__) && !defined(__clang__) &&   \
+    ((__GNUC__ == 10 && (__GNUC_MINOR__ <= 1)) || \
+     (__GNUC__ == 9 && (__GNUC_MINOR__ <= 3)) ||  \
+     (__GNUC__ == 8 && (__GNUC_MINOR__ <= 4)) || __GNUC__ <= 7)
+FORCE_INLINE uint8x16x4_t vld1q_u8_x4_1(const uint8_t *p)
```
jserv, line 324: For consistency, rename this function to `sse2neon_vld1q_u8_x4`, which implies the implementation specific wrapper.

sebpop: Done.  Thanks for pointing this out.

PR #297, opened 2021-02-07
test: Add test for _mm_div_ps

HowJMay:


HowJMay:
depends on the new validation function in #298

PR #298, opened 2021-02-07
feat: Add test for _mm_div_ss

HowJMay:


HowJMay:
@jserv @marktwtn I have implemented a new validation function which validate the result by error. I think an error is more straightforward to validate two floating point result

tests/impl.cpp
```diff
@@ -1594,7 +1594,26 @@ result_t test_mm_div_ps(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    float d1 = _a[1];
+    float d2 = _a[2];
+    float d3 = _a[3];
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 b = do_mm_load_ps(_b);
+    __m128 c = _mm_div_ss(a, b);
+
+#if defined(__arm__) && !defined(__aarch64__)
+    // The implementation of `_mm_div_ps()` on ARM 32bit doesn't use `DIV`
```
jserv, line 1610: Don't use  backticks (`) in comments. Instead, use typical quotes.

tests/common.cpp
```diff
@@ -318,6 +318,25 @@ result_t validateFloatEpsilon(__m128 a,
...
+                            float f1,
+                            float f2,
+                            float f3,
+                            float err)
+{
+    const float *t = (const float *) &a;
+    float df0 = fabsf((t[0] - f0) / f0);
+    float df1 = fabsf((t[1] - f1) / f1);
+    float df2 = fabsf((t[2] - f2) / f2);
+    float df3 = fabsf((t[3] - f3) / f3);
```
marktwtn, line 332: I wonder the purpose of the division.

marktwtn: I mean, I am not convinced by the description `I think an error is more straightforward`.
Could you give more detail or your thought of the validation?

HowJMay: I think the error of the two input is more representative than the difference of the two input, since the two input may be a relatively small number, in which case, a small difference would become relatively big, and the test can't pass in this situation.

marktwtn: In contrast, how about the situation when the two inputs are large number (float maximum: 3.4×10^38)?

HowJMay: I think we should use a proportional value which varies against the inputs numbers to validate the test result. Therefore, under the situation that the 2 inputs are large numbers, we still guarantee the accuracy of the result.
Instead, if using an absolute difference of validate the test result, than the accuracy would vary against the values of the input numbers which is inappropriate and make the result unpredictable.

marktwtn (approved):

PR #281, opened 2021-01-26
feat: Implement _mm_castsi128_pd

marktwtn:

PR #307, opened 2021-02-12
feat: Implement _mm_shuffle_pd

marktwtn:


sse2neon.h
```diff
@@ -2157,6 +2157,38 @@ FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
...
 
+// Shuffle double-precision (64-bit) floating-point elements using the control
+// in imm8, and store the results in dst.
+//
+//   dst[63:0] := (imm8[0] == 0) ? a[63:0] : a[127:64]
+//   dst[127:64] := (imm8[1] == 0) ? b[63:0] : b[127:64]
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_pd
+#if defined(__aarch64__)
+#if __has_builtin(__builtin_shufflevector)
```
jserv, line 2168: If we are using `__builtin_shufflevector`, both Arm and Aarch64 can use the same implementation, right?

marktwtn: Not really.
`ARM` does not support data type `float64_t`.

However, the implementation of `ARM` can be applied on `AArch64`, too.

jserv: > However, the implementation of `ARM` can be applied on `AArch64`, too.

Did you mean that we can have single non-shufflevector implementation?

marktwtn: We can have single shufflevector and non-shufflevector implementation.
For example, we can have the following code only.
```c
#if __has_builtin(__builtin_shufflevector)
#define _mm_shuffle_pd(a, b, imm8)                                          \
    vreinterpretq_m128d_s64(__builtin_shufflevector(                        \
        vreinterpretq_s64_m128d(a), vreinterpretq_s64_m128d(b), imm8 & 0x1, \
        ((imm8 & 0x2) >> 1) + 2))
#else
#define _mm_shuffle_pd(a, b, imm8)                                     \
    _mm_castsi128_pd(_mm_set_epi64x(                                   \
        vgetq_lane_s64(vreinterpretq_s64_m128d(b), (imm8 & 0x2) >> 1), \
        vgetq_lane_s64(vreinterpretq_s64_m128d(a), imm8 & 0x1)))
#endif
```
It would have not problem on the `AArch64`.

jserv: The above looks great.

marktwtn: I wonder whether we should use the same implementation or use the more suitable one?
For example, we have previous advise like [this](https://github.com/DLTcollab/sse2neon/pull/281/files#r564356730).

jserv: Let's follow KISS, which an acronym for keep it simple and stupid, design principle.

PR #311, opened 2021-02-14
Fix warnings about redefinition, uninitialized variables and unknown pragmas

brechtvl:
For embree/embree#316 and Blender integration to build without warnings.


jserv:
Thank @brechtvl for the great work! Look forward to Arm support of embree and Blender.

PR #317, opened 2021-02-19
feat: Implement _mm_cvtepi32_pd

marktwtn:


tests/impl.cpp
```diff
@@ -3497,7 +3497,15 @@ result_t test_mm_comineq_sd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtepi32_pd(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = impl.mTestIntPointer1;
+    __m128i a = do_mm_load_ps(_a);
+    double trun[2];
+    for (uint32_t i = 0; i < 2; i++) {
```
jserv, line 3503: It is not worthy to use loop here. Instead, you can initialize `turn` with values.
```c
double turn[] = {(double) _a[0], (double) _a[1]};
```

PR #319, opened 2021-02-20
feat: Add test for _mm_rsqrt_ps

HowJMay:


tests/impl.cpp
```diff
@@ -2172,7 +2172,17 @@ result_t test_mm_rcp_ss(const SSE2NEONTestImpl &impl, uint32_t i)
...
+
+    float f0 = 1 / sqrt(_a[0]);
+    float f1 = 1 / sqrt(_a[1]);
+    float f2 = 1 / sqrt(_a[2]);
+    float f3 = 1 / sqrt(_a[3]);
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 c = _mm_rsqrt_ps(a);
+
+    return validateFloatError(c, f0, f1, f2, f3, 0.01f);
```
jserv, line 2185: You shall explain the value `0.01f`.

jserv: I was not convinced. Why `1%`?

HowJMay: 1% is the smallest error we can achieve under the current implementation.

jserv: I defer to the review of @marktwtn .

marktwtn: It is hard to decide the relative error when it comes to division and square root.
I do not have any good advice.

We might accept the current relative error and it needs to be considered carefully in the future pull request.

tests/impl.cpp
```diff
@@ -2172,7 +2172,19 @@ result_t test_mm_rcp_ss(const SSE2NEONTestImpl &impl, uint32_t i)
...
+
+    float f0 = 1 / sqrt(_a[0]);
+    float f1 = 1 / sqrt(_a[1]);
+    float f2 = 1 / sqrt(_a[2]);
+    float f3 = 1 / sqrt(_a[3]);
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 c = _mm_rsqrt_ps(a);
+
+    // Here we ensure `_mm_rsqrt_ps()`'s error is under 1% compare to the C
```
jserv, line 2185: Fix the grammar error.

tests/impl.cpp
jserv: Later, we might integrate the precise `rsqrt` implementation. At the moment, we have to address and explain the acceptable error rate.

HowJMay: So should I open another PR which address the acceptable error here and leave this PR in current implementation

jserv: Simply do it here.

sse2neon.h
```diff
@@ -4032,6 +4032,7 @@ FORCE_INLINE __m128 _mm_sqrt_ss(__m128 in)
 
 // Computes the approximations of the reciprocal square roots of the four
 // single-precision floating point values of in.
+// The current precision is 1% error.
```
HowJMay, line 4035: @jserv I added it here

jserv: Mention in git commit message as well.

marktwtn (approved):

PR #327, opened 2021-02-26
feat: Implement _mm_cvtsd_si64[x]

HowJMay:
The behaviour of _mm_cvtsd_si64x and _mm_cvtsi128_si64x are the same
as the behaviour of _mm_cvtsd_si64 and _mm_cvtsi128_si64, respectively.

Therefore, we can run the test of the non-x-suffix.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -5731,6 +5731,23 @@ FORCE_INLINE double _mm_cvtsd_f64(__m128d a)
 #endif
 }
 
+// Convert the lower double-precision (64-bit) floating-point element in a to a
+// 64-bit integer, and store the result in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtsd_si64
+FORCE_INLINE int64_t _mm_cvtsd_si64(__m128d a)
+{
+#if defined(__aarch64__)
+    return (int64_t) vcvt_s64_f64(vget_low_f64(vreinterpretq_f64_m128d(a)));
+#else
+    return (int64_t)((double *) &a)[0];
+#endif
+}
+
+// Convert the lower double-precision (64-bit) floating-point element in a to a
+// 64-bit integer, and store the result in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtsd_si64x
+#define _mm_cvtsd_si64x(a) _mm_cvtsd_si64(a)
+
 // Convert packed single-precision (32-bit) floating-point elements in a to
 // packed double-precision (64-bit) floating-point elements, and store the
 // results in dst.
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -3674,7 +3674,7 @@ result_t test_mm_cvtsd_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtsd_si64x(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    return test_mm_cvtsd_si64(impl, i);
 }
 
 result_t test_mm_cvtsd_ss(const SSE2NEONTestImpl &impl, uint32_t i)
@@ -3708,7 +3708,7 @@ result_t test_mm_cvtsi128_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtsi128_si64x(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    return test_mm_cvtsi128_si64(impl, i);
 }
 
 result_t test_mm_cvtsi32_sd(const SSE2NEONTestImpl &impl, uint32_t i)
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -3669,7 +3669,24 @@ result_t test_mm_cvtsd_si32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtsd_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    // NEON intrinsic "vcvt_s64_f64()" uses rounding mode "Rounding To Zero"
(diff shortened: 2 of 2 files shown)


jserv (changes requested):
Change the subject to "Implement _mm_cvtsd_f64[x]"

marktwtn:
The commit header should be `_mm_cvtsd_si64 and _mm_cvtsd_si64x`.

sse2neon.h
```diff
@@ -5692,6 +5692,23 @@ FORCE_INLINE double _mm_cvtsd_f64(__m128d a)
...
+// Convert the lower double-precision (64-bit) floating-point element in a to a
+// 64-bit integer, and store the result in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtsd_si64
+FORCE_INLINE int64_t _mm_cvtsd_si64(__m128d a)
+{
+#if defined(__aarch64__)
+    return vgetq_lane_s64(vreinterpretq_s64_m128d(a), 0);
+#else
+    return ((int64_t *) &a)[0];
+#endif
```
marktwtn, line 5704: The implementation is wrong.
It just treats the data store in `double` as `int64_t`.
The data itself is not changed.

What you expect is like `3.0` -> `3`.
However, the `3.0` data stored in double would not be `0x03`.

tests/impl.cpp
```diff
@@ -3636,12 +3636,18 @@ result_t test_mm_cvtsd_si32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtsd_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    int64_t d = (int64_t) round(_a[0]);
```
marktwtn, line 3640: I am not sure whether the `round()` is necessary or not.

tests/impl.cpp
```diff
@@ -3669,12 +3669,23 @@ result_t test_mm_cvtsd_si32(const SSE2NEONTestImpl &impl, uint32_t i)
...
 result_t test_mm_cvtsd_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+#if defined(__x86_64__)
+    int64_t _c = (int64_t) round(_a[0]);
+#else
+    int64_t _c = (int64_t) _a[0];
+#endif
+
```
jserv, line 3678: Remove this blank line.

tests/impl.cpp
```diff
@@ -3669,12 +3669,22 @@ result_t test_mm_cvtsd_si32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cvtsd_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+#if defined(__x86_64__)
+    int64_t _c = (int64_t) round(_a[0]);
```
jserv, line 3674: The use of `round` is not generic here. You should explain.

marktwtn:
> The commit header should be `_mm_cvtsd_si64 and _mm_cvtsd_si64x`.

Use the correct commit header.

tests/impl.cpp
HowJMay: I think we set the different rounding mode in NEON and SSE.

### _mm_cvtsd_si64:
[`Converts the lower DP FP value of a to a 64-bit signed integer value according to the current rounding mode.`](
https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/intrinsics/intrinsics-for-intel-streaming-simd-extensions-2-intel-sse2/integer-intrinsics/conversion-intrinsics-1.html)

### vcvt_s64_f64
`vcvt_s64_f64` is using rounding to zero.
https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vcvt_s64_f64

jserv: Add comments in the test suite.

marktwtn: IEEE 754 has 5 rounding rules, which are

* Round to nearest, ties to even
* Round to nearest, ties away from zero
* Round toward 0
* Round toward +∞
* Round toward -∞

Reference: https://en.wikipedia.org/wiki/IEEE_754#Rounding_rules

In the previous development, I found out the SSE rounding mode would be `Round to nearest, ties to even`.
However, I guess this might be the default rounding mode based on the description
```
_mm_cvtsd_si64:
Converts the lower DP FP value of a to a 64-bit signed integer value according to the current rounding mode.
```

NEON offers different rounding mode intrinsics, such as `vcvt`, `vcvtn` , `vcvtm` , `vcvtp` and `vcvta`.
Can we use the proper intrinsic to make the test case simpler?

HowJMay: I wonder even the [C implementation](https://github.com/DLTcollab/sse2neon/pull/327/files#diff-4e3d6805bace5270202c37d5363eded99ce96bea91040c8123cdc16c4f858d36R5742) on ARM is using rounding to zero

jserv: @JishinMaster, can you comment this?

HowJMay: I got some idea here 
https://github.com/DLTcollab/sse2neon/blob/master/sse2neon.h#L5495

JishinMaster:
I also had trouble with the fact that X86 and ARM do not have the same default rounding method.
With the test did you notice major differences?
One way could be to first get the current rounding mode, and choose the right intrinsic accordingly, although we might loose some throughput.

tests/impl.cpp
marktwtn: I believe that the [C implementation](https://github.com/DLTcollab/sse2neon/pull/327/files#diff-4e3d6805bace5270202c37d5363eded99ce96bea91040c8123cdc16c4f858d36R5742) is wrong.
Anyway, the `_mm_cvtsd_si64()` would be affect by `_MM_SET_ROUNDING_MODE()`.
The test should include it to test 4 different rounding modes.

marktwtn:
Close with #420.

PR #333, opened 2021-02-27
feat: Implement _mm_bs[l|r]li_si128

HowJMay:


jserv (changes requested):
Change the subject to "Implement _mm_bs[lr]li_si128"

PR #334, opened 2021-03-02
feat: Implement _mm_max_sd

HowJMay:


sse2neon.h
```diff
@@ -4240,6 +4240,24 @@ FORCE_INLINE __m128d _mm_max_pd(__m128d a, __m128d b)
...
+// element from a to the upper element of dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_max_sd
+FORCE_INLINE __m128d _mm_max_sd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    return _mm_move_sd(a, _mm_max_pd(a, b));
+#else
+    double *da = (double *) &a;
+    double *db = (double *) &b;
+    double c[2];
```
jserv, line 4254: Rewrite with inlined form:
```c
double c[2] = {fmax(da[0], db[0]), da[1]};
```

HowJMay: Do we need to apply this inline form to other `*_sd` intrinsics?

jserv: > Do we need to apply this inline form to other `*_sd` intrinsics?

Such trivial changes can be applied later.

sse2neon.h
```diff
@@ -4240,6 +4240,22 @@ FORCE_INLINE __m128d _mm_max_pd(__m128d a, __m128d b)
...
 }
 
+// Compare the lower double-precision (64-bit) floating-point elements in a and
+// b, store the maximum value in the lower element of dst, and copy the upper
+// element from a to the upper element of dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_max_sd
+FORCE_INLINE __m128d _mm_max_sd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    return _mm_move_sd(a, _mm_max_pd(a, b));
```
marktwtn, line 4250: I think we can use this for `ARM32` and `ARM64` since `_mm_max_pd` has implemented for different architecture.
However, the `_mm_max_pd` itself might need some improvement in the future pull request like using `fmax` for simplicity.

marktwtn: I guess the way @HowJMay did would have a better performance and the way I suggested would have better code readability.
@jserv what do you think?
Shall we follow the `KISS` principle?

jserv: Is it feasible to unify with single statement `return _mm_move_sd(a, _mm_max_pd(a, b))` without the `fmax` implementation?

marktwtn: Yes, it is.

HowJMay: `_mm_max_pd()` is conducting C implementation on ARM32, so calling it would be slower than the current C implementation on ARM32

marktwtn: What we need to do is to decide which one we would like to choose.
Performance or code readability.

HowJMay: I personally prefer to choose performance, since I think this library is aiming to achieve the very edge of the limit of performance

marktwtn (approved):

PR #340, opened 2021-03-06
test: Add test for _mm_extract_epi64

HowJMay:


tests/impl.cpp
```diff
@@ -6968,7 +6968,14 @@ result_t test_mm_extract_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
...
 {
-    return TEST_UNIMPL;
+    int64_t *_a = (int64_t *) impl.mTestIntPointer1;
+    const int imm = 1;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    __int64_t c = _mm_extract_epi64(a, imm);
+
+    ASSERT_RETURN(c == *(_a + (imm & 1)));
+    return TEST_SUCCESS;
```
marktwtn, line 6978: Could you rewrite the test case using the switch case?
https://github.com/DLTcollab/sse2neon/blob/f3f567484d099f74aaa51880db46b1434c960638/tests/impl.cpp#L4803

tests/impl.cpp
```diff
@@ -6968,7 +6968,22 @@ result_t test_mm_extract_epi32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_extract_epi64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    int64_t *_a = (int64_t *) impl.mTestIntPointer1;
+
+    __m128i a = do_mm_load_ps((const int32_t *) _a);
+    __int64_t c;
```
jserv, line 6974: You can even rewrite as following:
```c
__int64_t c = _mm_extract_epi64(a, i & 0x1);
```

HowJMay: The provided implementation can't be compiled. The following error would occur

```
error: selector must be an integer constant in the range [0, 1]
```

jserv: My bad. I forgot the qualifier.

PR #344, opened 2021-03-07
feat: Implement _mm_storeu_si<16-64>

marktwtn:
GCC version before 11 does not implement intrinsic functions
_mm_storeu_si16 and _mm_storeu_si32.
The testing should be modified after the GCC 11 releases.
Check https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95483 for the
bug information.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -955,11 +955,32 @@ FORCE_INLINE void _mm_store_si128(__m128i *p, __m128i a)
     vst1q_s32((int32_t *) p, vreinterpretq_s32_m128i(a));
 }
 
-// Stores four 32-bit integer values as (as a __m128i value) at the address p.
-// https://msdn.microsoft.com/en-us/library/vstudio/edk11s13(v=vs.100).aspx
+// Stores 128-bits of integer data a at the address p.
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -25,6 +25,7 @@
 //   Yang-Hao Yuan <yanghau@biilabs.io>
 //   Syoyo Fujita <syoyo@lighttransport.com>
 //   Brecht Van Lommel <brecht@blender.org>
+//   Remi Arnaud <remi@acm.org>
```
jserv, line 28: Use `Co-authored-by` information in git commit messages instead. See https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

tests/impl.cpp
```diff
@@ -2568,12 +2568,33 @@ result_t test_mm_storeu_ps(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_storeu_si16(const SSE2NEONTestImpl &impl, uint32_t i)
 {
+    // FIXME:
+    // The GCC version before 11 does not implement intrinsic function
```
jserv, line 2572: Why didn't you specify the conditional build with something like `if __GNUC__ <= 10`?

sse2neon.h

PR #346, opened 2021-03-08
test: Remove useless tests

HowJMay:
Function 'mm_shufflehi_epi16_function' and
'mm_shufflelo_epi16_function' are tool functions for
`_mm_shufflehi_epi16` and `_mm_shufflelo_epi16`, respectively.

So we can removed them from test cases.


tests/impl.h
```diff
@@ -7,532 +7,530 @@
     case it_##c:                  \
         ret = test_##c(*this, i); \
         break;
-#define INTRIN_FOREACH(TYPE)          \
```
jserv, line 10: Why did you change many lines instead of two?

HowJMay: It is caused by the formatter

marktwtn (approved):

PR #354, opened 2021-03-12
feat: Implement _mm_round_ss

HowJMay:


tests/impl.cpp
```diff
@@ -7587,7 +7587,17 @@ result_t test_mm_round_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    const float *_b = impl.mTestFloatPointer1;
+    float dx = roundf(_b[0]);
+    float dy = _a[1];
+    float dz = _a[2];
+    float dw = _a[3];
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 b = do_mm_load_ps(_b);
+    __m128 c = _mm_round_ss(a, b, _MM_FROUND_CUR_DIRECTION);
+    return validateFloatError(c, dx, dy, dz, dw, 0.001f);
```
marktwtn, line 7600: The error is not necessary.
We can simply use `validateFloat` function.

sse2neon.h
```diff
@@ -6070,6 +6070,69 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
...
+                             floorf(_b[3]));
+        pos_inf =
+            _mm_set_ps(ceilf(_b[0]), ceilf(_b[1]), ceilf(_b[2]), ceilf(_b[3]));
+        return _mm_move_ss(
+            vreinterpretq_f32_m128(a),
+            _mm_blendv_ps(pos_inf, neg_inf, _mm_cmple_ps(b, zero)));
+    default:  //_MM_FROUND_CUR_DIRECTION
+        return (__m128){roundf(_b[0]), _a[1], _a[2], _a[3]};
+    }
+#endif
```
marktwtn, line 6133: Can we replace with `_mm_move_ss(a, _mm_round_ps(b, rounding))`?

HowJMay: I think we can, but I have modified a little in the current implementation, so compare to ` _mm_move_ss(a, _mm_round_ps(b, rounding))`, we actually take fewer operations in ARM32 architecture.

For example, in `case (_MM_FROUND_TO_NEG_INF | _MM_FROUND_NO_EXC):
        return (__m128){floorf(_b[0]), _a[1], _a[2], _a[3]};`, we need to call `floorf` for only one time.

HowJMay:
@marktwtn is this OK for merging?

tests/impl.cpp
HowJMay: If I use `validateFloat()` we can't even pass `make check`

marktwtn: However, I think rounding should not contain any error.

HowJMay: I saw the return value of `_mm_round_ps()` contains error, too.

https://github.com/DLTcollab/sse2neon/blob/master/tests/impl.cpp#L7723

HowJMay: Should I open another PR to solve them together?

marktwtn: You can fix this pull request first and then open another pull request for `_mm_round_ps()`.

HowJMay: Something weird I found is `0.5` is round to `0` instead of `1`, but other numbers round successfully, e.g. `0.6` to `1.0`

marktwtn: `0.5` round to `0` might happen since floating-point rounding has 5 different mode.
Please check https://en.wikipedia.org/wiki/Floating-point_arithmetic#Rounding_modes for reference.

HowJMay: done

jserv:
> @marktwtn is this OK for merging?

Can you resolve the pending regressions?

marktwtn:
> > @marktwtn is this OK for merging?
> 
> Can you resolve the pending regressions?

I will try to investigate the rounding behaviour of different CPU architecture.

tests/impl.cpp
```diff
@@ -7587,7 +7587,18 @@ result_t test_mm_round_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    const float *_a = impl.mTestFloatPointer1;
+    const float *_b = impl.mTestFloatPointer1;
+    float d0 = bankersRounding(_b[0]);
+    float d1 = _a[1];
+    float d2 = _a[2];
+    float d3 = _a[3];
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 b = do_mm_load_ps(_b);
+    __m128 c = _mm_round_ss(a, b, _MM_FROUND_CUR_DIRECTION);
```
marktwtn, line 7599: The `bankers rounding` is the same as the rounding mode `_MM_FROUND_TO_NEAREST_INT`.
`_MM_FROUND_CUR_DIRECTION` means it uses the rounding mode which is set by `_MM_SET_ROUNDING_MODE()` before.

The way to fix it is to add
```
_MM_SET_ROUNDING_MODE(_MM_ROUND_NEAREST);
```
or to replace it with
```
__m128 c = _mm_round_ss(a, b, _MM_ROUND_NEAREST);
```

However, the testing should cover the four different rounding mode mentioned in the [document](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_round_ss).

sse2neon.h
```diff
@@ -6070,6 +6070,69 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
...
+        zero = _mm_set_ps(0.0f, 0.0f, 0.0f, 0.0f);
+        neg_inf = _mm_set_ps(floorf(_b[0]), floorf(_b[1]), floorf(_b[2]),
+                             floorf(_b[3]));
+        pos_inf =
+            _mm_set_ps(ceilf(_b[0]), ceilf(_b[1]), ceilf(_b[2]), ceilf(_b[3]));
+        return _mm_move_ss(
+            vreinterpretq_f32_m128(a),
+            _mm_blendv_ps(pos_inf, neg_inf, _mm_cmple_ps(b, zero)));
+    default:  //_MM_FROUND_CUR_DIRECTION
+        return (__m128){roundf(_b[0]), _a[1], _a[2], _a[3]};
```
marktwtn, line 6131: The original implementation from `_mm_round_ps()` is wrong.
It would be affected by the `_MM_SET_ROUNDING_MODE()` and the corresponding implementation should be different.

sse2neon.h
```diff
@@ -6070,6 +6070,69 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
...
+    }
+#else
+    float *_a = (float *) &a;
+    float *_b = (float *) &b;
+    __m128 zero, neg_inf, pos_inf;
+
+    switch (rounding) {
+    case (_MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC):
+        return _mm_move_ss(vreinterpretq_f32_m128(a),
+                           _mm_cvtepi32_ps(_mm_cvtps_epi32(b)));
```
marktwtn, line 6116: The `_mm_cvtps_epi32` would be affected by the `_MM_SET_ROUNDING_MODE()`, too.

marktwtn (changes requested):

marktwtn:
@HowJMay you can reference the implementation and testing of `_mm_round_ps()` to revise the pull request.

tests/impl.cpp
```diff
@@ -8020,7 +8020,85 @@ result_t test_mm_round_sd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_round_ss(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const float *_a = impl.mTestFloatPointer1;
+    const float *_b = impl.mTestFloatPointer1;
```
marktwtn, line 8024: `impl.mTestFloatPointer2`.

tests/impl.cpp
```diff
@@ -8020,7 +8020,85 @@ result_t test_mm_round_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    __m128 ret;
+
+    __m128 a = do_mm_load_ps(_a);
+    __m128 b = do_mm_load_ps(_b);
+    switch (i & 0x7) {
+    case 0:
+        f[0] = bankersRounding(_b[0]);
+        f[1] = _a[1];
+        f[2] = _a[2];
+        f[3] = _a[3];
```
marktwtn, line 8035: Maybe we can move the code out of the switch case since they are the same in different scenario?

tests/impl.cpp
```diff
@@ -8020,7 +8020,85 @@ result_t test_mm_round_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+    case 2:
+        f[0] = ceilf(_b[0]);
+        f[1] = _a[1];
+        f[2] = _a[2];
+        f[3] = _a[3];
+
+        ret = _mm_round_ss(a, b, _MM_FROUND_TO_POS_INF | _MM_FROUND_NO_EXC);
+        break;
+    case 3:
+        f[0] = (int) _b[0];
```
marktwtn, line 8056: Replace with `f[0] = _b[0] > 0 ? floorf(_b[0]) : ceilf(_b[0]);` to avoid data lost.

tests/impl.cpp
```diff
@@ -8020,7 +8020,85 @@ result_t test_mm_round_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+        f[0] = ceilf(_b[0]);
+        f[1] = _a[1];
+        f[2] = _a[2];
+        f[3] = _a[3];
+
+        _MM_SET_ROUNDING_MODE(_MM_ROUND_UP);
+        ret = _mm_round_ss(a, b, _MM_FROUND_CUR_DIRECTION);
+        break;
+    case 7:
+        f[0] = (int) _b[0];
```
marktwtn, line 8091: Ditto.

marktwtn (changes requested):

marktwtn (approved):

PR #361, opened 2021-03-14
feat: Implement _mm_sqrt_pd

HowJMay:


tests/common.cpp
```diff
@@ -362,4 +362,23 @@ result_t validateDouble(__m128d a, double d0, double d1)
     return TEST_SUCCESS;
 }
 
+result_t validateDoubleError(__m128d a, double d0, double d1, double err)
```
jserv, line 365: Can you use C++ function overloading to extend?

tests/impl.cpp
```diff
@@ -5120,7 +5120,17 @@ result_t test_mm_slli_si128(const SSE2NEONTestImpl &impl, uint32_t i)
...
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+
+    float f0 = sqrt(_a[0]);
+    float f1 = sqrt(_a[1]);
+
+    __m128d a = do_mm_load_pd(_a);
+    __m128d c = _mm_sqrt_pd(a);
+
+    // Here, we ensure `_mm_sqrt_pd()`'s error is under 1e-7% compares to the C
```
jserv, line 5131: Use quoted `"_mm_sqrt_pd"`.

PR #365, opened 2021-03-15
refactor: Replace f's with UINT[64|16|8]_C macro

HowJMay:
All f's heximal number is hard to distinguish how many "f" it has, and
may cause barrier during debugging. Therefore, they are replaced with
macro in corresponding bit size, or type casting -1.


marktwtn:
It is hard to know what is `f's` from reading the git commit header.

jserv:
> It is hard to know what `f's` is from reading the git commit header.

In fact, it is meant to be "subject." 
See https://chris.beams.io/posts/git-commit/

marktwtn (approved):

PR #369, opened 2021-03-17
test: Fix inconsistent type validation in _mm_sqrt_pd test

HowJMay:


jserv (changes requested):
Revise the subject. You can say, "Fix inconsistent type validation in _mm_sqrt_pd test."

PR #373, opened 2021-03-20
test: Implement the test of _mm_cvtsi128_si64x

marktwtn:


marktwtn:
Close due to the implementation in https://github.com/DLTcollab/sse2neon/pull/327.

PR #381, opened 2021-03-31
feat: Implement _mm_cmpnge_sd

HowJMay:


tests/impl.cpp
```diff
@@ -3671,7 +3671,17 @@ result_t test_mm_cmpnge_pd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_cmpnge_sd(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    double *_a = (double *) impl.mTestFloatPointer1;
+    double *_b = (double *) impl.mTestFloatPointer2;
+
+    int64_t f0 = (_a[0] >= _b[0]) ? UINT64_C(0) : ~UINT64_C(0);
+    int64_t f1 = ((int64_t *) _a)[1];
```
marktwtn, line 3678: Replace `int64_t` with `double`.

marktwtn (changes requested):

tests/impl.cpp
HowJMay: But I want to make the declaration of `f0` and `f1` in consistent. So you want to amend to the following style?
```diff
- int64_t f0 = (_a[0] >= _b[0]) ? UINT64_C(0) : ~UINT64_C(0);
- int64_t f1 = ((int64_t *) _a)[1];
+ int64_t f0 = (_a[0] >= _b[0]) ? UINT64_C(0) : ~UINT64_C(0);
+ double f1 = a[1];
```

marktwtn: How about change the data type of `f0` to `double` as well?

HowJMay: but this would make casting `UINT64_C(0)` to double become really complex

marktwtn (approved):

PR #383, opened 2021-04-11
Optimize _mm_cmpgt_epi64 for ARMv7

aqrit:
The resultant sign bit from a signed-saturated subtraction is a `greater than` comparison predicate.


jserv (changes requested):
Please run `make indent` before submitting the patch. We are using `clang-format` version 6.

marktwtn (approved):
Amazing optimization.

PR #385, opened 2021-04-16
Mac M1 build update.

devnexen:
This suppports the crc32 word size various flavors instructions.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -7520,7 +7520,7 @@ FORCE_INLINE void _mm_free(void *addr)
 // https://msdn.microsoft.com/en-us/library/bb514036(v=vs.100)
 FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
 {
-#if defined(__aarch64__) && defined(__ARM_FEATURE_CRC32)
+#if defined(__aarch64__) && (defined(__ARM_FEATURE_CRC32) || defined(__APPLE__))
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -7520,7 +7520,7 @@ FORCE_INLINE void _mm_free(void *addr)
 // https://msdn.microsoft.com/en-us/library/bb514036(v=vs.100)
 FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
 {
-#if defined(__aarch64__) && defined(__ARM_FEATURE_CRC32)
+#if defined(__aarch64__) && (defined(__ARM_FEATURE_CRC32) || defined(__APPLE__))
```
jserv, line 7523: I have no idea why you would make this change. The command I executed on Apple M1 to retrieve predefined macros:
```shell
clang -x c /dev/null -dM -E | grep __ARM_FEATURE_CRC32
```

I got the message `#define __ARM_FEATURE_CRC32 1`
It seems that the above change is not necessary. Can you clarify?

devnexen: Sure but in the project it was used due to compile flags, it was not defined.

jserv: Can you show me some examples which rely on `__APPLE__` macro instead of `__ARM_FEATURE_CRC32`?

devnexen:
I think I ll just attempt to fix the project build instead.

PR #389, opened 2021-04-30
feat: Implement _mm_cmpord_pd

HowJMay:


sse2neon.h
```diff
@@ -5167,6 +5167,33 @@ FORCE_INLINE __m128d _mm_cmpnge_sd(__m128d a, __m128d b)
...
+{
+#if defined(__aarch64__)
+    // Note: NEON does not have ordered compare builtin
+    // Need to compare a eq a and b eq b to check for NaN
+    // Do AND of results to get final
+    uint64x2_t ceqaa =
+        vceqq_f64(vreinterpretq_f64_m128d(a), vreinterpretq_f64_m128d(a));
+    uint64x2_t ceqbb =
+        vceqq_f64(vreinterpretq_f64_m128d(b), vreinterpretq_f64_m128d(b));
+    return vreinterpretq_m128d_u64(vandq_u64(ceqaa, ceqbb));
```
jserv, line 5183: Can we use `vandq_u64` intrinsic for Armv7-A as well?

HowJMay: done

sse2neon.h
```diff
@@ -5163,6 +5163,29 @@ FORCE_INLINE __m128d _mm_cmpnge_sd(__m128d a, __m128d b)
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_cmpngt_pd
 #define _mm_cmpngt_pd(a, b) _mm_cmple_pd(a, b)
 
+// Compare packed double-precision (64-bit) floating-point elements in a and b
+// to see if neither is NaN, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpord_pd
+FORCE_INLINE __m128d _mm_cmpord_pd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    // Note: NEON does not have ordered compare builtin
```
jserv, line 5172: The comment was a bit misleading. Improve it to reflect the code.

HowJMay: How about `For quiet NaN, w NaNs are not equal to each other, which is NaN_1 != NaN_2`

jserv: It is fine to me. Append more grammatically correct descriptions.

jserv: I was not aware of the change against the above comments.

tests/impl.cpp
```diff
@@ -3789,7 +3789,20 @@ result_t test_mm_cmpnlt_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+    __m128d a = _mm_load_pd(_a);
+    __m128d b = _mm_load_pd(_b);
+
+    float result[2];
+
+    for (uint32_t i = 0; i < 2; i++) {
+        result[i] = compord(_a[i], _b[i]);
```
marktwtn, line 3800: The `compord` should have a `double` version instead of using the `float` version.

sse2neon.h
```diff
@@ -5192,6 +5192,39 @@ FORCE_INLINE __m128d _mm_cmpnge_sd(__m128d a, __m128d b)
...
+    // Need to compare a eq a and b eq b to check for NaN
+    // Do AND of results to get final
```
marktwtn, line 5203: Rewrite as
```
The NaN checking is implemented with equal comparsion,
then do the AND operation to get the final result.
```

marktwtn (changes requested):

sse2neon.h
HowJMay: my fault. I have amended it.

sse2neon.h
```diff
@@ -5192,6 +5192,38 @@ FORCE_INLINE __m128d _mm_cmpnge_sd(__m128d a, __m128d b)
...
 
+// Compare packed double-precision (64-bit) floating-point elements in a and b
+// to see if neither is NaN, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpord_pd
+FORCE_INLINE __m128d _mm_cmpord_pd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    // For comparing two quiet NaN, we would get the result that they are not
+    // equal. In orther words, NaN1 != NaN2
+    uint64x2_t ceqaa =
```
jserv, line 5203: The name schemes `ceqaa` and `ceqbb` are a bit confusing. Revise them.

sse2neon.h
```diff
@@ -5192,6 +5192,38 @@ FORCE_INLINE __m128d _mm_cmpnge_sd(__m128d a, __m128d b)
...
 #define _mm_cmpngt_sd _mm_cmple_sd
 
+// Compare packed double-precision (64-bit) floating-point elements in a and b
+// to see if neither is NaN, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpord_pd
+FORCE_INLINE __m128d _mm_cmpord_pd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    // For comparing two quiet NaN, we would get the result that they are not
+    // equal. In orther words, NaN1 != NaN2
```
marktwtn, line 5202: Rewrite as:
```
Compare two quiet NaN would get the result that they are not equal.
In other words, NaN != NaN.
```

Why using `NaN` instead of `NaN1` is because we are comparing the same value in this implementation.

marktwtn (changes requested):

tests/impl.cpp
```diff
@@ -324,6 +324,16 @@ float compord(float a, float b)
     return ret;
 }
 
+double compord(double a, double b)
+{
+    double ret;
+
+    bool isNANA = !(a == a);
+    bool isNANB = !(b == b);
```
marktwtn, line 332: Use `isNAN()` like `float compord(float, float)` does.
Or maybe we can just simplify it as
```c++
ret = (!isNAN(a) && !isNAN(b)) ? getNAN() : 0.0f;
```

marktwtn (changes requested):

tests/impl.cpp
HowJMay: new version has been added.

marktwtn (approved):

sse2neon.h
```diff
@@ -5192,6 +5192,38 @@ FORCE_INLINE __m128d _mm_cmpnge_sd(__m128d a, __m128d b)
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpngt_sd
 #define _mm_cmpngt_sd _mm_cmple_sd
 
+// Compare packed double-precision (64-bit) floating-point elements in a and b
+// to see if neither is NaN, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpord_pd
+FORCE_INLINE __m128d _mm_cmpord_pd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    // Compare two quiet NaN would get the result that they are not equal.
```
jserv, line 5201: should be "Comparing ..."

jserv, line 5201: Replace the comment to
> Excluding NaNs, any two floating point numbers can be compared.

HowJMay: done

PR #394, opened 2021-05-03
CI: Show the difference before and after clang-format

marktwtn:
The commit is for the developers who can not install the
clang-format-6.0.
They can view the correct coding style from the Github Actions CI.

### DIFF
--- a/.ci/check-format.sh
+++ b/.ci/check-format.sh
@@ -5,4 +5,9 @@ SOURCES=$(find $(git rev-parse --show-toplevel) | egrep "\.(cpp|h)\$" | egrep -v
 
 set -x
 
+for file in ${SOURCES};
+do
+    clang-format-6.0 ${file} > expected-format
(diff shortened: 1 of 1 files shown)


jserv:
You can follow the way how I check the indention for student homework: https://github.com/sysprog21/lab0-c/blob/master/scripts/pre-commit.hook#L53

.ci/check-format.sh
```diff
@@ -5,4 +5,9 @@ SOURCES=$(find $(git rev-parse --show-toplevel) | egrep "\.(cpp|h)\$" | egrep -v
 
 set -x
 
+for file in ${SOURCES};
+do
+    clang-format-6.0 ${file} > after-clang-format
```
jserv, line 10: Rename `after-clang-format` to `expected-format` or something similar.

marktwtn: I have replaced it with `expected-format`.

PR #397, opened 2021-05-03
test: Improve the test coverage of mm_extract* intrinsics

HowJMay:


tests/impl.cpp
```diff
@@ -1656,11 +1656,25 @@ result_t test_mm_extract_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
...
+            break;
+        case 2:
+            c = _mm_extract_pi16(a, 2);
+            break;
+        case 3:
+            c = _mm_extract_pi16(a, 3);
+            break;
+    }
+
+    ASSERT_RETURN((uint64_t) c == ((*_a >> ((i & 0x3) * 16)) & 0xFFFF));
```
marktwtn, line 1677: I suggest you to have a variable `idx` which equals to `i&03`, then you can replace `idx` in the code to make it more readable.

marktwtn (changes requested):

tests/impl.cpp
```diff
@@ -4071,11 +4085,37 @@ result_t test_mm_div_sd(const SSE2NEONTestImpl &impl, uint32_t i)
...
+            break;
+        case 6:
+            c = _mm_extract_epi16(a, 6);
+            break;
+        case 7:
+            c = _mm_extract_epi16(a, 7);
+            break;
+    }
+
+    ASSERT_RETURN(c == _a[i&0x7]);
```
marktwtn, line 4118: Ditto.
And you can keep the original code writing format like `ASSERT_RETURN(c == *(_a + idx));`.

marktwtn (changes requested):

marktwtn (changes requested):
Fix the coding style.

marktwtn (approved):

jserv (changes requested):
I expect the change against the subject, which can be
> Improve the test coverage of _mm_extract_* intrinsics

PR #398, opened 2021-05-05
feat: Implement _mm_stream_si64

HowJMay:


tests/impl.cpp
```diff
@@ -5649,7 +5649,16 @@ result_t test_mm_stream_si32(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_stream_si64(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int64_t a = (const int64_t) impl.mTestInts[i];
+#if defined(__x86_64__) || defined(__i386__)
+    long long int p[1];
```
marktwtn, line 5654: Why using `long long int` here?
I thought `int64_t` is better.

HowJMay: It seems the function argument on x86 is in `long long int*` as I present below, so I am using two different datatype here
```c
_mm_stream_si64(long long *__p, long long __a)
{
  __builtin_ia32_movnti64(__p, __a);
}
```

marktwtn: Where did you find this?
It seems like the document is still using the `__int64*`, which can be replaced with `int64_t*`.
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_stream_si64

HowJMay: I guess it depends on the compiler?
I was using `int64_t` on x86 on node1, but I failed to compile the source code

marktwtn: What is the error message?

HowJMay: the error message is as following 

```
tests/impl.cpp:5659:21: error: invalid conversion from ‘int64_t*’ {aka ‘long int*’} to ‘long long int*’ [-fpermissive]
 5659 |     _mm_stream_si64(p, a);
      |                     ^
      |                     |
      |                     int64_t* {aka long int*}
In file included from tests/common.h:7,
                 from tests/impl.h:3,
                 from tests/impl.cpp:1:
/usr/lib/gcc/x86_64-linux-gnu/9/include/emmintrin.h:1477:33: note:   initializing argument 1 of ‘void _mm_stream_si64(long long int*, long long int)’
 1477 | _mm_stream_si64 (long long int *__A, long long int __B)
      |                  ~~~~~~~~~~~~~~~^~~
make: *** [Makefile:48: tests/impl.o] Error 1
```

HowJMay: here is the unit test for the above case 
```c
result_t test_mm_stream_si64(const SSE2NEONTestImpl &impl, uint32_t i)
{
    const int64_t a = (const int64_t) impl.mTestInts[i];
    int64_t p[1];
    _mm_stream_si64(p, a);
    ASSERT_RETURN(p[0] == a);
    return TEST_SUCCESS;
}
```

marktwtn: I thought using `int64_t*` directly should be workable since `int64_t` and `long long int` are integers with 64-bit size.

But taking `stdint.h` of **GCC** as an example, it only says that `long int` is equal to `int64_t` when the architecture is 64-bit.
We should treat `int64_t(long int)` and `long long int` differently since they are different data type.
If the casting is added like the below code, the error would be eliminated.
```c++
_mm_stream_si64((long long int*)p, a);
```

The [__mm_stream_si64()](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_stream_si64) document uses `__int64` as the argument type, which is a data type used in **Microsoft Visual C++** and equals to `long long int`.
As I searched for `__mm_stream_si64()` definition in `emmintrin.h` header file, it turns out that the argument type is `long long int` in **GCC** and **Clang**.

I think we just need to add the type casting without extra `x86` detection.

@jserv do you have any suggestion?

jserv: It is a matter of [64-bit data models](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models): Microsoft Windows (x86-64) takes LLP64 while Most Unix and Unix-like systems take LP64.

marktwtn: I think we should do the casting to avoid architecture detection?

HowJMay: what kinds of casting does you mean?

marktwtn: This one:
```c
_mm_stream_si64((long long int*)p, a);
```

marktwtn (approved):

tests/impl.cpp
```diff
@@ -6081,7 +6081,17 @@ result_t test_mm_stream_si32(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_stream_si64(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    const int64_t a = (const int64_t) impl.mTestInts[iter];
+#if defined(__x86_64__) || defined(__i386__)
+    // Microsoft Windows (x86-64) takes LLP64
```
jserv, line 6086: The comment was misleading. Can you specify with macro `__LP64__`?

HowJMay: I have removed the comments, and add macro `__LLP64__` in the `_x86_64` case

tests/impl.cpp
```diff
@@ -6081,7 +6081,15 @@ result_t test_mm_stream_si32(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_stream_si64(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    const int64_t a = (const int64_t) impl.mTestInts[iter];
+#if defined(__x86_64__) || defined(__i386__) || defined(__LLP64__)
```
jserv, line 6085: `__LLP64__` is a matter of ABI. Can you narrow down? That is, x86_64 with Linux, Windows, macOS, etc.

HowJMay: I have added sub-cases for different OSes under `_86_64`

tests/impl.cpp
```diff
@@ -6081,7 +6081,17 @@ result_t test_mm_stream_si32(const SSE2NEONTestImpl &impl, uint32_t iter)
...
```
jserv, line 6085: `__LLP64__` should be applied to Microsoft Windows x86-64 only.

tests/impl.cpp
```diff
@@ -6081,7 +6081,17 @@ result_t test_mm_stream_si32(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+#if defined(_WIN32) || defined(__APPLE__) || defined(__linux__)
```
jserv, line 6086: It is not necessary to enumerate the operating systems in favor of ABI.

HowJMay: Sorry for answering late. I was taking the exam until this week.
I still don't understand the request.
Should I amend in the following way?

```c
#if defined(__x86_64__) || defined(__i386__)
#if defined(_WIN32)  || defined(__LLP64__)
    long long p[1];
#else
    int64_t p[1];
#endif
#else
    int64_t p[1];
#endif
```

jserv: Alternatively, consider the following:
```c
#if (defined(__x86_64__) || defined(__i386__)) && (defined(_WIN32) || defined(__LLP64__))
    long long p[1];
#else
    int64_t p[1];
```

HowJMay: I checked the test. The test can't successfully build on the machine which runs Linux on x86.

HowJMay: we should use `long long p[1]` on Linux on x86 machine

HowJMay: @jserv is the current fashion meets your reuqest?

tests/impl.cpp
```diff
@@ -6081,7 +6081,16 @@ result_t test_mm_stream_si32(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_stream_si64(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    const int64_t a = (const int64_t) impl.mTestInts[iter];
+#if defined(__x86_64__) || defined(__i386__) || \
+    (defined(_WIN32) && defined(__LLP64__))
```
jserv, line 6086: Why `_WIN32` **AND** `__LLP64__`?

HowJMay: Or should I remove the `(defined(_WIN32) && defined(__LLP64__))` case?

If I conduct the following way, then it won't meet the implementation on Linux on x86 machine

```c
#if (defined(__x86_64__) || defined(__i386__)) && (defined(_WIN32) || defined(__LLP64__))
    long long p[1];
#else
    int64_t p[1];
#endif
```

marktwtn: Any progress?

PR #401, opened 2021-05-05
feat: Implement _mm_hsub_pi32

HowJMay:


tests/impl.cpp
```diff
@@ -6619,7 +6619,17 @@ result_t test_mm_hsub_pi16(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_hsub_pi32(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const int32_t *_a = impl.mTestIntPointer1;
+    const int32_t *_b = impl.mTestIntPointer1;
```
marktwtn, line 6623: It should be `impl.mTestIntPointer2`.
After the modification, please remove `vrev64_s32()` since it is not necessary.

marktwtn (changes requested):

tests/impl.cpp
HowJMay: done

marktwtn (approved):

PR #403, opened 2021-05-07
Fix _mm_round_ps

marktwtn:
test: Fix testing of _mm_round_ps

The test covers more scenario of different rounding mode.
The validation does not have to consider inaccuracy.

fix: Fix the implementation of _mm_round_ps

Simplify the implmentation of rounding mode _MM_FROUND_TO_ZERO.
Fix the implementation of rounding mode _MM_FROUND_CUR_DIRECTION.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -6351,28 +6351,41 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
     }
 #else
     float *v_float = (float *) &a;
-    __m128 zero, neg_inf, pos_inf;
 
-    switch (rounding) {
-    case (_MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC):
-        return _mm_cvtepi32_ps(_mm_cvtps_epi32(a));
-    case (_MM_FROUND_TO_NEG_INF | _MM_FROUND_NO_EXC):
-        return (__m128){floorf(v_float[0]), floorf(v_float[1]),
-                        floorf(v_float[2]), floorf(v_float[3])};
-    case (_MM_FROUND_TO_POS_INF | _MM_FROUND_NO_EXC):
-        return (__m128){ceilf(v_float[0]), ceilf(v_float[1]), ceilf(v_float[2]),
-                        ceilf(v_float[3])};
-    case (_MM_FROUND_TO_ZERO | _MM_FROUND_NO_EXC):
-        zero = _mm_set_ps(0.0f, 0.0f, 0.0f, 0.0f);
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -6351,27 +6351,43 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
...
+               (rounding == _MM_FROUND_CUR_DIRECTION &&
+                _MM_GET_ROUNDING_MODE() == _MM_ROUND_DOWN)) {
+        return _mm_set_ps(floorf(v_float[3]), floorf(v_float[2]),
+                          floorf(v_float[1]), floorf(v_float[0]));
+    } else if (rounding == (_MM_FROUND_TO_POS_INF | _MM_FROUND_NO_EXC) ||
+               (rounding == _MM_FROUND_CUR_DIRECTION &&
+                _MM_GET_ROUNDING_MODE() == _MM_ROUND_UP)) {
+        return _mm_set_ps(ceilf(v_float[3]), ceilf(v_float[2]),
+                          ceilf(v_float[1]), ceilf(v_float[0]));
+    } else {  // rounding == (_MM_FROUND_TO_ZERO | _MM_FROUND_NO_EXC) ||
```
jserv, line 6386: Replace `} else {` with plain `}` since it is the default case.

sse2neon.h
```diff
@@ -6351,27 +6351,43 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
...
+            vreinterpretq_f32_m128(a),
+            vcvtq_f32_s32(r_trunc)); /* compute delta: delta = (a - [a]) */
+        uint32x4_t is_delta_half =
+            vceqq_f32(delta, half); /* delta == +/- 0.5 */
+        return vreinterpretq_m128_f32(
+            vcvtq_f32_s32(vbslq_s32(is_delta_half, r_even, r_normal)));
+    } else if (rounding == (_MM_FROUND_TO_NEG_INF | _MM_FROUND_NO_EXC) ||
+               (rounding == _MM_FROUND_CUR_DIRECTION &&
+                _MM_GET_ROUNDING_MODE() == _MM_ROUND_DOWN)) {
+        return _mm_set_ps(floorf(v_float[3]), floorf(v_float[2]),
```
jserv, line 6379: Is it possible to eliminate the use of `floorf` and `ceilf`?

marktwtn: I am afraid not.
The available rounding SIMD are mainly focusing on `aarch64` instead of `arm`.
I do not find any suitable intrinsic for replacing `floorf` and `ceilf`.

sse2neon.h

PR #406, opened 2021-05-09
feat: Implement _mm_[round|floor|ceil]_pd

marktwtn:


tests/impl.cpp
```diff
@@ -194,6 +195,37 @@ static inline float bankersRounding(float val)
...
+    if (val < 0)
+        return -bankersRounding(-val);
+
+    double ret;
+    double roundDown = floor(val);  // Round down value
+    double roundUp = ceil(val);     // Round up value
+    double diffDown = val - roundDown;
+    double diffUp = roundUp - val;
+
+    if (diffDown <
```
jserv, line 209: Rewrite as following
```c
// If it's closer to the round down value, then use it
if (diffDown < diffUp) {
```

sse2neon.h
```diff
@@ -6417,6 +6419,73 @@ FORCE_INLINE __m128 _mm_round_ps(__m128 a, int rounding)
...
+            if (diffDown < diffUp) {  // If it's closer to the round down value,
+                                      // then use it
+                res[i] = roundDown;
+            } else if (diffDown > diffUp) {  // If it's closer to the round up
+                                             // value, then use it
+                res[i] = roundUp;
+            } else {
+                // If it's equidistant between round up and round down value,
+                // pick the one which is an even number
+                if (*((int64_t *) &roundDown) &
```
jserv, line 6463: Ditto. Reorder the comments.

tests/impl.cpp
marktwtn: I wonder where should I put the comment in this case:
```c
    } else if (diffDown >
               diffUp) {  // If it's closer to the round up value, then use it
        ret = roundUp;
```

jserv: Let's make the condition expression as clear as possible.

sse2neon.h

PR #409, opened 2021-05-09
fix: Avoid -Wc99-designator warning

encounter:
`_mm_aesenclast_si128` emits a warning with clang++.

```
sse2neon.h:6053:9: error: array designators are a C99 extension [-Werror,-Wc99-designator]
        [0] = {SSE2NEON_sbox[vreinterpretq_nth_u8_m128i(a, 0)],
        ^~~
1 error generated.
```

The array designators are not necessary at all, so this simply removes them.


jserv:
Thank @encounter for contributing!

PR #410, opened 2021-05-10
fix: Fix Bankers rounding algorithm

marktwtn:
Avoid data lost caused by conversion from floating point to integer.
Use the correct way to detect whether the floating point which has the
exact integer value is odd or even.


jserv (changes requested):
Always use "floating point" (without hyphen)

PR #412, opened 2021-05-13
feat: Add different rounding mode of _mm_cvt_[ps2pi|ss2si]

marktwtn:
Extend the implementation to handle the different rounding mode.


jserv (changes requested):
The subject "Fix" was bit misleading. In fact, this change appropriately takes different rounding modes into considerations.

PR #413, opened 2021-05-13
fix: Avoid data truncation of _mm_cvttss_si64

marktwtn:
The implementation should not convert the 32-bit floating point to
32-bit integer.
It would cause the data truncation if the 32-bit floating point contains the
value which is out of the range of 32-bit integer.


jserv (changes requested):
The change was intended to prevent data truncation.

PR #416, opened 2021-05-17
Two lint fixes

atdt:


jserv:
Thank @atdt for contributing!

PR #425, opened 2021-05-25
refactor: Replace conditional op with binary operation

HowJMay:
Some conditional operations can be replaced with binary operations
which should be able to accelerate the functions by removing branch
prediction.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1244,7 +1244,7 @@ FORCE_INLINE int _mm_comieq_ss(__m128 a, __m128 b)
     uint32x4_t a_and_b_not_nan = vandq_u32(a_not_nan, b_not_nan);
     uint32x4_t a_eq_b =
         vceqq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b));
-    return (vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) != 0) ? 1 : 0;
+    return vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) & 0x1;
 }
 
 // Compares the lower single-precision floating point scalar values of a and b
@@ -1261,7 +1261,7 @@ FORCE_INLINE int _mm_comige_ss(__m128 a, __m128 b)
     uint32x4_t a_and_b_not_nan = vandq_u32(a_not_nan, b_not_nan);
     uint32x4_t a_ge_b =
         vcgeq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b));
-    return (vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_ge_b), 0) != 0) ? 1 : 0;
+    return vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_ge_b), 0) & 0x1;
 }
 
 // Compares the lower single-precision floating point scalar values of a and b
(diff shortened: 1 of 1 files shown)


marktwtn (changes requested):
Revise the commit message:
`accelerate the functions with better branch prediction.`
-> `accelerate the functions by removing branch prediction.`

I think `!!` is hard to read.
Use `& 0x1` may be a good way to implement.

HowJMay:
> Revise the commit message:
> `accelerate the functions with better branch prediction.`
> -> `accelerate the functions by removing branch prediction.`
> 
> I think `!!` is hard to read.
> Use `& 0x1` may be a good way to implement.

you mean change to this way ?
`return !vgetq_lane_u32(vorrq_u32(a_or_b_nan, a_neq_b), 0) & 0x1;`

marktwtn:
> > Revise the commit message:
> > `accelerate the functions with better branch prediction.`
> > -> `accelerate the functions by removing branch prediction.`
> > I think `!!` is hard to read.
> > Use `& 0x1` may be a good way to implement.
> 
> you mean change to this way ?
> `return !vgetq_lane_u32(vorrq_u32(a_or_b_nan, a_neq_b), 0) & 0x1;`
> 
> If so, I prefer `!!` tho

For example:
`!!vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0)` would be
`vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) & 0x1`.

HowJMay:
> > > Revise the commit message:
> > > `accelerate the functions with better branch prediction.`
> > > -> `accelerate the functions by removing branch prediction.`
> > > I think `!!` is hard to read.
> > > Use `& 0x1` may be a good way to implement.
> > 
> > 
> > you mean change to this way ?
> > `return !vgetq_lane_u32(vorrq_u32(a_or_b_nan, a_neq_b), 0) & 0x1;`
> > If so, I prefer `!!` tho
> 
> For example:
> `!!vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0)` would be
> `vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) & 0x1`.

Maybe I miscalculate something here, but I think `!!vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0)` is not equal to `vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) & 0x1`.

I think we are not able to ensure a non-zero number with 1 in the least significant bit

marktwtn:
> Maybe I miscalculate something here, but I think `!!vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0)` is not equal to `vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) & 0x1`.
> 
> I think we are not able to ensure a non-zero number with 1 in the least significant bit

However, the intrinsics before the return value are mostly composed with comparison intrinsics.
Therefore, the non-zero value would be `0xFFFFFFFF`.

HowJMay:
> > Maybe I miscalculate something here, but I think `!!vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0)` is not equal to `vgetq_lane_u32(vandq_u32(a_and_b_not_nan, a_eq_b), 0) & 0x1`.
> > I think we are not able to ensure a non-zero number with 1 in the least significant bit
> 
> However, the intrinsics before the return value are mostly composed with comparison intrinsics.
> Therefore, the non-zero value would be `0xFFFFFFFF`.

Got it. It is my fault. But I still prefer `!!`. Could we let @jserv decide which way we should choose?

marktwtn:
> Got it. It is my fault. But I still prefer `!!`. Could we let @jserv decide which way we should choose?

There is another opinion I would like to share that I think `!!` would have two logic operationw and `& 0x1` would only have one logic operation.

HowJMay:
> > Got it. It is my fault. But I still prefer `!!`. Could we let @jserv decide which way we should choose?
> 
> There is another opinion I would like to share that I think `!!` would have two logic operationw and `& 0x1` would only have one logic operation.

That is a good point. I will change it

marktwtn (approved):

PR #426, opened 2021-05-25
refactor: Replace C operations with SIMD

HowJMay:
Replace the implementation in C to NEON implementation in which way
all the operation can be handled in register, and it should accelerate
the function.


HowJMay:
Close this, since it actually takes more instruction than the original implementation
see the test https://godbolt.org/z/4nTYjzxsT

HowJMay:
Refer to the comment [here](https://github.com/DLTcollab/sse2neon/pull/440#issuecomment-851397747), I conducted level 2 optimization, then the modified one is better than the original one

original: https://godbolt.org/z/98d9q4Pjj
modified: https://godbolt.org/z/dc8PEYvor

PR #427, opened 2021-05-26
feat: Implement _mm_comineq_sd

HowJMay:


sse2neon.h
```diff
@@ -3439,6 +3439,27 @@ FORCE_INLINE int _mm_comieq_sd(__m128d a, __m128d b)
...
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_comineq_sd
+FORCE_INLINE int _mm_comineq_sd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    return !vgetq_lane_u64(vceqq_f64(a, b), 0);
+#else
+    uint32x4_t a_not_nan =
+        vceqq_u32(vreinterpretq_u32_m128d(a), vreinterpretq_u32_m128d(a));
+    uint32x4_t b_not_nan =
+        vceqq_u32(vreinterpretq_u32_m128d(b), vreinterpretq_u32_m128d(b));
```
marktwtn, line 3453: As you interpret to `u32`, the **NaN** comparison would be meaningless.

tests/impl.cpp
```diff
@@ -3901,7 +3901,16 @@ result_t test_mm_comilt_sd(const SSE2NEONTestImpl &impl, uint32_t i)
 
 result_t test_mm_comineq_sd(const SSE2NEONTestImpl &impl, uint32_t i)
 {
-    return TEST_UNIMPL;
+    const double *_a = (const double *) impl.mTestFloatPointer1;
+    const double *_b = (const double *) impl.mTestFloatPointer2;
+    int32_t _c = (_a[0] != _b[0]) ? 1 : 0;
```
marktwtn, line 3906: Shall we add the checking for `NaN` value?

sse2neon.h
HowJMay: Should I remove the NaN detection here?
We may leave a `FIXME` here, since I think there is no way for detect double precision NaN.
`_mm_cmpeq_pd` has the same problem too

marktwtn: Please leave a `FIXME` here.

tests/impl.cpp
HowJMay: I won't add checking NaN, since I won't handle the `NaN` condition here.

marktwtn (approved):

PR #430, opened 2021-05-27
feat: Implement _mm_undefined_pd

HowJMay:


jserv:
Did you implement the corresponding test case?

tests/impl.cpp
```diff
@@ -6280,7 +6280,9 @@ result_t test_mm_ucomineq_sd(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_undefined_pd(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    __m128d a = _mm_undefined_pd();
+    (void) a;
```
marktwtn, line 6284: What is this for?

HowJMay: To remove the unused variable warning

marktwtn: I guess we might just return `TEST_SUCESS`
or replace the code with `_mm_undefined_pd()` without declaring variable.

@jserv any suggestion?

jserv: We can rewrite as following:
```c
__m128d a = _mm_undefined_pd();
a = _mm_xor_pd(a, a);
return validateDouble(a, 0, 0);
```

marktwtn (approved):

PR #433, opened 2021-05-28
feat: Implement _mm_testnzc_si128

HowJMay:


jserv (changes requested):
Simply call `_mm_test_mix_ones_zeros` in the test item.

PR #434, opened 2021-05-28
feat: Implement compare not intrinsics

HowJMay:
_mm_cmpnle_sd, _mm_cmpnlt_pd, _mm_cmpnlt_sd are added in this commit


jserv (changes requested):
Since the new intrinsics were implemented as macros, the corresponding test items can be invoking the defined functions instead.

jserv (changes requested):
Fix the typo in the subject of git commit.

PR #437, opened 2021-05-30
feat: Implement _mm_cmpunord_[pd|sd]

HowJMay:


sse2neon.h
```diff
@@ -3463,6 +3463,58 @@ FORCE_INLINE __m128d _mm_cmpord_sd(__m128d a, __m128d b)
 #endif
 }
 
+FORCE_INLINE __m128d _mm_cmpunord_pd(__m128d a, __m128d b)
```
jserv, line 3466: Add the descriptions from Intel Intrinsics website.

sse2neon.h
```diff
@@ -3463,6 +3463,61 @@ FORCE_INLINE __m128d _mm_cmpord_sd(__m128d a, __m128d b)
 #endif
 }
 
+// Compare packed double-precision (64-bit) floating-point elements in a and b
+// to see if either is NaN, and store the results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpunord_pd
+FORCE_INLINE __m128d _mm_cmpunord_pd(__m128d a, __m128d b)
+{
+#if defined(__aarch64__)
+    // Excluding NaNs, any two floating point numbers can be compared.
```
marktwtn, line 3472: NaNs can be compared.
Though the compared result is opposite to ordinary comparison.
You should clarify the behaviour of comparing with NaN and comparing with non-NaN.

marktwtn (changes requested):

sse2neon.h
HowJMay: How about `Two NaNs are not equal with comparison operation`?

marktwtn: That will do.
Please change `with comparison operation` with `in comparison operation`.

marktwtn (approved):

PR #439, opened 2021-05-30
_mm_cvtepu8_epi16(): fix comment

rouault:
See https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtepu8_epi16 that precises that it operates on 8 uint8 values

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -6979,12 +6979,13 @@ FORCE_INLINE __m128i _mm_cvtepu32_epi64(__m128i a)
         vmovl_u32(vget_low_u32(vreinterpretq_u32_m128i(a))));
 }
 
-// Converts the four unsigned 8-bit integers in the lower 16 bits to four
-// unsigned 32-bit integers.
+// Zero extend packed unsigned 8-bit integers in a to packed 16-bit integers,
+// and store the results in dst.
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -6979,12 +6979,12 @@ FORCE_INLINE __m128i _mm_cvtepu32_epi64(__m128i a)
         vmovl_u32(vget_low_u32(vreinterpretq_u32_m128i(a))));
 }
 
-// Converts the four unsigned 8-bit integers in the lower 16 bits to four
-// unsigned 32-bit integers.
+// Converts the eight unsigned 8-bit integers in the lower 64 bits to eight
+// unsigned 16-bit integers.
```
marktwtn, line 6983: Could you replace the comment with the following text?
```c
// Zero extend packed unsigned 8-bit integers in a to packed 16-bit integers,
// and store the results in dst.
// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtepu8_epi16
```

In the future, we would unify the comment of each intrinsic with the corresponding description on [Intel Intrinsic Guide](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#).

marktwtn (changes requested):

sse2neon.h
rouault: done

marktwtn (approved):

PR #440, opened 2021-05-31
_mm_sad_epu8(): use only Neon intrinsics instead of scalar operations

rouault:


HowJMay (approved):
It dramatically reduce the number of instructions

original:
https://godbolt.org/z/5oYxoeYdq

modified:
https://godbolt.org/z/aK71fWaYz

marktwtn (approved):

rouault:
to be more realistic, you should build with -02. So original (https://godbolt.org/z/ojW16M3zE) vs modified (https://godbolt.org/z/Mcz4drhEh)

marktwtn:
@rouault thank you for your contribution.

HowJMay:
@rouault Thank you for your contribution and advice

PR #441, opened 2021-05-31
refactor: Unify the implementation of compare not intrinsics

HowJMay:


jserv:
Would you shrink the test items of `tests/impl.cpp` as well? We can avoid the duplications.

PR #442, opened 2021-06-04
feat: Implement _m_pshufw

HowJMay:


jserv (changes requested):
Use the existing test case of _mm_shuffle_pi16 instead of duplicating.

HowJMay:
> Use the existing test case of _mm_shuffle_pi16 instead of duplicating.

@jserv but `_mm_shuffle_pi16` is declared and defined below `_m_pshufw`. If we want to call the test of `_mm_shuffle_pi16`, then we need to forward declaring `_mm_shuffle_pi16`. Is it ok?

jserv:
> @jserv but `_mm_shuffle_pi16` is declared and defined below `_m_pshufw`. If we want to call the test of `_mm_shuffle_pi16`, then we need to forward declaring `_mm_shuffle_pi16`. Is it ok?

It makes sense to use the forward declaration.

jserv (changes requested):
You have to rebase the latest master branch because I have migrated the GNU toolchain to v10.2 for CI.

sse2neon.h
```diff
@@ -6567,6 +6572,33 @@ FORCE_INLINE __m128i _mm_shuffle_epi8(__m128i a, __m128i b)
 #endif
 }
 
+FORCE_INLINE __m64 _mm_shuffle_pi8(__m64 a, __m64 b)
```
jserv, line 6575: Don't mix unrelated changes!

HowJMay: sorry for adding that part accidentally

PR #443, opened 2021-06-04
refactor: Rename test iteration count

HowJMay:
The iterating number which is given to the all tests as function
argument is too common, so I may conflict with the iterator
variable "i" in each tests. We can simply rename to other name.


jserv:
Can you simply use the term `iter` or `iterator`?

jserv:
The term "iterating number" sounds weird. It might be better to use "iteration number" or "iteration count."

marktwtn (approved):

PR #445, opened 2021-06-04
Drop FMA intrinsic

jserv:
Danila Kutenin pointed out:
> Technically speaking, _mm_fmadd_ps is not an SSE extension, this was
> introduced with fma extension which took place even after AVX.

To clarify the purpose of SSE2NEON, this pach would drop the existing
FMA implementation.

Related: #82

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -359,8 +359,6 @@ FORCE_INLINE __m128d _mm_round_pd(__m128d, int);
 FORCE_INLINE __m128 _mm_round_ps(__m128, int);
 // SSE4.2
 FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t, uint8_t);
-// FMA
-FORCE_INLINE __m128 _mm_fmadd_ps(__m128, __m128, __m128);
 
 /* Backwards compatibility for compilers with lack of specific type support */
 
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -6025,7 +6023,13 @@ FORCE_INLINE __m128d _mm_addsub_pd(__m128d a, __m128d b)
...
 {
     __m128 mask = {-1.0f, 1.0f, -1.0f, 1.0f};
-    return _mm_fmadd_ps(b, mask, a);
+#if defined(__aarch64__)
+    return vreinterpretq_m128_f32(vfmaq_f32(vreinterpretq_f32_m128(a),
+                                            vreinterpretq_f32_m128(mask),
+                                            vreinterpretq_f32_m128(b)));
+#else
+    return _mm_add_ps(_mm_mul_ps(b, mask), a);
+#endif
```
marktwtn, line 6032: Remove the ARM 32-bit implementation since `vfmaq_f32` is also supported in A32.

marktwtn (changes requested):

sse2neon.h
jserv: `vfmaq_f32` is only available for VFPv4+. Thus, for Armv7-A targets, we have to take the following cases into consideration:
- VFPv3, which is implemented on Cortex-R4 and R5 processors and the Tegra 2 (Cortex-A9).
- VFPv4, which is implemented on the A15 and Cortex-A7, or later

Reference: https://embeddedartistry.com/blog/2017/10/11/demystifying-arm-floating-point-compiler-options/

marktwtn (approved):

PR #446, opened 2021-06-04
feat: Implement _mm_cvtsi64x_si128

HowJMay:


HowJMay:
@jserv we are covering more than 90% now!

PR #455, opened 2021-06-11
feat: Implement _mm_maddubs_pi16

HowJMay:


tests/impl.cpp
```diff
@@ -7137,7 +7137,27 @@ result_t test_mm_maddubs_epi16(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+    int16_t d3 = (int16_t)(_a[3] * _b[3]);
+    int16_t d4 = (int16_t)(_a[4] * _b[4]);
+    int16_t d5 = (int16_t)(_a[5] * _b[5]);
+    int16_t d6 = (int16_t)(_a[6] * _b[6]);
+    int16_t d7 = (int16_t)(_a[7] * _b[7]);
+
+    int16_t e0 = saturate_16(d0 + d1);
+    int16_t e1 = saturate_16(d2 + d3);
+    int16_t e2 = saturate_16(d4 + d5);
+    int16_t e3 = saturate_16(d6 + d7);
```
marktwtn, line 7154: `saturate_s16()` should have argument of `int32_t`.
`int16_t` + `int16_t` would produce `int16_t` only.

sse2neon.h
```diff
@@ -6656,6 +6656,34 @@ FORCE_INLINE __m128i _mm_maddubs_epi16(__m128i _a, __m128i _b)
...
 
+// Vertically multiply each unsigned 8-bit integer from a with the corresponding
+// signed 8-bit integer from b, producing intermediate signed 16-bit integers.
+// Horizontally add adjacent pairs of intermediate signed 16-bit integers, and
+// pack the saturated results in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_maddubs_pi16
+FORCE_INLINE __m64 _mm_maddubs_pi16(__m64 _a, __m64 _b)
+{
+    // This would be much simpler if x86 would choose to zero extend OR sign
+    // extend, not both. This could probably be optimized better.
```
marktwtn, line 6667: I am not sure about the meaning of the comment.
As the conversion is implemented with NEON, why mention x86 here?

tests/impl.cpp
HowJMay: The result of `d0 + d1` would surpass the range of `int16_t`, since argument `a` is treated as `uint8_t` and `b` is treated as `int8_t`. The maximum positive value of `int16_t` is `2^15 - 1`, and the maximum possible result would be `2*(2^8 - 1)*(2^7 - 1)` which would be more than `2^15 - 1`. Therefore, we need `saturate_16()`

marktwtn: I might misread in the previous review.

marktwtn (approved):

PR #457, opened 2021-06-12
test: Add test for _mm_undefined_ps

HowJMay:


tests/impl.cpp
```diff
@@ -2849,6 +2849,9 @@ result_t test_mm_ucomineq_ss(const SSE2NEONTestImpl &impl, uint32_t iter)
 result_t test_mm_undefined_ps(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
     return TEST_UNIMPL;
```
jserv, line 2851: Remove this line.

PR #458, opened 2021-06-27
feat: Implement and rewrite _mm_cvtps_pi[8|16|32]

marktwtn:
The modification includes:
- Write the correct tests of _mm_cvtps_pi[8|16|32]
- Fix the original implementation of _mm_cvtps_pi16
- Implement _mm_cvtps_pi8

The test suite would check different rounding mode.

_mm_cvtps_pi[8|16] should set different output value based on the range
of the input value.
Take _mm_cvtps_pi16() for example:
0x7FFF(INT16_MAX) <= input <= 0x7FFFFFFF(INT32_MAX)
  output = 0x7FFF(INT16_MAX)
0x8000(INT16_MIN) < input < 0x7FFF(INT16_MAX)
  output = conversion from floating point to integer with corresponding roundinging mode
other input values
  output = 0x8000(INT16_MIN)

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -344,12 +344,21 @@ typedef union ALIGN_STRUCT(16) SIMDVec {
 // SSE
 FORCE_INLINE unsigned int _MM_GET_ROUNDING_MODE();
 FORCE_INLINE __m128 _mm_move_ss(__m128, __m128);
+FORCE_INLINE __m128 _mm_or_ps(__m128, __m128);
+FORCE_INLINE __m128 _mm_set_ps1(float);
+FORCE_INLINE __m128 _mm_setzero_ps(void);
 // SSE2
+FORCE_INLINE __m128i _mm_and_si128(__m128i, __m128i);
+FORCE_INLINE __m128i _mm_castps_si128(__m128);
+FORCE_INLINE __m128i _mm_cmpeq_epi32(__m128i, __m128i);
 FORCE_INLINE __m128i _mm_cvtps_epi32(__m128);
 FORCE_INLINE __m128d _mm_move_sd(__m128d, __m128d);
+FORCE_INLINE __m128i _mm_or_si128(__m128i, __m128i);
 FORCE_INLINE __m128i _mm_set_epi32(int, int, int, int);
 FORCE_INLINE __m128i _mm_set_epi64x(int64_t, int64_t);
 FORCE_INLINE __m128d _mm_set_pd(double, double);
+FORCE_INLINE __m128i _mm_set1_epi32(int);
+FORCE_INLINE __m128i _mm_setzero_si128();
 // SSE4.1
 FORCE_INLINE __m128d _mm_ceil_pd(__m128d);
 FORCE_INLINE __m128 _mm_ceil_ps(__m128);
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -1469,11 +1478,33 @@ FORCE_INLINE __m128 _mm_cvtpi8_ps(__m64 a)
...
+//       dst[i+15:i] := Convert_FP32_To_Int16(a[k+31:k])
+//     FI
+//   ENDFOR
+//
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtps_pi16
 FORCE_INLINE __m64 _mm_cvtps_pi16(__m128 a)
 {
-    return vreinterpret_m64_s16(
-        vmovn_s32(vreinterpretq_s32_m128i(_mm_cvtps_epi32(a))));
+    __m128 i16Min = _mm_set_ps1(INT16_MIN);
```
jserv, line 1494: I have no idea about the necessity of these changes.

sse2neon.h
```diff
@@ -1469,11 +1478,33 @@ FORCE_INLINE __m128 _mm_cvtpi8_ps(__m64 a)
...
+    __m128 i16Max = _mm_set_ps1(INT16_MAX);
+    __m128 i32Max = _mm_set_ps1(INT32_MAX);
+    __m128i maxMask = _mm_castps_si128(
+        _mm_and_ps(_mm_cmpge_ps(a, i16Max), _mm_cmple_ps(a, i32Max)));
+    __m128i betweenMask = _mm_castps_si128(
+        _mm_and_ps(_mm_cmpgt_ps(a, i16Min), _mm_cmplt_ps(a, i16Max)));
+    __m128i minMask = _mm_cmpeq_epi32(_mm_or_si128(maxMask, betweenMask),
+                                      _mm_setzero_si128());
+    __m128i max = _mm_and_si128(maxMask, _mm_set1_epi32(INT16_MAX));
```
jserv, line 1503: Since the above masks are not changed, you shall declare them as `const` as possible.

sse2neon.h
marktwtn: The original implementation does not take care of the description
```
Note: this intrinsic will generate 0x7FFF,
rather than 0x8000, for input values between 0x7FFF and 0x7FFFFFFF.
```

The intrinsic should set the corresponding value
```
0x7FFF <= input <= 0x7FFFFFFF
output = 0x7FFF(INT16_MAX)
0x8000 < input < 0x7FFF
output = _mm_cvtps_epi32(input)
other situation
output = 0x8000(INT16_MIN)
```

The original implementation calls the conversion intrinsic `_mm_cvtps_epi32()` only.

sse2neon.h
marktwtn: Done.

sse2neon.h
jserv: Mention the above in git commit message.

sse2neon.h
```diff
@@ -1469,11 +1478,33 @@ FORCE_INLINE __m128 _mm_cvtpi8_ps(__m64 a)
...
+//     FI
+//   ENDFOR
+//
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtps_pi16
 FORCE_INLINE __m64 _mm_cvtps_pi16(__m128 a)
 {
-    return vreinterpret_m64_s16(
-        vmovn_s32(vreinterpretq_s32_m128i(_mm_cvtps_epi32(a))));
+    __m128 i16Min = _mm_set_ps1(INT16_MIN);
+    __m128 i16Max = _mm_set_ps1(INT16_MAX);
```
jserv, line 1495: Both `i16Max` and `i32Max` are constant, right?

sse2neon.h
```diff
@@ -1487,6 +1518,45 @@ FORCE_INLINE __m64 _mm_cvtps_pi16(__m128 a)
...
+//       dst[i+7:i] := 0x7F
+//     ELSE
+//       dst[i+7:i] := Convert_FP32_To_Int8(a[k+31:k])
+//     FI
+//   ENDFOR
+//
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtps_pi8
+FORCE_INLINE __m64 _mm_cvtps_pi8(__m128 a)
+{
+    __m128 i8Min = _mm_set_ps1(INT8_MIN);
```
jserv, line 1539: Ditto. Add `const` qualifier.

sse2neon.h
marktwtn: Yes, I'll add the `const` keyword.

jserv (changes requested):
The subject was a bit misleading. It was intended to consolidate _mm_cvtps_pi16 rather than implementing something new.

marktwtn:
> The subject was a bit misleading. It was intended to consolidate _mm_cvtps_pi16 rather than implementing something new.

It contains three different parts
- Write the correct tests of `_mm_cvtps_pi[8|16|32]`
- Fix the implementation of `_mm_cvtps_pi16`
- Implement `_mm_cvtps_pi8`

I think I'll rewrite is as `feat: Implement and rewrite _mm_cvtps_pi[8|16|32]` and describe the detailed modification in the commit message.

sse2neon.h

sse2neon.h

sse2neon.h

jserv:
Message tweak from
> The correct tests would test with different rounding mode

to
> The test suite would check different rounding mode

PR #466, opened 2021-07-23
feat: Implement _mm_mpsadbw_epu8

HowJMay:


HowJMay:
The official website of intel intrinsics seems crashed. We can check the intrinsic behavior by these two website
[mirror of intel intrinsics](https://www.laruence.com/sse/#text=_mm_mpsadbw_epu8&expand=3897)
[doc of rustlang](https://doc.rust-lang.org/beta/core/arch/x86_64/fn._mm_mpsadbw_epu8.html)

sse2neon.h
```diff
@@ -7748,6 +7748,80 @@ FORCE_INLINE __m128i _mm_minpos_epu16(__m128i a)
...
+// dst. Eight SADs are performed using one quadruplet from b and eight
+// quadruplets from a. One quadruplet is selected from b starting at on the
+// offset specified in imm8. Eight quadruplets are formed from sequential 8-bit
+// integers selected from a starting at the offset specified in imm8.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_mpsadbw_epu8
+FORCE_INLINE __m128i _mm_mpsadbw_epu8(__m128i a, __m128i b, const int imm)
+{
+    uint8x16_t _a, _b;
+
+    switch (imm & 0x4) {
```
jserv, line 7762: Is it possible to have uninitialized `_a`? e.g., when `(imm & 0x4) == 3`

HowJMay: But I think the result of `imm&0x4` will be either 0 or 4, therefore, other values are impossible

jserv: Would you add some compiler hints such as `__builtin_unreachable` there?

sse2neon.h
```diff
@@ -7749,6 +7749,86 @@ FORCE_INLINE __m128i _mm_minpos_epu16(__m128i a)
...
+    case 0:
+        // do nothing
+        _a = vreinterpretq_u8_m128i(a);
+        break;
+    case 4:
+        _a = vreinterpretq_u8_u32(vextq_u32(vreinterpretq_u32_m128i(a),
+                                            vreinterpretq_u32_m128i(a), 1));
+        break;
+    default:
+        __builtin_unreachable();
```
jserv, line 7773: `__builtin_unreachable` is specific to gcc and clang. Use `#if defined(__GNUC__) || defined(__clang__)`.

sse2neon.h
HowJMay: I have added `__builtin_unreachable()` to the `default` two switch-case

sse2neon.h
HowJMay: added

PR #467, opened 2021-07-23
_mm_cvtps_pi8: Change uint to uint32_t

betajippity:
Using just "uint", sse2neon does not compile on macOS using Apple's
clang variant. Changing "uint" to "uint32_t" solves the problem.


marktwtn (approved):

marktwtn:
@betajippity thank you for fixing the problem.

PR #468, opened 2021-07-25
fix: Fix implicit conversion warning

HowJMay:
Fix the following warning which is shown on Apple M1
```
warning: implicit conversion from 'int' to 'int8_t' (aka 'signed char')
    changes value from 135 to -121 [-Wconstant-conversion]
        vand_s8(vreinterpret_s8_m64(b), vdup_n_s8((0x1 << 7 | 0x07)));
                                        ~~~~~~~~~  ~~~~~~~~~^~~~~~
```

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -1442,7 +1442,7 @@ result_t test_mm_cvtps_pi16(const SSE2NEONTestImpl &impl, uint32_t iter)
     int16_t rnd[4];
 
     for (int i = 0; i < 4; i++) {
-        if (INT16_MAX <= _a[i] && _a[i] <= INT32_MAX) {
+        if ((float) INT16_MAX <= _a[i] && _a[i] <= (float) INT32_MAX) {
             rnd[i] = INT16_MAX;
         } else if (INT16_MIN < _a[i] && _a[i] < INT16_MAX) {
             switch (iter & 0x3) {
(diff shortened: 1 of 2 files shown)


jserv:
Is the message identical when you build SSE2NEON with clang/linux?

HowJMay:
@jserv Based on the CI results which were run on ubuntu, the same error doesn't exist. However, there are more warnings that needs to handle. Should I open another PR for it or solve them in this PR?

jserv:
> @jserv Based on the CI results which were run on ubuntu, the same error doesn't exist. However, there are more warnings that need to be handled. Should I open another PR for it or solve them in this PR?

You shall eliminate the potential warnings raised by known compilers. Let's work on this pull request.

HowJMay:
@jserv all warnings are fixed.

tests/impl.cpp
```diff
@@ -1442,7 +1442,7 @@ result_t test_mm_cvtps_pi16(const SSE2NEONTestImpl &impl, uint32_t iter)
     int16_t rnd[4];
 
     for (int i = 0; i < 4; i++) {
-        if (INT16_MAX <= _a[i] && _a[i] <= INT32_MAX) {
+        if (INT16_MAX <= (int) _a[i] && (int) _a[i] <= INT32_MAX) {
```
marktwtn, line 1445: I prefer to cast the `INT16_MAX` to `float` since casting with `int` would have data loss.

tests/impl.cpp
```diff
@@ -1513,7 +1513,7 @@ result_t test_mm_cvtps_pi8(const SSE2NEONTestImpl &impl, uint32_t iter)
     int8_t rnd[4];
 
     for (int i = 0; i < 4; i++) {
-        if (INT8_MAX <= _a[i] && _a[i] <= INT32_MAX) {
+        if (INT8_MAX <= (int) _a[i] && (int) _a[i] <= INT32_MAX) {
```
marktwtn, line 1516: Ditto.

marktwtn (changes requested):

marktwtn (approved):

jserv:
Don't use Markdown-specific syntax in git commit messages. Instead, always use plain text. Append `Close #471` at the end of the git commit message.

Issue #465, opened 2021-07-23
Missing denormal mode macros


audioboy77:
The macros of 
_MM_SET_DENORMALS_ZERO_MODE
_MM_SET_FLUSH_ZERO_MODE
and associated flags aren't covered, so I cant use as a drop in replacement.  

Please consider to add this, thanks

marktwtn:
The **x86** [MXCSR register](https://help.totalview.io/previous_releases/2019/html/index.html#page/Reference_Guide/Intelx86MXSCRRegister.html) has two bits for `FZ`(flush to zero) and `DAZ`(denormals are zeros).
In **ARM**, the **AArch32** [FPSCR register](https://developer.arm.com/documentation/ddi0595/2021-06/AArch32-Registers/FPSCR--Floating-Point-Status-and-Control-Register?lang=en) and **AArch64** [FPCR register](https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Registers/FPCR--Floating-point-Control-Register) have bit for `FZ`(flush to zero) only.

However, I guess the behaviour of `_MM_SET_DENORMALS_ZERO_MODE` and `_MM_SET_FLUSH_ZERO_MODE` are pretty similar.
I will have more investigation about the floating point intrinsics and operations.
If they do have the same behaviour, the implementation for **ARM** would be the same.

PR #476, opened 2021-08-19
feat: Implement _MM_[GET|SET]_FLUSH_ZERO_MODE

marktwtn:
The behavior of denormal number and flush zero mode has not been
fully investigated yet.
The validation of __MM_SET_FLUSH_ZERO_MODE should be added in the
future.

Related to #465.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -158,6 +158,10 @@
 #define _MM_ROUND_DOWN 0x2000
 #define _MM_ROUND_UP 0x4000
 #define _MM_ROUND_TOWARD_ZERO 0x6000
+/* Flush zero mode macros. */
+#define _MM_FLUSH_ZERO_MASK 0x8000
+#define _MM_FLUSH_ZERO_ON 0x8000
+#define _MM_FLUSH_ZERO_OFF 0x0000
 
 /* indicate immediate constant argument in a given range */
 #define __constrange(a, b) const
@@ -332,6 +336,10 @@ typedef union ALIGN_STRUCT(16) SIMDVec {
 #define vreinterpretq_nth_u32_m128i(x, n) (((SIMDVec *) &x)->m128_u32[n])
 #define vreinterpretq_nth_u8_m128i(x, n) (((SIMDVec *) &x)->m128_u8[n])
 
+/* SSE macros */
+#define _MM_GET_FLUSH_ZERO_MODE _sse2neon_mm_get_flush_zero_mode
+#define _MM_SET_FLUSH_ZERO_MODE _sse2neon_mm_set_flush_zero_mode
+
 // Function declaration
 // SSE
 FORCE_INLINE unsigned int _MM_GET_ROUNDING_MODE();
@@ -480,7 +488,8 @@ typedef struct {
     uint8_t res1 : 6;
     uint8_t bit22 : 1;
     uint8_t bit23 : 1;
-    uint8_t res2;
+    uint8_t bit24 : 1;
+    uint8_t res2 : 7;
 #if defined(__aarch64__)
     uint32_t res3;
 #endif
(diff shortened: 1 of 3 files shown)


sse2neon.h
```diff
@@ -1744,6 +1749,34 @@ FORCE_INLINE void _mm_free(void *addr)
...
+#endif
+    } r;
+
+#if defined(__aarch64__)
+    asm volatile("mrs %0, FPCR" : "=r"(r.value)); /* read */
+#else
+    asm volatile("vmrs %0, FPSCR" : "=r"(r.value)); /* read */
+#endif
+
+    if (r.field.bit24) {
```
jserv, line 1773: Rewrite as following:
```c
return r.field.bit24 ? _MM_FLUSH_ZERO_ON : _MM_FLUSH_ZERO_OFF;
```

sse2neon.h
```diff
@@ -2320,6 +2353,40 @@ FORCE_INLINE __m64 _mm_sad_pu8(__m64 a, __m64 b)
...
+#endif
+    } r;
+
+#if defined(__aarch64__)
+    asm volatile("mrs %0, FPCR" : "=r"(r.value)); /* read */
+#else
+    asm volatile("vmrs %0, FPSCR" : "=r"(r.value)); /* read */
+#endif
+
+    if ((flag & _MM_FLUSH_ZERO_MASK) == _MM_FLUSH_ZERO_ON) {
```
jserv, line 2377: Rewrite with branchless. That is, no if-else statement.

tests/impl.cpp
```diff
@@ -1823,6 +1823,23 @@ result_t test_mm_free(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+                                     uint32_t iter)
+{
+    int res_flush_zero_on, res_flush_zero_off;
+    _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);
+    res_flush_zero_on = _MM_GET_FLUSH_ZERO_MODE() == _MM_FLUSH_ZERO_ON ? 1 : 0;
+    _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_OFF);
+    res_flush_zero_off =
+        _MM_GET_FLUSH_ZERO_MODE() == _MM_FLUSH_ZERO_OFF ? 1 : 0;
+
+    if (res_flush_zero_on && res_flush_zero_off) {
```
jserv, line 1836: Ditto. Remove if-else.

tests/impl.cpp
```diff
@@ -2455,6 +2472,12 @@ result_t test_mm_sad_pu8(const SSE2NEONTestImpl &impl, uint32_t iter)
     return validateUInt16(c, d, 0, 0, 0);
 }
 
+result_t test_mm_set_flush_zero_mode(const SSE2NEONTestImpl &impl,
+                                     uint32_t iter)
+{
+    return TEST_UNIMPL;
```
jserv, line 2478: How can we validate this macro?

sse2neon.h
marktwtn: The branch has been removed.

tests/impl.cpp
marktwtn: Done.

tests/impl.cpp
marktwtn: First, we need to create a denormal number, do the addition to the normal floating point numbers,
and check the result.

A denormal number is where the mantissa is non zero, but the exponent value is zero in an IEEE* floating-point representation.
It would be treated as zero when flush zero mode is on.

However, I am not really sure about the exact behaviour and need more investigation.
Hence, I left the validation to the future.

jserv: Put some considerations in comment(s). Then, add `FIXME` or `TODO`.

marktwtn: Done.

sse2neon.h
```diff
@@ -1744,6 +1749,30 @@ FORCE_INLINE void _mm_free(void *addr)
     free(addr);
 }
 
+// Macro: Get the flush zero bits from the MXCSR control and status register.
+// The flush zero may contain any of the following flags: _MM_FLUSH_ZERO_ON or
+// _MM_FLUSH_ZERO_OFF
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_MM_GET_FLUSH_ZERO_MODE
+FORCE_INLINE unsigned int _MM_GET_FLUSH_ZERO_MODE()
```
jserv, line 1756: Both `_MM_GET_FLUSH_ZERO_MODE` and `_MM_SET_FLUSH_ZERO_MODE` were referred to be macros in Intel Intrinsics Guide. Maybe we should follow the behavior. That is, we can implement static inline functions such as `_sse2neon_get_flush_zero_mode` and define the corresponding macros.

tests/impl.cpp
```diff
@@ -1823,6 +1823,22 @@ result_t test_mm_free(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+{
+    int res_flush_zero_on, res_flush_zero_off;
+    _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);
+    res_flush_zero_on = _MM_GET_FLUSH_ZERO_MODE() == _MM_FLUSH_ZERO_ON;
+    _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_OFF);
+    res_flush_zero_off = _MM_GET_FLUSH_ZERO_MODE() == _MM_FLUSH_ZERO_OFF;
+
+    if (res_flush_zero_on && res_flush_zero_off) {
+        return TEST_SUCCESS;
+    } else {
```
jserv, line 1837: Remove `else` statement since it is not necessary.

marktwtn: I'll rewrite it as `return (res_flush_zero_on && res_flush_zero_off) ? TEST_SUCCESS : TEST_FAIL;`

sse2neon.h

tests/impl.cpp

PR #478, opened 2021-08-27
test: Rewrite load rountine functions to templates

marktwtn:
The templates can avoid extra casting.
The validations are removed since the corresponding intrinsic testings
have been implemented.


jserv (changes requested):
The subject was misleading. According to https://www.techopedia.com/definition/25329/data-loading , Data loading is the process of copying and loading data or data sets from a source file, folder or application to a database or similar application. However, the proposed changes were dedicated to "load routine."

PR #479, opened 2021-09-03
feat: Implement _MM_[GET|SET]_DENORMALS_ZERO_MODE

### LINKED ISSUE #465: Missing denormal mode macros

marktwtn:
X86 has two modes, which are flush to zero mode and denormals are zeros
mode.
ARM has one mode only, which is flush to zero mode.
The ARM flush to zero mode is like the combination of the two modes of
X86.

However, AArch32 Advanced SIMD arithmetic always uses the flush to zero
setting, regardless of the value of the FZ(flush to zero) bit.

The implementation would change the FZ bit and the testing would focus
on whether the input denormal floating point is treated as zero or not.

Close #465.


tests/impl.h
```diff
@@ -7,531 +7,533 @@
     case it_##c:                  \
         ret = test_##c(*this, i); \
         break;
-#define INTRIN_FOREACH(TYPE)       \
```
jserv, line 10: Why did you change this a lot?

marktwtn: Because the new intrinsics `_mm_[get|set]_denormals_zero_mode` are added, and they are longer than other intrinsic names.
The changing is done by applying the coding style checking command.

sse2neon.h
```diff
@@ -162,6 +162,10 @@
 #define _MM_FLUSH_ZERO_MASK 0x8000
 #define _MM_FLUSH_ZERO_ON 0x8000
 #define _MM_FLUSH_ZERO_OFF 0x0000
+/* Denormals are zeros mode macros. */
```
jserv, line 165: The comment in GCC looks better. See https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/pmmintrin.h#L40

jserv: However, can we implement a generic [_mm_getcsr](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_getcsr) instead?

marktwtn: I am not sure whether GCC comment is better since Clang does not have similar comment and I do not realize why it is an additional bit in MXCSR.

marktwtn: _mm_getcsr would be more complicated.
The mapping of the X86 and ARM register should be fully investigated.
At least, it would not be implemented in this pull request.

PR #483, opened 2021-09-09
refactor: Rewrite _mm_alignr_epi8() to function

### LINKED ISSUE #482: Problem with _mm_alignr_epi8 and constants
I'm seeing the following error in my code:
```
GSVector4i.h:676:21: error: argument value -8 is outside the valid range [0, 15]
                return GSVector4i(_mm_alignr_epi8(v.m, m, i));
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
sse2neon.h:6552:44: note: expanded from macro '_mm_alignr_epi8'
                    vreinterpretq_m128i_u8(vextq_u8(tmp_low, tmp_high, idx)); \
                                           ^                           ~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/12.0.5/include/arm_neon.h:6812:24: note: expanded from macro 'vextq_u8'
  __ret = (uint8x16_t) __builtin_neon_vextq_v((int8x16_t)__s0, (int8x16_t)__s1, __p2, 48); \
                       ^                                                        ~~~~
sse2neon.h:222:56: note: expanded from macro 'vreinterpretq_m128i_u8'
#define vreinterpretq_m128i_u8(x) vreinterpretq_s64_u8(x)
                                                       ^
GSBlock.h:1280:12: note: in instantiation of function template specialization 'GSVector4i::srl<8>' requested here
                        v4 = v5.srl<8>(v6);
                                ^
6 errors generated.
```

The root is compiling the following code with a specialisation of 8
```
	template <int i>
	__forceinline GSVector4i srl(const GSVector4i& v)
	{
		return GSVector4i(_mm_alignr_epi8(v.m, m, i));
	}
```

That should be fine but it barfs on the last line of the snippet below:
```
#define _mm_alignr_epi8(a, b, imm)                                            \
    __extension__({                                                           \
        __m128i ret;                                                          \
        if (_sse2neon_unlikely((imm) >= 32)) {                                \
            ret = _mm_setzero_si128();                                        \
        } else {                                                              \
            uint8x16_t tmp_low, tmp_high;                                     \
            if (imm >= 16) {                                                  \
                const int idx = imm - 16;                                     \
                tmp_low = vreinterpretq_u8_m128i(a);                          \
                tmp_high = vdupq_n_u8(0);                                     \
                ret =                                                         \
                    vreinterpretq_m128i_u8(vextq_u8(tmp_low, tmp_high, idx)); \
```
It looks like the macro expansion for vextq_u8 is triggering the compiler error on a bounds check. The code would never execute because of the imm >= 16 check but the code is still generated and the const propagation looks to then trigger the error. Whether the last line get expanded I guess is down to the whim of the optimiser.

This is compiling with a M1 mac mini with the latest xcode:

```
% gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 12.0.5 (clang-1205.0.22.11)
Target: x86_64-apple-darwin20.6.0
Thread model: posix
```

marktwtn:
The test has been rewritten for testing more immediate values.
The selected values are able to represent different scenarios.

The implementation replaces vextq_u8() with vld1q_u8() since the latter
does not require constant immediate value.
As the immediate value is not required, the implementation can be
changed from macro to function.

The commit fixes the issue #482 as well.

Close #482.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -6545,29 +6545,23 @@ FORCE_INLINE __m64 _mm_abs_pi8(__m64 a)
 //   dst[127:0] := tmp[127:0]
 //
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_alignr_epi8
-#define _mm_alignr_epi8(a, b, imm)                                            \
-    __extension__({                                                           \
-        __m128i ret;                                                          \
-        if (_sse2neon_unlikely((imm) >= 32)) {                                \
-            ret = _mm_setzero_si128();                                        \
-        } else {                                                              \
-            uint8x16_t tmp_low, tmp_high;                                     \
-            if (imm >= 16) {                                                  \
-                const int idx = imm - 16;                                     \
-                tmp_low = vreinterpretq_u8_m128i(a);                          \
-                tmp_high = vdupq_n_u8(0);                                     \
-                ret =                                                         \
-                    vreinterpretq_m128i_u8(vextq_u8(tmp_low, tmp_high, idx)); \
-            } else {                                                          \
-                const int idx = imm;                                          \
-                tmp_low = vreinterpretq_u8_m128i(b);                          \
-                tmp_high = vreinterpretq_u8_m128i(a);                         \
-                ret =                                                         \
-                    vreinterpretq_m128i_u8(vextq_u8(tmp_low, tmp_high, idx)); \
-            }                                                                 \
-        }                                                                     \
-        ret;                                                                  \
-    })
+FORCE_INLINE __m128i _mm_alignr_epi8(__m128i a, __m128i b, int imm)
+{
+    if (_sse2neon_unlikely((unsigned int) imm >= 32))
+        return _mm_setzero_si128();
+    int idx;
+    uint8x16_t tmp[2];
+    if (imm >= 16) {
+        idx = imm - 16;
+        tmp[0] = vreinterpretq_u8_m128i(a);
+        tmp[1] = vdupq_n_u8(0);
+    } else {
+        idx = imm;
+        tmp[0] = vreinterpretq_u8_m128i(b);
+        tmp[1] = vreinterpretq_u8_m128i(a);
+    }
+    return vreinterpretq_m128i_u8(vld1q_u8(((uint8_t const *) tmp) + idx));
+}
 
 // Concatenate 8-byte blocks in a and b into a 16-byte temporary result, shift
 // the result right by imm8 bytes, and store the low 8 bytes in dst.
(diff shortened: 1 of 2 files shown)


tests/impl.cpp
```diff
@@ -6879,9 +6879,25 @@ result_t test_mm_alignr_epi8(const SSE2NEONTestImpl &impl, uint32_t iter)
 #else
     const uint8_t *_a = (const uint8_t *) impl.mTestIntPointer1;
     const uint8_t *_b = (const uint8_t *) impl.mTestIntPointer2;
-    // FIXME: The different immediate value should be tested in the future
-    const int shift = 18;
+    unsigned int shift;
```
jserv, line 6882: Rewrite as
```c
unsigned int shift = (iter % 5) << 3;
```

tests/impl.cpp
```diff
@@ -6899,7 +6915,24 @@ result_t test_mm_alignr_epi8(const SSE2NEONTestImpl &impl, uint32_t iter)
 
     __m128i a = load_m128i(_a);
     __m128i b = load_m128i(_b);
-    __m128i ret = _mm_alignr_epi8(a, b, shift);
+    __m128i ret;
+    switch (iter % 5) {
```
jserv, line 6919: Ditto. Rewrite as
```c
__m128i ret = _mm_alignr_epi8(a, b, (iter % 5) << 3);
```

marktwtn: Done.

tests/impl.cpp

tests/impl.cpp
marktwtn: The code can not be rewritten since the third parameter should be an immediate value, which means the value should be determined at compile time.

sse2neon.h
```diff
@@ -6545,29 +6545,27 @@ FORCE_INLINE __m64 _mm_abs_pi8(__m64 a)
...
-                ret =                                                         \
-                    vreinterpretq_m128i_u8(vextq_u8(tmp_low, tmp_high, idx)); \
-            }                                                                 \
-        }                                                                     \
-        ret;                                                                  \
-    })
+FORCE_INLINE __m128i _mm_alignr_epi8(__m128i a, __m128i b, int imm)
+{
+    __m128i ret;
+    if (_sse2neon_unlikely((unsigned int) imm >= 32)) {
```
jserv, line 6551: Rewrite as following:
```c
if (_sse2neon_unlikely((unsigned int) imm >= 32))
    return _mm_setzero_si128();

int idx;
...
```

PR #485, opened 2021-09-14
fix: Add precise implementation of _mm_dp_pd()

### LINKED ISSUE #480: Comment regarding _mm_dp_pd
This comment was pointed out to me regarding the sse2neon implemention "There's a small issue with the _mm_dp_pd impl, though: you do the mul first and then mask the result according to imm8[4:5]. That is the opposite order to what the sse41 definition says; the sse41 original will avoid unwanted ops, eg. on NaNs, your impl won't"

Reviewing https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_dp_pd&expand=2714 that does seem to be the case.

marktwtn:
The commit adds the new macro SSE2NEON_PRECISE_DP for precise
implementation of _mm_dp_pd().
The precise implementation removes the possible extra multiplications.
Enable the macro might decrease the performance but the behaviour
would be matched with the intrinsic description of Intel Intrinsics Guide.

Close #480.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Considering the balance between correctness and performance, `sse2neon` recogniz
 * `SSE2NEON_PRECISE_MINMAX`: Enable precise implementation of `_mm_min_ps` and `_mm_max_ps`. If you need consistent results such as NaN special cases, enable it.
 * `SSE2NEON_PRECISE_DIV`: Enable precise implementation of `_mm_rcp_ps` and `_mm_div_ps` by additional Netwon-Raphson iteration for accuracy.
 * `SSE2NEON_PRECISE_SQRT`: Enable precise implementation of `_mm_sqrt_ps` and `_mm_rsqrt_ps` by additional Netwon-Raphson iteration for accuracy.
+* `SSE2NEON_PRECISE_DP`: Enable precise implementation of `_mm_dp_pd`. When the conditional bit is not set, the corresponding multiplication would not be executed.
 
 The above are turned off by default, and you should define the corresponding macro(s) as `1` before including `sse2neon.h` if you need the precise implementations.
 
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -52,7 +52,7 @@
 
 /* Enable precise implementation of math operations
  * This would slow down the computation a bit, but gives consistent result with
- * x86 SSE2. (e.g. would solve a hole or NaN pixel in the rendering result)
+ * x86 SSE. (e.g. would solve a hole or NaN pixel in the rendering result)
  */
 /* _mm_min_ps and _mm_max_ps */
 #ifndef SSE2NEON_PRECISE_MINMAX
@@ -65,6 +65,10 @@
 /* _mm_sqrt_ps and _mm_rsqrt_ps */
 #ifndef SSE2NEON_PRECISE_SQRT
 #define SSE2NEON_PRECISE_SQRT (0)
+#endif
+/* _mm_dp_pd */
+#ifndef SSE2NEON_PRECISE_DP
+#define SSE2NEON_PRECISE_DP (0)
 #endif
 
 /* compiler specific definitions */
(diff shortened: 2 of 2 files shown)


jserv (changes requested):
Can you summarize the problem in the subject?

sse2neon.h
```diff
@@ -7521,18 +7521,25 @@ FORCE_INLINE __m128d _mm_dp_pd(__m128d a, __m128d b, const int imm)
...
+                                   vgetq_lane_f64(vreinterpretq_f64_m128d(b), 0)
+                             : 0;
+    double d1 = (imm & 0x20) ? vgetq_lane_f64(vreinterpretq_f64_m128d(a), 1) *
+                                   vgetq_lane_f64(vreinterpretq_f64_m128d(b), 1)
+                             : 0;
+#else
+    double d0 = (imm & 0x10) ? ((double *) &a)[0] * ((double *) &b)[0] : 0;
+    double d1 = (imm & 0x20) ? ((double *) &a)[1] * ((double *) &b)[1] : 0;
+#endif
+    mul = _mm_set_pd(d1, d0);
```
jserv, line 7537: Move the declaration of `__m128d mul` here.

sse2neon.h
```diff
@@ -7521,18 +7521,25 @@ FORCE_INLINE __m128d _mm_dp_pd(__m128d a, __m128d b, const int imm)
...
     const int64_t bit1Mask = imm & 0x02 ? UINT64_MAX : 0;
-    const int64_t bit4Mask = imm & 0x10 ? UINT64_MAX : 0;
-    const int64_t bit5Mask = imm & 0x20 ? UINT64_MAX : 0;
     // Conditional multiplication
-    __m128d mul = _mm_mul_pd(a, b);
-    const __m128d mulMask =
-        _mm_castsi128_pd(_mm_set_epi64x(bit5Mask, bit4Mask));
-    __m128d tmp = _mm_and_pd(mul, mulMask);
+    __m128d mul;
+#if defined(__aarch64__)
```
jserv, line 7526: So, no more bit mask?

marktwtn:
> Can you summarize the problem in the subject?

First, the behaviour of `_mm_dp_pd()` is to read the specific bit of the third input argument.
If the bit is set, then do the corresponding multiplication of the `__m128d` elelment and get the result.
If not, do not do the multiplication and set the result as `zero`.

The original implementation would do the multiplication to all the element in `__m128d` with SIMD.
It is efficient, but not correct.

If extra multiplication is done, it might cause exceptions or make the corresponding flags set.
For example, the unwanted multiplication might multiply with NaN, or cause overflow, underflow or invalid operations.
That would be the potential bugs.

sse2neon.h
marktwtn: The bit mask for conditional multiplication is no longer required.

sse2neon.h

jserv:
By the way, can we provide the shortcuts which ensure faster execution paths for intrinsics such as `_mm_dp_pd`? That is, some configuration(s) as what `SSE2NEON_PRECISE_MINMAX`, `SSE2NEON_PRECISE_DIV`, and `SSE2NEON_PRECISE_SQRT` do. The consideration for NaN check always consumes system resources.

marktwtn:
> By the way, can we provide the shortcuts which ensure faster execution paths for intrinsics such as `_mm_dp_pd`? That is, some configuration(s) as what `SSE2NEON_PRECISE_MINMAX`, `SSE2NEON_PRECISE_DIV`, and `SSE2NEON_PRECISE_SQRT` do. The consideration for NaN check always consumes system resources.

I guess we could add configuration like `SSE2NEON_PRECISE_DP` or other with better naming.

marktwtn:
The new configuration `SSE2NEON_PRECISE_DP` has been added.

README.md
```diff
@@ -68,6 +68,7 @@ Considering the balance between correctness and performance, `sse2neon` recogniz
 * `SSE2NEON_PRECISE_MINMAX`: Enable precise implementation of `_mm_min_ps` and `_mm_max_ps`. If you need consistent results such as NaN special cases, enable it.
 * `SSE2NEON_PRECISE_DIV`: Enable precise implementation of `_mm_rcp_ps` and `_mm_div_ps` by additional Netwon-Raphson iteration for accuracy.
 * `SSE2NEON_PRECISE_SQRT`: Enable precise implementation of `_mm_sqrt_ps` and `_mm_rsqrt_ps` by additional Netwon-Raphson iteration for accuracy.
+* `SSE2NEON_PRECISE_DP`: Enable precise implementation of `_mm_dp_pd` for avoiding unwanted multiplication.
```
jserv, line 71: I didn't get it. What was "avoiding unwanted multiplication" intended for?

marktwtn: It means when the specific bit is not set, the corresponding multiplication would not be done.

jserv: Improve the above description.

marktwtn: Done.

PR #488, opened 2021-09-15
test: Add compiler detection for NaN comparison testing

marktwtn:
According to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98612,
the intrinsics _mm_comi[eq|le|lt|neq]_[sd|ss] do not return the
correct value when comparing with NaN values.
The testing would be unimplemented when using GCC.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -72,13 +72,30 @@ public:
             if (test == it_mm_cmpord_ps || test == it_mm_comilt_ss ||
                 test == it_mm_comile_ss || test == it_mm_comige_ss ||
                 test == it_mm_comieq_ss || test == it_mm_comineq_ss ||
-                test == it_mm_comigt_ss) {  // if testing for NAN's make sure we
-                                            // have some nans
-                // One out of four times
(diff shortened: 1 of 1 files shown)


jserv:
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98612

 Hong-tao Liu at Intel fixed the problem this year: https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=240f0a490dae0fb9ef72fa21a93e8088d17fb682
Maybe we should list the gcc versions which are known to comply with Intel Intrinsics.

marktwtn:
> Hong-tao Liu at Intel fixed the problem this year: https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=240f0a490dae0fb9ef72fa21a93e8088d17fb682
> Maybe we should list the gcc versions which are known to comply with Intel Intrinsics.

I thought Hong-tao Liu means that the issue would be fixed later and the commit is for removing the dead code only.
Nevertheless, I still don't know which GCC version has fixed the issue from the commit page.

PR #491, opened 2021-09-17
fix: Fix the intrinsics of comparing floating points

marktwtn:
Take not greater than (ngt) as an example:
When comparing with NaN values, not greater than (ngt) is not equal
to less equal (le).
The implementation and testing of not greater than (ngt) should be the reverse
of greater than (gt).
Hence, the testing should add the input data of NaN values and the
implementation and testing should be rewritten.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -69,10 +69,24 @@ public:
                 mTestFloatPointer1[3] = mTestFloatPointer2[3];
             }
 
-            if (test == it_mm_cmpord_ps || test == it_mm_comilt_ss ||
-                test == it_mm_comile_ss || test == it_mm_comige_ss ||
-                test == it_mm_comieq_ss || test == it_mm_comineq_ss ||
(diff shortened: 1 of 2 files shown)


tests/impl.cpp
```diff
@@ -94,10 +118,14 @@ class SSE2NEONTestImpl : public SSE2NEONTest
                 // Make sure the NaN values are included in the testing
                 // one out of four times.
                 if ((rand() & 3) == 0) {
+                    // FIXME:
+                    // The argument "0xFFFFFFFFFFFF" is a tricky workaround to
```
jserv, line 122: Can you avoid using string literal "0xFFFFFFFFFFFF" ?

marktwtn: Any suggestion?
I would like to leave the whole `FIXME` comment and fix it in the future pull request.
To avoid using `0xFFFFFFFFFFFF` in `nanf()`, the testing code structure should be rewritten.

PR #492, opened 2021-09-28
refactor: Extract unordered comparison as function

marktwtn:
The extracted functions are refactored to avoid extra variable
declarations.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -348,24 +348,24 @@ result_t do_mm_store_ps(int32_t *p, int32_t x, int32_t y, int32_t z, int32_t w)
     return TEST_SUCCESS;
 }
 
-float compord(float a, float b)
+float cmp_noNaN(float a, float b)
 {
-    float ret;
-
-    bool isNANA = isnan(a);
-    bool isNANB = isnan(b);
-    ret = (!isNANA && !isNANB) ? ALL_BIT_1_32 : 0.0f;
-    return ret;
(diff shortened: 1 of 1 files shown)


tests/impl.cpp
```diff
@@ -350,22 +350,22 @@ result_t do_mm_store_ps(int32_t *p, int32_t x, int32_t y, int32_t z, int32_t w)
...
 {
-    double ret;
+    return (!isnan(a) && !isnan(b)) ? ALL_BIT_1_64 : 0.0f;
+}
 
-    bool isNANA = isnan(a);
-    bool isNANB = isnan(b);
-    ret = (!isNANA && !isNANB) ? ALL_BIT_1_64 : 0.0f;
-    return ret;
+float compunord(float a, float b)
```
jserv, line 361: Shall we revise the function names for readability?

marktwtn: Like `cmpunord` or `existNaNOp`?

jserv: We can introduce the name scheme such as `cmp_` (for comparison) `isNaN` (`NaN` may look better than `NAN`).

jserv: By the way, this document is useful: https://www.agner.org/optimize/nan_propagation.pdf

marktwtn: I will rewrite `compord` as `cmp_noNaN` and `compunord` as `cmp_hasNaN`.

PR #493, opened 2021-09-30
fix: Fix _mm_[max|min]_[ps|ss|pd|sd] and tests

marktwtn:
Extract tunable configurations to a separate header file config.h.
Add the precise implementation for _mm_[min|max]_[pd|sd].
Add the testing which includes NaN inputs for max/min comparsion.
Replace fmax, fmin with pure C comparison.
When one of the operand is NaN, fmax and fmin always return the other
operand which is not a NaN.
In contrast, SSE intrinsics would always return the second operand.


README.md
```diff
@@ -71,6 +71,7 @@ Considering the balance between correctness and performance, `sse2neon` recogniz
 * `SSE2NEON_PRECISE_DP`: Enable precise implementation of `_mm_dp_pd`. When the conditional bit is not set, the corresponding multiplication would not be executed.
 
 The above are turned off by default, and you should define the corresponding macro(s) as `1` before including `sse2neon.h` if you need the precise implementations.
+The configuration macros are in the `config.h`.
```
jserv, line 74: Don't do that. At present, SSE2NEON follows the single-file style as [stb](https://github.com/nothings/stb) does. Only one header file is deployed and used. The extra `config.h` might cause unexpected path matching order.

marktwtn: I will find another way to handle the testing without extra header file.

marktwtn: I reverse it and add the macro in `tests/common.h` when the testing suite does not include `sse2neon.h`.

tests/impl.cpp
```diff
@@ -129,6 +129,36 @@ class SSE2NEONTestImpl : public SSE2NEONTest
...
             }
 
+#if SSE2NEON_PRECISE_MINMAX
+            if (test == it_mm_max_ps || test == it_mm_max_ss ||
+                test == it_mm_min_ps || test == it_mm_min_ss) {
+                // Make sure the NaN values are included in the testing
+                // one out of four times.
+                if ((rand() & 3) == 0) {
+                    uint32_t r1 = rand() & 3;
+                    uint32_t r2 = rand() & 3;
```
jserv, line 139: What if `r1` equals to `r2`? Would you enforce different indexes?

tests/impl.cpp
```diff
@@ -129,6 +129,36 @@ class SSE2NEONTestImpl : public SSE2NEONTest
...
+            if (test == it_mm_max_pd || test == it_mm_max_sd ||
+                test == it_mm_min_pd || test == it_mm_min_sd) {
+                // Make sure the NaN values are included in the testing
+                // one out of four times.
+                if ((rand() & 3) == 0) {
+                    // FIXME:
+                    // The argument "0xFFFFFFFFFFFF" is a tricky workaround to
+                    // set the NaN value for doubles. The code is not intuitive
+                    // and should be fixed in the future.
+                    uint32_t r1 = ((rand() & 1) << 1) + 1;
```
jserv, line 154: Ditto. I am not sure if it makes sense if `r1` and `r2` have the same value as indexing.

tests/impl.cpp
marktwtn: The indexes can be the same or different since the situation of both operands are `NaN` should be tested as well.

PR #495, opened 2021-10-07
Add _mm_getcsr and rounding defines

nlacey:
Add _mm_getcsr and rounding defines

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -158,22 +158,13 @@
 #define _MM_FROUND_TO_ZERO 0x03
 #define _MM_FROUND_CUR_DIRECTION 0x04
 #define _MM_FROUND_NO_EXC 0x08
-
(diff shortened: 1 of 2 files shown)


jserv:
Please manage to implement the corresponding test item for `_mm_getscr`.

nlacey:
I'm a little confused with the unit tests running on x86?
when I try and run on my x86-64 (on master branch) I get 
tests/impl.cpp:2023:19: error: ‘_mm_loadu_si64’ was not declared in this scope.

marktwtn:
The commit header should be `_mm_getcsr` not `_mm_getscr`.

nlacey:
I saved off the original csr value and restored the original csr value after the unit test was done so I shouldn't modify the running of other tests now.

marktwtn:
@nlacey use `$ make indent` to fix the coding style, and please rebase the commits to a single one.

PR #503, opened 2021-11-18
fix 2 preprocessor checks and static const correctness

toxieainc:


jserv:
Thank @toxieainc for contributing!

PR #508, opened 2022-01-09
asm volatile -> __asm__ __volatile__

sleepybishop:
There are places where `asm volatile` is used and others which use `__asm__ __volatile__`.

This PR makes consistent use of the latter since it's more widely supported.


jserv:
Thank @sleepybishop for contributing!

PR #509, opened 2022-01-11
Fix various typos

luzpaz:
Found via downstream https://github.com/embree/embree


jserv:
Thank @luzpaz for contributing!

Issue #505, opened 2021-12-11
Is the current sse2neon.h accepted by C compilers?


h-murai:
Hi, I think the line 6439 isn't correct as a C statement.

`    static const __m128d mask = _mm_set_pd(1.0f, -1.0f);`

gcc actually detects an error of "initializer element is not a compile-time constant" for the line.

jserv:
Reported by recent [GitHub Actions result](https://github.com/DLTcollab/sse2neon/actions/runs/1510120922), both gcc-10 and clang-11 are able to build the test suite without compilation errors. Can you show more information about the package and configurations?

h-murai:
I used GCC8 and Spack v0.17.0 to install the bwa package, which depends on sse2neon. Ok. I'll try it again with GCC10. Thanks.

QwertyJack:
The same here. Using `g++` to compile everything works like a charm, while `gcc` failed. Btw, I have gcc-7.5 for aarch64.

QwertyJack:
The C language does not allow initializing a variable with a function call.
The testing is all C++, but it should fail with `gcc` and a `.c` file.

jserv:
@QwertyJack, I managed to reproduce the issue. Can you check the following steps?
1. Download GNU Toolchain based on gcc-7.5: https://releases.linaro.org/components/toolchain/binaries/latest-7/
2. Run "make check" to see if the builtin test suite would pass or not.

Alternatively, please provide the test program and the corresponding procedure for me to reproduce.

QwertyJack:
As I said, this is a language "feature", not related to the compiler (please see [this post](https://stackoverflow.com/questions/3025050/error-initializer-element-is-not-constant-when-trying-to-initialize-variable-w) for details). 

The test should pass, because all files are ended with `.cpp` and the compiler will treat these files as C++ langeuage.
A minimal example to reproduce the issue would be like this:

```c
// a.c

#include "sse2neon.h"

int main() {
    static const __m128d mask = _mm_set_pd(1.0f, -1.0f);
    return 0;
}
```

The C compiler will complain about `error: initializer element is not constant` when compiling with: `gcc a.c`.
However, everything would work if a C++ compiler were called, e.g.:

- use `g++` instead of `gcc`: `g++ a.c`
- change the suffix: `cp a.c a.cpp && gcc -c a.cpp`
- force GCC to call the C++ backend: `gcc -xc++ -c a.c`

jserv:
[...]

PR #511, opened 2022-01-25
Remove static keyword

efbicief:
The current version does not compile - removing this keyword lets it compile, and whilst it would be better to find an elegant way to keep it static, it is better for it to work


jserv:
Append `Close #505` at the end of git commit message.

jserv:
Close this pull request in favor of #516

PR #520, opened 2022-03-13
Add support for 32-bit targets on ARMv8 architectures

jonathanhue:
This adds support for using sse2neon when building 32-bit executables to run on ARMv8 architectures.  Prior to this change, you could not use sse2neon when building a 32-bit executable while targeting ARMv8.  So you couldn't do this:

$ clang++ -O2 -mcpu=cortex-a73 -mfpu=neon-fp-armv8 myprogram.cpp

when the target was armv7l-unknown-linux-gnueabihf, as sse2neon looks for __ARM_ARCH == 7, or __aarch64__ defined, and neither is true for a 32-bit executable targeting ARMv8.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,13 @@ but SSE intrinsic `_mm_maddubs_epi16` has to be implemented with 13+ NEON instru
 ```
 
 - Explicitly specify platform-specific options to gcc/clang compilers.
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Change the content "Usage" in file `README.md` to reflect the description of NEON specific configurations.

jserv:
To satisfy the consistent indention, please run `clang-format-12 -i sse2neon.h` and submit again. Versions after 12 should work out of box.

jserv:
Thank @jonathanhue for contributing!

PR #522, opened 2022-03-16
Use CRC and directed rounding intrinsics on A32

jonathanhue:
Enable the use of CRC and directed rounding intrinsics when
targeting ARMv8 32-bit.  These are intrinsics that were added
in ARMv8 for both the 32-bit execution mode and AArch64.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -126,13 +126,20 @@
 #error \
     "You must enable NEON instructions (e.g. -mfpu=neon-fp-armv8) to use SSE2NEON."
 #endif
+#if !defined(__clang__)
 #pragma GCC push_options
+#endif
 #else
 #error "Unsupported target. Must be either ARMv7-A+NEON or ARMv8-A."
 #endif
 #endif
 
 #include <arm_neon.h>
+#if !defined(__aarch64__) && (__ARM_ARCH == 8)
+#if defined __has_include && __has_include(<arm_acle.h>)
+#include <arm_acle.h>
+#endif
+#endif
 
 /* Rounding functions require either Aarch64 instructions or libm failback */
 #if !defined(__aarch64__)
@@ -680,7 +687,8 @@ FORCE_INLINE void _sse2neon_kadd_f32(float *sum, float *c, float y)
     *sum = t;
 }
 
-#if defined(__ARM_FEATURE_CRYPTO)
+#if defined(__ARM_FEATURE_CRYPTO) && \
+    (defined(__aarch64__) || __has_builtin(__builtin_arm_crypto_vmullp64))
 // Wraps vmull_p64
 FORCE_INLINE uint64x2_t _sse2neon_vmull_p64(uint64x1_t _a, uint64x1_t _b)
 {
@@ -1376,7 +1384,7 @@ FORCE_INLINE __m128 _mm_cvt_pi2ps(__m128 a, __m64 b)
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvt_ps2pi
 FORCE_INLINE __m64 _mm_cvt_ps2pi(__m128 a)
 {
-#if defined(__aarch64__)
+#if defined(__aarch64__) || defined(__ARM_FEATURE_DIRECTED_ROUNDING)
     return vreinterpret_m64_s32(
         vget_low_s32(vcvtnq_s32_f32(vrndiq_f32(vreinterpretq_f32_m128(a)))));
 #else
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -126,13 +126,22 @@
...
 #pragma GCC push_options
+#endif
 #else
 #error "Unsupported target. Must be either ARMv7-A+NEON or ARMv8-A."
 #endif
 #endif
 
 #include <arm_neon.h>
+#if !defined(__aarch64__) && (__ARM_ARCH == 8)
+#if defined __has_include
```
jserv, line 139: Rewrite as the following for consistency.
```c
#if defined(__has_include) && __has_include(<arm_acle.h>)
#include <arm_acle.h>
#endif
```

sse2neon.h
```diff
@@ -680,7 +689,7 @@ FORCE_INLINE void _sse2neon_kadd_f32(float *sum, float *c, float y)
     *sum = t;
 }
 
-#if defined(__ARM_FEATURE_CRYPTO)
+#if defined(__aarch64__) && defined(__ARM_FEATURE_CRYPTO)
```
jserv, line 692: I am not confident to use `__ARM_ARCH == 8` or `__aarch64__` for such pre-condition.
Can you unify the convention?

jonathanhue: The code that follows that uses intrinsics that are only available on aarch64, so it needs that test.  32-bit ARMv8 targets can support CRC (most do, that's why clang enables it by default), and will define __ARM_FEATURE_CRYPTO with the correct compiler options (-mfpu=crypto-neon-fp-armv8), and thus need to be prevented from trying to compile that code.

sse2neon.h
jonathanhue: I did it that way because https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html claimed combining the tests into a single expression wouldn't work in environments where __has_include wasn't available.  But GCC also says && in the preprocessor follows the same short-circuit rules as C, so I don't know why they'd claim it wouldn't work.  I'll change it.

sse2neon.h
jonathanhue: Ok, I see what's going on.  The support for the polynomial intrinsics that use poly64_t types is still missing for AArch32 in clang 12.  vreinterpret_p64_u64, vget_lane_p64, and vmull_p64 are all unavailable.

GCC supports them on AArch32, at least back through version 7.5.  So that code should be enabled for \_\_aarch64__, and AArch32 when compiled with GCC.  Since only ARMv8 and higher can set __ARM_FEATURE_CRYPTO, it isn't necessary to test for ARMv8 so that line could read:

`#if (!defined(__clang__) || defined(__aarch64__)) && defined(__ARM_FEATURE_CRYPTO)`

jonathanhue: One more thing: I'm not sure the __ARM_FEATURE_CRYPTO test is correct or even needed here.  That macro indicates support for the AES/SHA1/SHA256/SHA512 Neon intrinsics.  The code that is inside that test doesn't use any of those.  It really needs to be wrapped inside a test for poly64_t polynomial intrinsics support.  That support is supported in baseline ARMv8 for both 32-bit and 64-bit, but toolchain support only seems complete for AArch64, and AArch32 using GCC.

jonathanhue: It looks like __ARM_FEATURE_CRYPTO is required because of what appears to be an error in GCC's arm_neon.h for 32-bit ARM.  The poly64_t stuff is wrapped with  #pragma GCC target ("fpu=crypto-neon-fp-armv8"), so if you don't enable crypto support on the command line (which then defines __ARM_FEATURE_CRYPTO), you'll get a compilation error.  So the inclusion of __ARM_FEATURE_CRYPTO almost looks like something that was needed to work around this problem.

GCC's arm_neon.h for aarch64 does not do this - it doesn't care whether you've specified crypto support or not.

So to make this behavior match what it was before, but also enable it for GCC targeting AArch32, I might do something like:

```
#if defined(__ARM_FEATURE_CRYPTO) && \
   (defined(__aarch64__) || __has_builtin(__builtin_arm_crypto_vmullp64))
```

jserv: Exactly, you can even add more comments around polynomial intrinsics.
Should the original `vmull_p8` fallback for A32 targets without polynomial intrinsics (clang?) be modified as well?

jonathanhue: vzip1q_u64 and vzip2q_u64 are only available on A64, so that code needs to remain conditional on __aarch64__.

I also confirmed that GCC's arm_neon.h for 64-bit ARM places a similar (seemingly unnecessary) requirement on vmull_p64 to compile with an option that enables crypto, so the test for __ARM_FEATURE_CRYPTO needs to stay the way it is.

PR #523, opened 2022-03-18
fix: Fix alignment in tests

### LINKED ISSUE #521: Crash in test_mm_stream_si128() due to unaligned access
Crash has been observed with 32-bit executables running on actual ARM hardware running both 32-bit and 64-bit operating systems.  The crash is due to trying to store to an insufficiently aligned address.  _mm_stream_si128 requires that the destination address be aligned on a 16-byte boundary, otherwise a general protection fault may occur.  sse2neon maps this to vst1q_s64, and the address only has 4-byte alignment, so a bus error occurs.  So the test needs to be fixed to align the memory address. 

Changing:

` int32_t p[4];
`

to

` int32_t p[4] __attribute__ ((aligned (16)));
`

is one way to enable the test to pass.  Note that _mm_store_si128 has the same alignment requirements as _mm_stream_si128, so test_mm_store_si128 would also crash on an actual x86 if p were not sufficiently aligned.  It does not crash on an ARM though because _mm_store_si128 is mapped to vst1q_s32, and it has relaxed alignment requirements in my environment.

HowJMay:
_mm_store_si128 and _mm_stream_si128 are aligned in 16 bytes

close #521

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -2,6 +2,7 @@
 #include <assert.h>
 #include <float.h>
 #include <math.h>
+#include <stdalign.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -2963,7 +2964,7 @@ result_t test_mm_stream_ps(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
     const float *_a = impl.mTestFloatPointer1;
     __m128 a = load_m128(_a);
-    float p[4];
+    alignas(16) float p[4];
 
     _mm_stream_ps(p, a);
     ASSERT_RETURN(p[0] == _a[0]);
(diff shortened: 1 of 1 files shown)


jserv:
@jonathanhue, Can you help review this patch?

jonathanhue:
Maybe use ALIGN_STRUCT(x) defined in common.h instead of  __attribute__ ((aligned(16))).  I'm not sure the latter is supported when building on Windows w/Visual Studio.

jserv:
There is an operation `#pragma pop_macro("ALIGN_STRUCT")` in file `tests/common.h`. We might avoid restoring the macro `ALIGN_STRUCT` by eliminating the use of `push_macro` and `pop_macro`. Thus, we can explicitly use `ALIGN_STRUCT` in test programs.

jserv:
Alternatively, we can make use of C11 [_Alignas](https://en.cppreference.com/w/c/language/_Alignas) without caring about compiler extensions.

jserv:
@jonathanhue, Please review and see if we encounter further issues.

jonathanhue (commented):
Looks fine.  Some old versions of GCC won't honor an alignment request > 8 via alignas when targeting 32-bit ARM, but will honor \_\_attribute__((aligned(16))).  I think that's ok for the tests.

Issue #530, opened 2022-05-27
Implement `__rdtsc`


Cuda-Chen:
Currently I am implementing the `_rdtsc` Intel intrinsic function.
I confirm I follow the instruction of [adding test case](https://github.com/DLTcollab/sse2neon/blob/master/tests/README.md).
However, whenever I tried to `make check` on Intel platform, I always receive the following compile error:
```bash
$ make check
g++ -o tests/binding.o -Wall -Wcast-qual -I. -maes -mpclmul -mssse3 -msse4.2 -std=gnu++14 -c -MMD -MF tests/binding.o.d tests/binding.cpp
g++ -o tests/common.o -Wall -Wcast-qual -I. -maes -mpclmul -mssse3 -msse4.2 -std=gnu++14 -c -MMD -MF tests/common.o.d tests/common.cpp
g++ -o tests/impl.o -Wall -Wcast-qual -I. -maes -mpclmul -mssse3 -msse4.2 -std=gnu++14 -c -MMD -MF tests/impl.o.d tests/impl.cpp
tests/impl.cpp: In function ‘SSE2NEON::result_t SSE2NEON::test_rdtsc(const SSE2NEON::SSE2NEONTestImpl&, uint32_t)’:
tests/impl.cpp:9402:22: error: ‘_rdtsc’ was not declared in this scope; did you mean ‘it_rdtsc’?
 9402 |     uint64_t start = _rdtsc();
      |                      ^~~~~~
      |                      it_rdtsc
```

For your convenience, I attach the changed I have made:
- sse2neon.h
```c
// _rdtsc declaration
FORCE_INLINE uint64_t _rdtsc(void);
...
// _rdtsc definition
#if defined(__x86_64__) || defined(__i386__)
FORCE_INLINE uint64_t _rdtsc(void)
{
    unsigned hi, lo;
    __asm__ __volatile__("rdtsc" : "=a"(lo), "=d"(hi));
    return ((uint64_t) lo) | (((uint64_t) hi) << 32); 
}

#elif defined(__aarch64__)
FORCE_INLINE uint64_t _rdtsc(void)
{
    uint64_t val; 

    /* According to ARM DDI 0487F.c, from Armv8.0 to Armv8.5 inclusive, the
     * system counter is at least 56 bits wide; from Armv8.6, the counter
     * must be 64 bits wide.  So the system counter could be less than 64
     * bits wide and it is attributed with the flag 'cap_user_time_short'
     * is true.
     */
[...]

PR #532, opened 2022-06-02
feat: Implement _rdtsc

Cuda-Chen:
Related to #530

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -146,6 +146,15 @@
 #include <math.h>
 #endif
 
+/* On ARMv7, some registers, such as PMUSERENR and PMCCNTR, are read-only
+ * or even not accessible in user mode.
+ * To write or access to these registers in user mode,
+ * we have to perform syscall instead.
+ */
+#if !defined(__aarch64__)
+#include <sys/time.h>
+#endif
+
 /* "__has_builtin" can be used to query support for built-in functions
  * provided by gcc/clang and other compilers that support it.
  */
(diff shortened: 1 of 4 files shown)


sse2neon.h
```diff
@@ -146,6 +146,11 @@
 #include <math.h>
 #endif
 
+/* Fallback to syscall as some registers require privilege mode on ARMv7. */
+#if defined(__arm__)
```
jserv, line 150: Change to `#if !defined(__aarch64__)` and explain the fallback.

sse2neon.h
```diff
@@ -415,6 +420,10 @@ FORCE_INLINE __m128d _mm_round_pd(__m128d, int);
 FORCE_INLINE __m128 _mm_round_ps(__m128, int);
 // SSE4.2
 FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t, uint8_t);
+// Others
+#if defined(__aarch64__)
+FORCE_INLINE uint64_t _rdtsc(void);
```
jserv, line 425: There is no need for forward declaration.

sse2neon.h
```diff
@@ -8789,6 +8798,47 @@ FORCE_INLINE void _sse2neon_mm_set_denormals_zero_mode(unsigned int flag)
...
+     * system counter is at least 56 bits wide; from Armv8.6, the counter
+     * must be 64 bits wide.  So the system counter could be less than 64
+     * bits wide and it is attributed with the flag 'cap_user_time_short'
+     * is true.
+     */
+    asm volatile("mrs %0, cntvct_el0" : "=r"(val));
+
+    return val;
+}
+#elif defined(__arm__)
```
jserv, line 8818: Simply use `#else` since we only care about Aarch64 and Arm32.

sse2neon.h
```diff
@@ -8789,6 +8798,47 @@ FORCE_INLINE void _sse2neon_mm_set_denormals_zero_mode(unsigned int flag)
...
+FORCE_INLINE uint64_t _rdtsc(void)
+{
+    uint32_t pmccntr;
```
jserv, line 8821: Rewrite as ` uint32_t pmccntr, pmuseren, pmcntenset;` for short.

jserv (changes requested):
Append `Close #530` at the end of git commit message.

tests/impl.cpp
```diff
@@ -9393,6 +9395,15 @@ result_t test_mm_set_denormals_zero_mode(const SSE2NEONTestImpl &impl,
     return TEST_SUCCESS;
 }
 
+result_t test_rdtsc(const SSE2NEONTestImpl &impl, uint32_t iter)
+{
+    uint64_t start = _rdtsc();
+    for (int i = 0; i < 100000; i++)
+        ;
```
jserv, line 9402: Add an explicit compiler barrier here. That is, `asm volatile("" ::: "memory");`
Reference: https://preshing.com/20120625/memory-ordering-at-compile-time/

jserv:

PR #534, opened 2022-08-03
Implement _mm_cmpestri and _mm_cmpestrm

Cuda-Chen:
1. Implement `_mm_cmpestri` and `_mm_cmpestrm`.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -25,6 +25,7 @@
 //   Yang-Hao Yuan <yanghau@biilabs.io>
 //   Syoyo Fujita <syoyo@lighttransport.com>
 //   Brecht Van Lommel <brecht@blender.org>
+//   Cuda Chen <clh960524@gmail.com>
 
 /*
  * sse2neon is freely redistributable under the MIT License.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ ifeq ($(processor),$(filter $(processor),aarch64 arm64))
 else ifeq ($(processor),$(filter $(processor),i386 x86_64))
     ARCH_CFLAGS = -maes -mpclmul -mssse3 -msse4.2
 else ifeq ($(processor),$(filter $(processor),arm armv7l))
-    ARCH_CFLAGS = -mfpu=neon
+    ARCH_CFLAGS = -mfpu=neon -marm
```
jserv, line 32: We should avoid specifying the option `-marm` since it can be set internally in toolchain defaults.

sse2neon.h
```diff
@@ -8358,6 +8359,649 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        "sub %w[b], w1, %w[b], asr #31\n"
+        "cmp %w[a], %w[bd]\n"
+        "csel %w[a], %w[bd], %w[a], gt\n"
+        "cmp %w[b], %w[bd]\n"
+        "csel %w[b], %w[bd], %w[b], gt\n"
+        : [a] "+r"(la), [b] "+r"(lb)
+        : [bd] "r"(upper)
+        : "w0", "w1");
+#else
+    __asm__ __volatile__(
```
jserv, line 8920: The instruction `csel` is essentially like the ternary operator in C, and it is absent in ARMv7-A. Thus, we can have C implementation for Arm32 rather than assembly listing.

sse2neon.h
```diff
@@ -8358,6 +8359,649 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        "sub %w[b], w1, %w[b], asr #31\n"
+        "cmp %w[a], %w[bd]\n"
+        "csel %w[a], %w[bd], %w[a], gt\n"
+        "cmp %w[b], %w[bd]\n"
+        "csel %w[b], %w[bd], %w[b], gt\n"
+        : [a] "+r"(la), [b] "+r"(lb)
+        : [bd] "r"(bound)
+        : "w0", "w1");
+#else
+    __asm__ __volatile__(
```
jserv, line 8962: Ditto. Avoid using Arm assembly.

sse2neon.h
Cuda-Chen: For my curiosity, what's the reason not to use inline assembly but use C implementation instead in ARMv7-A? Currently I can only come up with the idea of best portability between ARM and thumb instruction set.

jserv: If we implement in generic C, we can benefit from compiler optimizations without paying attention to arm-thumb interworking.

sse2neon.h
```diff
@@ -8358,6 +8359,638 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        "csel %w[b], %w[bd], %w[b], gt\n"
+        : [a] "+r"(la), [b] "+r"(lb)
+        : [bd] "r"(upper)
+        : "w0", "w1");
+#else
+    int tmp1 = la ^ (la >> 31);
+    la = tmp1 - (la >> 31);
+    int tmp2 = lb ^ (lb >> 31);
+    lb = tmp2 - (lb >> 31);
+    la = la > upper ? upper : la;
```
jserv, line 8924: Make use of `min` or `max` macro.

sse2neon.h
```diff
@@ -8358,6 +8359,638 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    _MM_CMPINT_LE = 2,    /* Less than or Equal */
+    _MM_CMPINT_FALSE = 3, /* Always False */
+    _MM_CMPINT_NE = 4,    /* Not Equal */
+    _MM_CMPINT_NLT = 5,   /* Not Less than */
+    _MM_CMPINT_NLE = 6,   /* Not Less than or Equal */
+    _MM_CMPINT_TRUE = 7   /* Always True */
+} _MM_CMPINT_ENUM;
+
+static uint16_t g_mask_epi16[8] __attribute__((aligned(16))) = {
+    0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80};
```
jserv, line 8374: Add a comma (`,`) right after `0x80`, so that `clang-format` can indent better.

sse2neon.h
```diff
@@ -8358,6 +8359,638 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+static uint8_t g_mask_epi8[16] __attribute__((aligned(16))) = {
+    0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80,
+    0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80};
```
jserv, line 8377: Ditto. Indention.

sse2neon.h
```diff
@@ -8358,6 +8359,638 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    _MM_CMPINT_LT = 1,    /* Less than */
+    _MM_CMPINT_LE = 2,    /* Less than or Equal */
+    _MM_CMPINT_FALSE = 3, /* Always False */
+    _MM_CMPINT_NE = 4,    /* Not Equal */
+    _MM_CMPINT_NLT = 5,   /* Not Less than */
+    _MM_CMPINT_NLE = 6,   /* Not Less than or Equal */
+    _MM_CMPINT_TRUE = 7   /* Always True */
+} _MM_CMPINT_ENUM;
+
+static uint16_t g_mask_epi16[8] __attribute__((aligned(16))) = {
```
jserv, line 8373: Both `g_mask_epi16` and `g_mask_epi8` are not meaningful variable names. Can you improve?

sse2neon.h
```diff
@@ -8358,6 +8359,638 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+                     vreinterpretq_u8_m128i(a)));                              \
+        mtx[15] = vreinterpretq_m128i_u8(                                      \
+            vceqq_u8(vdupq_n_u8(vgetq_lane_u8(vreinterpretq_u8_m128i(b), 15)), \
+                     vreinterpretq_u8_m128i(a)));                              \
+    } while (0)
+
+#define PCMPSTR_RNG_U16x8(a, b, mtx)                                           \
+    do {                                                                       \
+        uint16x8_t vec_b[8];                                                   \
+        __m128i mask = vreinterpretq_m128i_u32(vdupq_n_u32(0xffff));           \
```
jserv, line 8489: Instead of manually unrolling loops, we can explicitly use GNU extension `#pragma unroll`.
See [Arm Compiler for Embedded User Guide: Optimizing loops](https://developer.arm.com/documentation/100748/0618/Writing-Optimized-Code/Optimizing-loops)

Cuda-Chen: So after the compilation on ARMv8-A and ARMv7-A, it seems that `vgetq_lane_xxx` needs constant intermediate. Taking ARMv7-A as example:
```
In function ‘uint8_t vgetq_lane_u8(uint8x16_t, int)’,
    inlined from ‘int _sse2neon_cmp_byte_equal_any(__m128i, int, __m128i, int)’ at ./sse2neon.h:8558:5:
/usr/lib/gcc-cross/arm-linux-gnueabihf/9/include/arm_neon.h:6274:50: error: argument 2 must be a constant immediate
 6274 |   return (uint8_t)__builtin_neon_vget_laneuv16qi ((int8x16_t) __a, __b);
      |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
      ```

File: tests/impl.cpp [hunk 9]
```diff
@@ -9149,14 +9149,71 @@ result_t test_mm_cmpestrc(const SSE2NEONTestImpl &impl, uint32_t iter)
...
 {
-    return TEST_UNIMPL;
+    int16_t *a = test_mm_cmpestrm_data.a, *b = test_mm_cmpestrm_data.b;
+    int la = test_mm_cmpestrm_data.la, lb = test_mm_cmpestrm_data.lb;
+    const int imm8 = test_mm_cmpestrm_data.imm8;
+    int16_t *expect = test_mm_cmpestrm_data.expect;
+    __m128i ma, mb, res;
+    memcpy(&ma, a, sizeof(ma));
+    memcpy(&mb, b, sizeof(mb));
+#if defined(__arm__) || defined(__aarch64__)
```
jserv, line 9209: Can you avoid the special cases (Arm/Aarch64) for invoking `_mm_cmpestrm`?

tests/impl.cpp
```diff
@@ -9149,14 +9149,71 @@ result_t test_mm_cmpestrc(const SSE2NEONTestImpl &impl, uint32_t iter)
...
-    return TEST_UNIMPL;
-}
+    int16_t *a = test_mm_cmpestri_data.a, *b = test_mm_cmpestri_data.b;
+    int la = test_mm_cmpestri_data.la, lb = test_mm_cmpestri_data.lb;
+    const int imm8 = test_mm_cmpestri_data.imm8;
+    int expect = test_mm_cmpestri_data.expect;
+    __m128i ma, mb;
+    memcpy(&ma, a, sizeof(ma));
+    memcpy(&mb, b, sizeof(mb));
+#if defined(__arm__) || defined(__aarch64__)
```
jserv, line 9176: Can you avoid the special cases (Arm/Aarch64) for invoking `_mm_cmpestri`?

jserv (changes requested):
Attempt to eliminate Arm/Aarch64 special cases in test suite.

sse2neon.h
```diff
@@ -8358,6 +8359,658 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    _MM_CMPINT_LT = 1,    /* Less than */
+    _MM_CMPINT_LE = 2,    /* Less than or Equal */
+    _MM_CMPINT_FALSE = 3, /* Always False */
+    _MM_CMPINT_NE = 4,    /* Not Equal */
+    _MM_CMPINT_NLT = 5,   /* Not Less than */
+    _MM_CMPINT_NLE = 6,   /* Not Less than or Equal */
+    _MM_CMPINT_TRUE = 7   /* Always True */
+} _MM_CMPINT_ENUM;
+
+static uint16_t mask_for_16bit_elements[8] __attribute__((aligned(16))) = {
```
jserv, line 8373: For the sake of potential symbol conflicts, we might rename to `_sse2neon_cmpestr_mask16b`.

sse2neon.h
```diff
@@ -8358,6 +8359,658 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define __UNIQUE_ID(prefix) __PASTE(__PASTE(__UNIQUE_ID_, prefix), __COUNTER__)
+
+#define __max(t1, t2, max1, max2, x, y) \
+    ({                                  \
+        t1 max1 = (x);                  \
+        t2 max2 = (y);                  \
+        (void) (&max1 == &max2);        \
+        max1 > max2 ? max1 : max2;      \
+    })
+#define max(x, y) \
```
jserv, line 8913: Don't use this! Over-engineering.

sse2neon.h
```diff
@@ -8358,6 +8359,658 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+
+/* specify the output selection in _mm_cmpXstri */
+#define _SIDD_LEAST_SIGNIFICANT 0x00
+#define _SIDD_MOST_SIGNIFICANT 0x40
+
+/* specify the output selection in _mm_cmpXstrm */
+#define _SIDD_BIT_MASK 0x00
+#define _SIDD_UNIT_MASK 0x40
+
+#define PCMPSTR_EQ_16x8(a, b, mtx)                                      \
```
jserv, line 8408: Alternatively, we can make use of [X macro](https://en.wikipedia.org/wiki/X_Macro) technique to simplify the manual loop unrolling and corresponding logic.

tests/impl.cpp
Cuda-Chen: If I simply change the statement from:
```
#if defined(__arm__) || defined(__aarch64__)
    res = _mm_cmpestrm(ma, la, mb, lb, imm8);
#else
    res = _mm_cmpestrm(
        ma, la, mb, lb,
        _SIDD_SWORD_OPS | _SIDD_CMP_RANGES | _SIDD_MOST_SIGNIFICANT);
#endif
```
to:
```
res = _mm_cmpestrm(ma, la, mb, lb, imm8);
```
Just like NEON's `vgetq_lane_xxx` case, I will receive a compilation error when testing on x86 platform:
```
error: the fifth argument must be an 8-bit immediate
 9216 |     res = _mm_cmpestrm(ma, la, mb, lb, imm8);
      |           ^~~~~~~~~~~~
```

I think some kinds of macro usages such as X Macro will help me. Let me come up with the solution.

tests/impl.cpp
jserv: Conditional build with `imm` rather than intrinsic call. i.e., 
```c
#if defined(__arm__) || defined(__aarch64__)
#define IMM (something)
#else
#define IMM (_SIDD_SWORD_OPS | _SIDD_CMP_RANGES | _SIDD_MOST_SIGNIFICANT)
```

sse2neon.h
```diff
@@ -8358,6 +8359,645 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        res ^= (1 << lb) - 1;
+        break;
+    default:
+        break;
+    }
+
+    return res & ((bound == 8) ? 0xFF : 0xFFFF);
+}
+
+#define _sse2neon_max(x, y) (x) > (y) ? (x) : (y)
```
jserv, line 8900: Capitalize the macro.

tests/impl.cpp
```diff
@@ -9149,14 +9149,71 @@ result_t test_mm_cmpestrc(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+    const int imm8;
+    int16_t expect[8];
+} test_mm_cmpestrm_data_t;
+static test_mm_cmpestrm_data_t test_mm_cmpestrm_data = {
+    {13, 6, 5, 4, 3, 2, 1, 3},
+    {-7, 16, 5, 4, -1, 6, 1, 3},
+    10,
+    10,
+    _SIDD_SWORD_OPS | _SIDD_CMP_RANGES | _SIDD_MOST_SIGNIFICANT,
+    {0, 0, 0, 0, 0, 0, -1, -1}};
```
jserv, line 9198: Add a comma between `}` and another `}`, so that `clang-format` can indent better.

jserv:
The article [Pretty-Printable Enumerations in Pure C](https://hirrolot.github.io/posts/pretty-printable-enumerations-in-pure-c.html) shows several C macro tricks, and we might benefit from them.

sse2neon.h
```diff
@@ -8358,6 +8359,645 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+typedef enum {
+    _MM_CMPINT_EQ = 0,    /* Equal */
+    _MM_CMPINT_LT = 1,    /* Less than */
+    _MM_CMPINT_LE = 2,    /* Less than or Equal */
+    _MM_CMPINT_FALSE = 3, /* Always False */
+    _MM_CMPINT_NE = 4,    /* Not Equal */
+    _MM_CMPINT_NLT = 5,   /* Not Less than */
+    _MM_CMPINT_NLE = 6,   /* Not Less than or Equal */
+    _MM_CMPINT_TRUE = 7   /* Always True */
+} _MM_CMPINT_ENUM;
```
jserv, line 8371: Here, typedef is not required.

sse2neon.h
```diff
@@ -8358,6 +8359,645 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+            vshrq_n_u16(vreinterpretq_u16_m128i(mtx[j]), 8));
+        uint16x8_t vec_res = vandq_u16(vreinterpretq_u16_m128i(mtx[j]),
+                                       vreinterpretq_u16_m128i(tmp));
+        int t = vaddvq_u16(vec_res) ? 1 : 0;
+        res |= (t << j);
+    }
+    return res;
+}
+
+static int _sse2neon_cmp_ubyte_ranges(__m128i a, int la, __m128i b, int lb)
```
jserv, line 8712: You can generate these functions by means of macros. See https://github.com/sysprog21/rv32emu/blob/master/main.c#L19 for example.

sse2neon.h
```diff
@@ -8358,6 +8359,645 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
 
 /* SSE4.2 */
 
+enum {
```
jserv, line 8362: The `_MM_CMPINT_ENUM` type used to specify comparison operations in AVX-512 intrinsics. Since we are concentrating on SSE 4.2 intrinsics, we can safely remove this enumeration.

jserv: Check the headers shipped in LLVM and GCC, and you will figure out that the definition appears in `<avx512fintrin.h>`.

sse2neon.h
```diff
@@ -8358,6 +8359,526 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        : [bd] "r"(upper)
+        : "w0", "w1");
+#else
+    int tmp1 = la ^ (la >> 31);
+    la = tmp1 - (la >> 31);
+    int tmp2 = lb ^ (lb >> 31);
+    lb = tmp2 - (lb >> 31);
+    la = SSE2NEON_MAX(la, upper);
+    lb = SSE2NEON_MAX(lb, upper);
+
```
jserv, line 8813: Remove this blank line.

sse2neon.h
```diff
@@ -8358,6 +8359,526 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define RNG_U8(i) \
+    vec_b[i] = vdupq_n_u8(vgetq_lane_u8(vreinterpretq_u8_m128i(b), i));
+#define RNG_S8(i) \
+    vec_b[i] = vdupq_n_s8(vgetq_lane_s8(vreinterpretq_s8_m128i(b), i));
+#define RNG_U16(i) \
+    vec_b[i] = vdupq_n_u16(vgetq_lane_u16(vreinterpretq_u16_m128i(b), i));
+#define RNG_S16(i) \
+    vec_b[i] = vdupq_n_s16(vgetq_lane_s16(vreinterpretq_s16_m128i(b), i));
+
+#define EIGHT(f) f(0) f(1) f(2) f(3) f(4) f(5) f(6) f(7)
```
jserv, line 8414: Use the manner of implementing loops: https://stackoverflow.com/questions/28231743/self-unrolling-macro-loop-in-c-c

jserv (changes requested):
There are few test items. Follow the way how SIMDe provides for SSE 4.2 intrinsics.
See https://github.com/simd-everywhere/simde/blob/master/test/x86/sse4.2.c

tests/impl.cpp
```diff
@@ -9189,33 +9189,66 @@ typedef struct {
...
 
-#if defined(__arm__) || defined(__aarch64__)
-#define IMM (test_mm_cmpestrm_data.imm8)
-#else
-#define IMM (_SIDD_SWORD_OPS | _SIDD_CMP_RANGES | _SIDD_MOST_SIGNIFICANT)
-#endif
 result_t test_mm_cmpestrm(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    int16_t *a = test_mm_cmpestrm_data.a, *b = test_mm_cmpestrm_data.b;
+    /*int16_t *a = test_mm_cmpestrm_data.a, *b = test_mm_cmpestrm_data.b;
```
jserv, line 9240: Remove dead code.

tests/impl.cpp
```diff
@@ -9149,14 +9149,98 @@ result_t test_mm_cmpestrc(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+
+#define SSE2NEON_MM_CMPESTRM_TEST_CASES(f)   \
+    f(SWORD_OPS_CMP_RANGES_MOST_SIGNIFICANT) \
+        f(SWORD_OP_CMP_EQUAL_EACH_UNIT_MASK)
+
+#define SSE2NEON_ENUM_MM_CMPESTRM_TEST_CASES(name, list)               \
+    typedef enum { list(SSE2NEON_DEF_ENUM_MM_CMPESTRM_VARIANT) } name; \
+    list(SSE2NEON_EVAL_MM_CMPESTRM_TEST_CASE)
+
+#define SSE2NEON_DEF_ENUM_MM_CMPESTRM_VARIANT(c) c,
```
jserv, line 9221: The macros appear in a C source file rather than a public header, meaning that the identifier conflicts should be much fewer. You can safely use shorter identifiers.

sse2neon.h
```diff
@@ -8358,6 +8359,526 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
 
 /* SSE4.2 */
 
+static uint16_t _sse2neon_cmpestr_mask16b[8] __attribute__((aligned(16))) = {
```
jserv, line 8362: Use macro `ALIGN_STRUCT` instead.

sse2neon.h
```diff
@@ -8358,6 +8359,526 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define RNG_U8(i) \
+    vec_b[i] = vdupq_n_u8(vgetq_lane_u8(vreinterpretq_u8_m128i(b), i));
+#define RNG_S8(i) \
+    vec_b[i] = vdupq_n_s8(vgetq_lane_s8(vreinterpretq_s8_m128i(b), i));
+#define RNG_U16(i) \
+    vec_b[i] = vdupq_n_u16(vgetq_lane_u16(vreinterpretq_u16_m128i(b), i));
+#define RNG_S16(i) \
+    vec_b[i] = vdupq_n_s16(vgetq_lane_s16(vreinterpretq_s16_m128i(b), i));
+
+#define EIGHT(f, i) \
```
jserv, line 8414: It does not look elegant. Can you check this? https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms

Cuda-Chen: >It does not look elegant. Can you check this?
Hi @jserv , you mean the `RNG_S16` or `EIGHT` macro?

tests/impl.cpp
```diff
@@ -9149,14 +9149,287 @@ result_t test_mm_cmpestrc(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+        int8_t val[16];                                                       \
+        memcpy(val, &ma, sizeof(val));                                        \
+        printf("sbyte a %d ", c);                                             \
+        printf(                                                               \
+            "Numerical: %d %d %d %d %d %d %d %d %d %d %d %d %d %d %d %d \n",  \
+            val[0], val[1], val[2], val[3], val[4], val[5], val[6], val[7],   \
+            val[8], val[9], val[10], val[11], val[12], val[13], val[14],      \
+            val[15]);                                                         \
+        memcpy(val, &mb, sizeof(val));                                        \
+        printf("sbyte b %d ", c);                                             \
```
jserv, line 9388: Is `printf` necessary?

Cuda-Chen: > Is `printf` necessary?

This `printf` is used for solely debug purpose. I will remove it as soon as I finish the test cases verification.

sse2neon.h
```diff
@@ -8358,6 +8359,530 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
 
 /* SSE4.2 */
 
+static uint16_t _sse2neon_cmpestr_mask16b[8] ALIGN_STRUCT(16) = {
```
jserv, line 8362: Add `const` modifier to aid compiler optimizations.

sse2neon.h
```diff
@@ -8358,6 +8359,530 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80,
+};
+static uint8_t _sse2neon_cmpestr_mask8b[16] ALIGN_STRUCT(16) = {
```
jserv, line 8365: Ditto.

sse2neon.h
```diff
@@ -8358,6 +8359,530 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define EQ_U8(i) EQ(u8, i)
+
+#define EQ_U16(i) EQ(u16, i)
+
+#define EQ(type, i)                                                            \
+    mtx[i] = vreinterpretq_m128i_##type(vceqq_##type(                          \
+        vdupq_n_##type(vgetq_lane_##type(vreinterpretq_##type##_m128i(b), i)), \
+        vreinterpretq_##type##_m128i(a)));
+
+#define RNG_U8(i) RNG(u8, i)
```
jserv, line 8406: If you change the naming scheme to `RNG_u8`, `RNG_s8`, `RNG_u16 and `RNG_s16`, [concatenation](https://gcc.gnu.org/onlinedocs/cpp/Concatenation.html) can be used to shorten the implementation.

sse2neon.h
```diff
@@ -8358,6 +8359,530 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+// in imm8, and store the generated index in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpestri
+FORCE_INLINE int _mm_cmpestri(__m128i a,
+                              int la,
+                              __m128i b,
+                              int lb,
+                              const int imm8)
+{
+    int upper = (imm8 & 0x01) ? 8 : 16;
+#if defined(__aarch64__)
```
jserv, line 8798: Can you check if generic C code benefits from compiler optimizations (at least `-O2`) which generate instructions better than hand-written inline assembly?

sse2neon.h
```diff
@@ -8358,6 +8359,530 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    for (int j = lb; j < bound; j++) {
+        mtx[j] = vreinterpretq_m128i_u8(vbslq_u8(vec1, vec_minusone, vec_zero));
+    }
+    unsigned char *ptr = (unsigned char *) mtx;
+    for (int i = 0; i < bound; i++) {
+        int val = 1;
+        for (int j = 0, k = i; j < bound - i && k < bound; j++, k++) {
+            val &= ptr[k * bound + j];
+        }
+        res = (val << i) + res;
```
jserv, line 8696: Rewrite as
```c
res += val << i;
```

sse2neon.h
jserv: I meant `EIGHT`, which might cause potential re-definition.

sse2neon.h
Cuda-Chen: Got it. I will post the assembly in this conversation.

sse2neon.h
```diff
@@ -8358,6 +8359,530 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+                                            int lb)
+{
+    __m128i mtx[16];
+    PCMPSTR_EQ_16x8(a, b, mtx);
+    return aggregate_equal_ordered_16x8(8, la, lb, mtx);
+}
+
+typedef int (*cmpestr_func_t)(__m128i a, int la, __m128i b, int lb);
+
+static cmpestr_func_t _sse2neon_cmpfunc_table[] = {
```
jserv, line 8752: Embrace C99 array initialization. i.e.,
```c
static const cmpestr_func_t _sse2neon_cmpfunc_table[] = {
    [CMP_UBYTE_EQUAL_ANY] = _sse2neon_cmp_byte_equal_any,
    [CMP_UWORD_EQUAL_ANY] = _sse2neon_cmp_word_equal_any,
    ...
};

```

sse2neon.h
jserv: Good. The original inline assembly listing can be replaced with generic C implementation once we confirm the effectiveness of compiler optimizations.

tests/impl.cpp
```diff
@@ -9149,14 +9149,643 @@ result_t test_mm_cmpestrc(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+    typedef enum { list(DEF_ENUM_MM_CMPESTRI_VARIANT) } name; \
+    list(EVAL_MM_CMPESTRI_UWORD_TEST_CASE)
+
+#define ENUM_MM_CMPESTRI_SWORD_TEST_CASES(name, list)         \
+    typedef enum { list(DEF_ENUM_MM_CMPESTRI_VARIANT) } name; \
+    list(EVAL_MM_CMPESTRI_SWORD_TEST_CASE)
+
+#define DEF_ENUM_MM_CMPESTRI_VARIANT(c) c,
+
+#define EVAL_MM_CMPESTRI_UBYTE_TEST_CASE(c)                 \
```
jserv, line 9373: Introduce "generator" macros to simplify tests.

sse2neon.h
jserv: By the way, [X macro](https://en.wikipedia.org/wiki/X_Macro) is a useful technique. Check https://github.com/sysprog21/rv32emu/blob/master/src/syscall.c and look for `__SYSCALL_LIST_BASE` for details.

sse2neon.h
Cuda-Chen: After the experiment on Compiler Explorer, in brief, generic C implementation outputs less assembly codes compared to hand-written inline assembly with `-O2` compiler optimization flag (there are no any differences when applying `-O3` flag, so I will  only post the generated assembly with `-O2` flag).

The test program is written as follows:

<table>
<tr>
<td> hand-written inline assembly </td> <td> generic C implementation </td>
</tr>
<tr>
<td>

```c
#include <stdio.h>

#define SSE2NEON_MAX(x, y) (x) > (y) ? (x) : (y)

int main() {
    int la = 5, lb = 10, upper = 16;
    __asm__ __volatile__(
        "eor w0, %w[a], %w[a], asr #31\n"
        "sub %w[a], w0, %w[a], asr #31\n"
        "eor w1, %w[b], %w[b], asr #31\n"
        "sub %w[b], w1, %w[b], asr #31\n"
        "cmp %w[a], %w[bd]\n"
        "csel %w[a], %w[bd], %w[a], gt\n"
        "cmp %w[b], %w[bd]\n"
        "csel %w[b], %w[bd], %w[b], gt\n"
        : [a] "+r"(la), [b] "+r"(lb)
        : [bd] "r"(upper)
        : "w0", "w1");

    printf("%d %d", la, lb);

    return 0;
}
```

</td>
<td>

```c
#include <stdio.h>

#define SSE2NEON_MAX(x, y) (x) > (y) ? (x) : (y)

int main() {
    int la = 5, lb = 10, upper = 16;
    int tmp1 = la ^ (la >> 31);
    la = tmp1 - (la >> 31);
    int tmp2 = lb ^ (lb >> 31);
    lb = tmp2 - (lb >> 31);
    la = SSE2NEON_MAX(la, upper);
    lb = SSE2NEON_MAX(lb, upper);

    printf("%d %d", la, lb);

    return 0;
}
```

</td>
</tr>
</table>

And the results of GCC 10.2 (Compiler Explorer does not provide GCC 10, so I use GCC 10.2 instead):
<table>
<tr>
<td> hand-written inline assembly </td> <td> generic C implementation </td>
</tr>
<tr>
<td>

```assembly
.LC0:
        .string "%d %d"
main:
        stp     x29, x30, [sp, -16]!
        mov     w3, 5
        mov     w2, 10
        mov     x29, sp
        mov     w4, 16
        eor w0, w3, w3, asr #31
sub w3, w0, w3, asr #31
eor w1, w2, w2, asr #31
sub w2, w1, w2, asr #31
cmp w3, w4
csel w3, w4, w3, gt
cmp w2, w4
csel w2, w4, w2, gt

        adrp    x0, .LC0
        mov     w1, w3
        add     x0, x0, :lo12:.LC0
        bl      printf
        mov     w0, 0
        ldp     x29, x30, [sp], 16
        ret
```

</td>
<td>

```assembly
.LC0:
        .string "%d %d"
main:
        stp     x29, x30, [sp, -16]!
        mov     w2, 16
        adrp    x0, .LC0
        mov     x29, sp
        mov     w1, w2
        add     x0, x0, :lo12:.LC0
        bl      printf
        mov     w0, 0
        ldp     x29, x30, [sp], 16
        ret
```

</td>
</tr>
</table>

Also the results of Clang 11.0.0:
<table>
<tr>
<td> hand-written inline assembly </td> <td> generic C implementation </td>
</tr>
<tr>
<td>

```assembly
main:                                   // @main
        stp     x29, x30, [sp, #-16]!           // 16-byte Folded Spill
        mov     x29, sp
        mov     w8, #5
        mov     w2, #10
        mov     w9, #16
        eor     w0, w8, w8, asr #31
        sub     w8, w0, w8, asr #31
        eor     w1, w2, w2, asr #31
        sub     w2, w1, w2, asr #31
        cmp     w8, w9
        csel    w8, w9, w8, gt
        cmp     w2, w9
        csel    w2, w9, w2, gt

        adrp    x0, .L.str
        add     x0, x0, :lo12:.L.str
        mov     w1, w8
        bl      printf
        mov     w0, wzr
        ldp     x29, x30, [sp], #16             // 16-byte Folded Reload
        ret
.L.str:
        .asciz  "%d %d"
```

</td>
<td>

```assembly
main:                                   // @main
        stp     x29, x30, [sp, #-16]!           // 16-byte Folded Spill
        mov     x29, sp
        adrp    x0, .L.str
        add     x0, x0, :lo12:.L.str
        mov     w1, #16
        mov     w2, #16
        bl      printf
        mov     w0, wzr
        ldp     x29, x30, [sp], #16             // 16-byte Folded Reload
        ret
.L.str:
        .asciz  "%d %d"
```

</td>
</tr>
</table>

jserv: It looks promising. Let's drop the use of inline assembly, which does not outperform the generic C code.

Cuda-Chen: After testing, I found the generic C implementation will fail on `EQUAL_ORDERED` test cases. I will investigate the root cause.

sse2neon.h
```diff
@@ -8795,7 +8795,7 @@ FORCE_INLINE int _mm_cmpestri(__m128i a,
                               const int imm8)
 {
     int upper = (imm8 & 0x01) ? 8 : 16;
-#if defined(__aarch64__)
+#if 0
```
jserv, line 8798: Remove the dead code.

sse2neon.h
Cuda-Chen: So it seems I misinterpret these lines to `MAX`:
```assembly
"cmp %w[a], %w[bd]\n"
"csel %w[a], %w[bd], %w[a], gt\n"
```

It should be `MIN`, though.

Cuda-Chen: After testing, the generic C implementation still gives promising results with `-O2` flag.

Testing program:

```c
#include <stdio.h>

#define SSE2NEON_MIN(x, y) (x) < (y) ? (x) : (y)

int main() {
    int la = 5, lb = 10, upper = 16;
    int tmp1 = la ^ (la >> 31);
    la = tmp1 - (la >> 31);
    int tmp2 = lb ^ (lb >> 31);
    lb = tmp2 - (lb >> 31);
    la = SSE2NEON_MIN(la, upper);
    lb = SSE2NEON_MIN(lb, upper);

    printf("%d %d", la, lb);

    return 0;
}
```

And the assembly generated by each compiler:

<table>
<tr>
<td> ARM64 GCC 10.2 </td> <td> armv8-a clang 11.0.0 </td>
</tr>

<tr>
<td>

```asm
.LC0:
        .string "%d %d"
main:
        stp     x29, x30, [sp, -16]!
        mov     w2, 10
        mov     w1, 5
        mov     x29, sp
        adrp    x0, .LC0
        add     x0, x0, :lo12:.LC0
        bl      printf
        mov     w0, 0
        ldp     x29, x30, [sp], 16
        ret
```

</td>
<td>

```asm
main:                                   // @main
        stp     x29, x30, [sp, #-16]!           // 16-byte Folded Spill
        mov     x29, sp
        adrp    x0, .L.str
        add     x0, x0, :lo12:.L.str
        mov     w1, #5
        mov     w2, #10
        bl      printf
        mov     w0, wzr
        ldp     x29, x30, [sp], #16             // 16-byte Folded Reload
        ret
.L.str:
        .asciz  "%d %d"
```

</td>
</tr>
</table>

sse2neon.h
```diff
@@ -8358,6 +8359,512 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+
+/* specify the output selection in _mm_cmpXstrm */
+#define _SIDD_BIT_MASK 0x00
+#define _SIDD_UNIT_MASK 0x40
+
+#define EQ_U8(i) EQ(u8, i)
+
+#define EQ_U16(i) EQ(u16, i)
+
+#define EQ(type, i)                                                            \
```
jserv, line 8401: `EQ` might cause potential conflicts with other macros. Use [push/pop_macro](https://gcc.gnu.org/onlinedocs/gcc/Push_002fPop-Macro-Pragmas.html) for gcc.

sse2neon.h
```diff
@@ -8358,6 +8359,512 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+                                            int la,
+                                            __m128i b,
+                                            int lb)
+{
+    __m128i mtx[16];
+    PCMPSTR_EQ_16x8(a, b, mtx);
+    return aggregate_equal_ordered_16x8(8, la, lb, mtx);
+}
+
+#define SSE2NEON_ENUM_ENUM_MM_CMPESTRX                     \
```
jserv, line 8750: Rename this macro to `SSE2NEON_CMPESTR_LIST`.

sse2neon.h
```diff
@@ -8358,6 +8359,512 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    for (int i = 0; i < bound; i++) {
+        int val = 1;
+        for (int j = 0, k = i; j < bound - i && k < bound; j++, k++)
+            val &= ptr[k * bound + j];
+        res = (val << i) + res;
+    }
+    return res;
+}
+
+static int _sse2neon_cmp_byte_equal_ordered(__m128i a,
```
jserv, line 8730: Both `_sse2neon_cmp_byte_equal_ordered` and `_sse2neon_cmp_word_equal_ordered` can be generated via macro.

sse2neon.h
```diff
@@ -8358,6 +8359,512 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    uint16x8_t vec0 = vtstq_u16(vdupq_n_u16(m0), vec_mask);
+    uint16x8_t vec1 = vtstq_u16(vdupq_n_u16(m1), vec_mask);
+    uint16x8_t tmp = vtstq_u16(vdupq_n_u16(tb), vec_mask);
+    mtx = vbslq_u16(vec0, vdupq_n_u16(0), mtx);
+    mtx = vbslq_u16(vec1, tmp, mtx);
+    mtx = vandq_u16(mtx, vec_mask);
+    return vaddvq_u16(mtx);
+}
+
+static int aggregate_equal_ordered_8x16(int bound,
```
jserv, line 8671: Both `aggregate_equal_ordered_8x16` and `aggregate_equal_ordered_16x8` have the common part. Let's use macro to generate their implementations.

sse2neon.h
jserv: Can you avoid using macro `EIGHT`?

Cuda-Chen: Of course, I will give avoid using macro `EIGHT` a shot.

sse2neon.h
```diff
@@ -8482,12 +8482,12 @@ const static uint8_t _sse2neon_cmpestr_mask8b[16] ALIGN_STRUCT(16) = {
 
 #define PCMPSTR_EQ_16x8(a, b, mtx) \
```
jserv, line 8483: You can rename `EIGHT` to `x8` and `SIXTEEN` to `x16`. Later, make use of catenate trick. See https://github.com/sysprog21/rv32emu/blob/master/src/common.h

sse2neon.h
```diff
@@ -8358,6 +8359,495 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define x8(f, i) \
+    f(i + 0) f(i + 1) f(i + 2) f(i + 3) f(i + 4) f(i + 5) f(i + 6) f(i + 7)
+
+#define x16(f, i) x8(f, i) x8(f, i + 8)
+
+#define PCMPSTR_RNG_U_8x16(a, b, mtx)                                     \
+    do {                                                                  \
+        uint8x16_t vec_b[16];                                             \
+        __m128i mask = vreinterpretq_m128i_u16(vdupq_n_u16(0xff));        \
+        x16(RNG_u8, 0) for (int i = 0; i < 16; i++)                       \
```
jserv, line 8440: If you pass types such as `u8` to macro `x16`, it is feasible to avoid defining macros like `RNG_u8`.

sse2neon.h
```diff
@@ -8358,6 +8359,495 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define EQ_u8(i) EQ(u8, i)
+
+#define EQ_u16(i) EQ(u16, i)
+
+#define EQ(type, i)                                                            \
+    mtx[i] = vreinterpretq_m128i_##type(vceqq_##type(                          \
+        vdupq_n_##type(vgetq_lane_##type(vreinterpretq_##type##_m128i(b), i)), \
+        vreinterpretq_##type##_m128i(a)));
+
+#define RNG_u8(i) RNG(u8, i)
```
jserv, line 8419: Ditto. Think of the feasibility not to define macro `RNG_u8`, `RNG_s8`, `RNG_u16` and `RNG_s16`.

sse2neon.h
```diff
@@ -8358,6 +8359,495 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define _SIDD_UNIT_MASK 0x40
+
+/* Pattern Matching for C macros.
+ * https://github.com/pfultz2/Cloak/wiki/C-Preprocessor-tricks,-tips,-and-idioms
+ */
+
+/* catenate */
+#define PRIMITIVE_CAT(a, ...) a##__VA_ARGS__
+
+#define IIF(c) PRIMITIVE_CAT(IIF_, c)
```
jserv, line 8404: To avoid potential macro naming conflicts, add prefix `SSE2NEON_` to these macros.

sse2neon.h
```diff
@@ -8358,6 +8359,495 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+
+#define RNG_u16(i) RNG(u16, i)
+
+#define RNG_s16(i) RNG(s16, i)
+
+#define RNG(type, i) \
+    vec_b[i] =       \
+        vdupq_n_##type(vgetq_lane_##type(vreinterpretq_##type##_m128i(b), i));
+
+#define x8(f, i) \
```
jserv, line 8431: For both `x8` and `x16`, you shall add prefix `SSE2NEON_` to avoid potential naming conflicts.

sse2neon.h
```diff
@@ -8358,6 +8359,495 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#undef _
+};
+typedef int (*cmpestr_func_t)(__m128i a, int la, __m128i b, int lb);
+static cmpestr_func_t _sse2neon_cmpfunc_table[] = {
+#define _(name, number, func_suffix) [name] = _sse2neon_##func_suffix,
+    SSE2NEON_CMPESTR_LIST
+#undef _
+};
+
+static inline int neg_fun(int res, int lb, int imm8, int bound)
```
jserv, line 8763: Let's rename this function to avoid potential conflicts. How about `_sse2neon_sido_negative`?

sse2neon.h
```diff
@@ -8358,6 +8359,495 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+/* catenate */
+#define SSE2NEON_PRIMITIVE_CAT(a, ...) a##__VA_ARGS__
+
+#define SSE2NEON_IIF(c) SSE2NEON_PRIMITIVE_CAT(SSE2NEON_IIF_, c)
+/* run the 2nd parameter */
+#define SSE2NEON_IIF_0(t, ...) __VA_ARGS__
+/* run the 1st parameter */
+#define SSE2NEON_IIF_1(t, ...) t
+
+#define EQ_u8(i) EQ(u8, i)
```
jserv, line 8410: Can you avoid defining `EQ` and its variants?

sse2neon.h
```diff
@@ -8358,6 +8359,549 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    uint16x8_t vec0 = vtstq_u16(vdupq_n_u16(m0), vec_mask);
+    uint16x8_t vec1 = vtstq_u16(vdupq_n_u16(m1), vec_mask);
+    uint16x8_t tmp = vtstq_u16(vdupq_n_u16(tb), vec_mask);
+    mtx = vbslq_u16(vec0, vdupq_n_u16(0), mtx);
+    mtx = vbslq_u16(vec1, tmp, mtx);
+    mtx = vandq_u16(mtx, vec_mask);
+    return vaddvq_u16(mtx);
+}
+
+#define UBYTE 1
```
jserv, line 8741: Can you avoid defining macro `UBYTE` and `UWORD`? They might cause potential conflicts.

sse2neon.h
```diff
@@ -8358,6 +8359,549 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+}
+
+static int _sse2neon_cmp_word_equal_any(__m128i a, int la, __m128i b, int lb)
+{
+    __m128i mtx[16];
+    PCMPSTR_EQ(a, b, mtx, 16, 8);
+    return aggregate_equal_any_16x8(la, lb, mtx);
+}
+
+static int aggregate_ranges_16x8(int la, int lb, __m128i mtx[16])
```
jserv, line 8642: Both `aggregate_ranges_16x8` and `aggregate_ranges_8x16` should have prefix `_sse2neon_`.

sse2neon.h
```diff
@@ -8358,6 +8359,549 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define SSE2NEON_WHEN(c) SSE2NEON_IF(c)(SSE2NEON_EXPAND, SSE2NEON_EAT)
+
+/* recursion */
+/* deferred expression */
+#define SSE2NEON_EMPTY()
+#define SSE2NEON_DEFER(id) id SSE2NEON_EMPTY()
+#define SSE2NEON_OBSTRUCT(...) __VA_ARGS__ SSE2NEON_DEFER(SSE2NEON_EMPTY)()
+#define SSE2NEON_EXPAND(...) __VA_ARGS__
+
+#define SSE2NEON_EVAL(...) \
```
jserv, line 8456: Can you use fewer macro `SSE2NEON_EVAL` variants? That is, drop `SSE2NEON_EVAL5` or something else.

sse2neon.h
```diff
@@ -8358,6 +8359,549 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+
+#define SSE2NEON_REPEAT(count, macro, ...)                         \
+    SSE2NEON_WHEN(count)                                           \
+    (SSE2NEON_OBSTRUCT(SSE2NEON_REPEAT_INDIRECT)()(                \
+        SSE2NEON_DEC(count), macro,                                \
+        __VA_ARGS__) SSE2NEON_OBSTRUCT(macro)(SSE2NEON_DEC(count), \
+                                              __VA_ARGS__))
+#define SSE2NEON_REPEAT_INDIRECT() SSE2NEON_REPEAT
+
+#define EQ(i, type)                                                            \
```
jserv, line 8476: Both `EQ` and `RNG` might conflict with existing macros.

sse2neon.h
```diff
@@ -8358,6 +8359,549 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+#define SSE2NEON_IIF_0(t, ...) __VA_ARGS__
+/* run the 1st parameter */
+#define SSE2NEON_IIF_1(t, ...) t
+
+#define SSE2NEON_COMPL(b) SSE2NEON_PRIMITIVE_CAT(SSE2NEON_COMPL_, b)
+#define SSE2NEON_COMPL_0 1
+#define SSE2NEON_COMPL_1 0
+
+#define SSE2NEON_DEC(x) SSE2NEON_PRIMITIVE_CAT(SSE2NEON_DEC_, x)
+#define SSE2NEON_DEC_0 0
```
jserv, line 8415: Since we only iterate with certain counts (`8` and `16`), we don't have to rely on the generic macro `SSE2NEON_DEC_0`. Can you write down the specialized ones instead?

sse2neon.h
```diff
@@ -8358,6 +8359,576 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        __VA_ARGS__) SSE2NEON_OBSTRUCT(macro)(SSE2NEON_DEC(count), \
+                                              __VA_ARGS__))
+#define SSE2NEON_REPEAT_INDIRECT() SSE2NEON_REPEAT
+
+#define COMPARE_EQUAL_THEN_FILL_LANE(i, type)                                  \
+    mtx[i] = vreinterpretq_m128i_##type(vceqq_##type(                          \
+        vdupq_n_##type(vgetq_lane_##type(vreinterpretq_##type##_m128i(b), i)), \
+        vreinterpretq_##type##_m128i(a)));
+
+#define FILL_LANE(i, type) \
```
jserv, line 8475: Add prefix `SSE2NEON_`.

sse2neon.h
```diff
@@ -8358,6 +8359,576 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    vec_b[i] =             \
+        vdupq_n_##type(vgetq_lane_##type(vreinterpretq_##type##_m128i(b), i));
+
+#define SSE2NEON_x8(f, i) \
+    f(i + 0) f(i + 1) f(i + 2) f(i + 3) f(i + 4) f(i + 5) f(i + 6) f(i + 7)
+
+#define SSE2NEON_x16(f, i) SSE2NEON_x8(f, i) SSE2NEON_x8(f, i + 8)
+
+/*#define PCMPSTR_RNG_u8x16(a, b, mtx)                                      \
```
jserv, line 8484: Don't leave the dead code.

sse2neon.h
```diff
@@ -8358,6 +8359,576 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+               : ((imm8 & 0x40) ? (31 - __builtin_clz(r2)) : __builtin_ctz(r2));
+}
+
+// Compare packed strings in a and b with lengths la and lb using the control
+// in imm8, and store the generated mask in dst.
+// https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmpestrm
+FORCE_INLINE __m128i
+_mm_cmpestrm(__m128i a, int la, __m128i b, int lb, const int imm8)
+{
+    int bound = (imm8 & 0x01) ? 8 : 16;
```
jserv, line 8893: Add `const` modifier to aid compiler optimization.

sse2neon.h
```diff
@@ -8358,6 +8359,518 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    }
+
+SSE2NEON_AGGREGATE_EQUAL_ORDER_IMPL(8,
+                                    16,
+                                    SSE2NEON_AGGREGATE_EQUAL_ORDER_IS_UBYTE)
+SSE2NEON_AGGREGATE_EQUAL_ORDER_IMPL(16,
+                                    8,
+                                    SSE2NEON_AGGREGATE_EQUAL_ORDER_IS_UWORD)
+
+#undef UBYTE
```
jserv, line 8753: Can you confirm the use of `#undef`? Right identifiers?

jserv (changes requested):
Improve git commit messages. See https://cbea.ms/git-commit/

You should summarize the changes in the body of git commit message.

sse2neon.h
```diff
@@ -8358,6 +8359,518 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        int t = vaddvq_u16(vec_res) ? 1 : 0;
+        res |= (t << j);
+    }
+    return res;
+}
+
+#define SSE2NEON_CMP_RANGES_IS_BYTE 1
+#define SSE2NEON_CMP_RANGES_IS_WORD 0
+
+SSE2NEON_CMP_RANGES_IMPL(ubyte, uint, u, 8, 16, SSE2NEON_CMP_RANGES_IS_BYTE)
```
jserv, line 8656: Utilize concatenation in the C Preprocessor to shorten. That is,
```
SSE2NEON_CMP_RANGES_IMPL(ubyte, uint, u, BYTE)
```

Both "8" and "16" can be obtained from `sizeof` operator.

Cuda-Chen: >Both "8" and "16" can be obtained from sizeof operator.

Currently I can only come up with the following solution:

```c
#define SSE2NEON_CMP_RANGES_IMPL(type, data_type, us, byte_word) \
    int size = sizeof(ubyte); \ // get the size of vector's element
    int number_of_lane = 128 / size; \ // get the size of vector
    PCMPSTR_RNG(a, b, mtx, data_type, us, size, number_of_lanes,         \
                    byte_or_word);

SSE2NEON_CMP_RANGES_IMPL(ubyte, uint, u, BYTE)
```

So maybe some kind of macro definition (e.g., `ubyte`) can achieve the goal?

jserv: Sort of. We should minimize the count of parameters.

sse2neon.h
```diff
@@ -8358,6 +8359,518 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+        for (int i = 0; i < bound; i++) {                                      \
+            int val = 1;                                                       \
+            for (int j = 0, k = i; j < bound - i && k < bound; j++, k++)       \
+                val &= ptr[k * bound + j];                                     \
+            res += val << i;                                                   \
+        }                                                                      \
+        return res;                                                            \
+    }
+
+SSE2NEON_AGGREGATE_EQUAL_ORDER_IMPL(8,
```
jserv, line 8746: Instead of specifying `SSE2NEON_AGGREGATE_EQUAL_ORDER_IS_UBYTE` and `SSE2NEON_AGGREGATE_EQUAL_ORDER_IS_UWORD`, utilize catenation of C preprocessor to shorten.

sse2neon.h
```diff
@@ -8358,6 +8359,518 @@ FORCE_INLINE int _mm_testz_si128(__m128i a, __m128i b)
...
+    _(CMP_UWORD_EQUAL_EACH, 9, cmp_word_equal_each)        \
+    _(CMP_SBYTE_EQUAL_EACH, 10, cmp_byte_equal_each)       \
+    _(CMP_SWORD_EQUAL_EACH, 11, cmp_word_equal_each)       \
+    _(CMP_UBYTE_EQUAL_ORDERED, 12, cmp_byte_equal_ordered) \
+    _(CMP_UWORD_EQUAL_ORDERED, 13, cmp_word_equal_ordered) \
+    _(CMP_SBYTE_EQUAL_ORDERED, 14, cmp_byte_equal_ordered) \
+    _(CMP_SWORD_EQUAL_ORDERED, 15, cmp_word_equal_ordered)
+
+enum {
+#define _(name, number, func_suffix) name = number,
```
jserv, line 8778: Using `enum`, it is necessary to specify each number since it counts from zero. Therefore, we can skip the number 0, 1, 2, ... , 15.

jserv (changes requested):
Squash the commits for reviewing.

jserv:

PR #535, opened 2022-09-26
gcc-12 still does not support vld1q_u8_x4 for 32 arm builds.

balister:
Signed-off-by: Philip Balister <philip@balister.org>


jserv:
Thank @balister for contributing! If you have more successful stories for adopting SSE2NEON in GNU Radio and/or OpenSDR, please inform us.

PR #541, opened 2022-10-18
Exclude _mm_malloc/free definitions on Windows

invertego:
Definitions for these functions are provided by malloc.h from both MinGW-w64 and MSVC. When they are defined again by sse2neon, it causes multiple definition errors. Furthermore, the implementation in sse2neon relies on posix_memalign which is unavailable on Windows.


sse2neon.h
```diff
@@ -1773,10 +1773,12 @@ FORCE_INLINE __m128 _mm_div_ss(__m128 a, __m128 b)
 
 // Free aligned memory that was allocated with _mm_malloc.
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_free
+#if !defined(_WIN32)
```
jserv, line 1776: Can you define `SSE2NEON_ALLOC_DEFINED` right after `#include <stdlib.h>`? That is, I expect the following:
```c
#include <stdint.h>
#include <stdlib.h>

#if defined(_WIN32) || defined(_WIN64)
/* Definitions for _mm_{malloc,free} are provided by <malloc.h>
 * from both MinGW-w64 and MSVC.
 */
#define SSE2NEON_ALLOC_DEFINED
#endif
...
#if !defined(SSE2NEON_ALLOC_DEFINED)
FORCE_INLINE void _mm_free(void *addr)
{
    free(addr);
}
#endif


```

jserv:
By the way, are you able to build SSE2NEON with MSVC?

sse2neon.h
invertego: Done with latest push. I only tested for _WIN32 because it is defined for all architectures.

invertego:
> By the way, are you able to build SSE2NEON with MSVC?

I've started looking at it. If I complete the work, I'll send a separate PR.

jserv:

PR #542, opened 2022-10-18
Remove designated initialization of an array

invertego:
C supports designated initialization of arrays, but C++ does not. The one usage of it does not serve a functional purpose and so can be removed. This code was triggering warnings under clang and errors under MSVC.


invertego:
There are numerous other changes required for a clean MSVC build. This PR is specifically addressing warnings with clang.

Cuda-Chen (approved):

jserv:

PR #484, opened 2021-09-13
fix: Rewrite implementation and testing of shift intrinsics

marktwtn:
Implementation:
Rewrite conditional statement with bitwise AND operation.
Fix _mm_slli_si128() and _mm_srli_si128() intrinsics
and rewrite as functions.
Remove extra conditional statements for simplicity.

Testing:
Rewrite conditional statement with bitwise AND operation.
Test shift value in a reasonable range.
Test shift intrinsics required immediate shift value with different
input data.
Make the testing code follow the sequence order as variables declaration,
C implementation, SIMD implmentation and validation.

Close #393.

PR #543, opened 2022-10-20
Reintroduce `ext`-based implementations for shift intrinsics

AymenQ:
Use the vector extract instruction (ext) to implement `_mm_srli_si128`,
`_mm_slli_si128` and `_mm_alignr_epi8`, significantly improving performance
by avoiding memory.

These were originally changed in #483 and #484 to instead use a store
followed by a shifted load. This was done in order to resolve issue
#482, where the compiler would throw an error if there were an invalid
immediate in the intrinsic arguments after the macros are expanded.

These commits avoid the recurrence of this issue by guarding the
immediate arguments with additional (redundant) ternary expressions.
These are evaluated at compile time and should cause no performance
loss; they only prevent the compilers from attempting to propagate
invalid immediates. This does now require that the immediate argument
to these intrinsics be a compile-time constant expression, but this
matches the requirement imposed by the original SSE intrinsics.

Sample codegen diffs are included in the individual commit messages.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -5600,14 +5600,18 @@ FORCE_INLINE __m128i _mm_slli_epi64(__m128i a, int imm)
 //   dst[127:0] := a[127:0] << (tmp*8)
 //
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_slli_si128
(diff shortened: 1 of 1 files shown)


jserv:
Thank @AymenQ for contributing! I appreciate the efforts from Arm.

PR #545, opened 2022-10-21
Support `__builtin_shuffle` as an alternative to `__builtin_shufflevector`

AymenQ:
GCC versions < 12 do not support `__builtin_shufflevector`, though they do
support an alternative builtin named `__builtin_shuffle`. It requires that
the indices are passed as an integer vector as opposed to variadic
arguments.

This enables us to benefit from potentially improved and better
performing code generation for intrinsics that are implemented with this
builtin.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -174,6 +174,14 @@
 #define __has_builtin(x) HAS##x
 #define HAS__builtin_popcount 1
 #define HAS__builtin_popcountll 1
+
+// __builtin_shuffle introduced in GCC 4.7.0
+#if (__GNUC__ >= 5) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 7))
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -190,6 +198,26 @@
 #define _MM_SHUFFLE(fp3, fp2, fp1, fp0) \
     (((fp3) << 6) | ((fp2) << 4) | ((fp1) << 2) | ((fp0)))
 
+#if __has_builtin(__builtin_shufflevector)
+#define sse2neon_shuffle(type, a, b, ...) \
```
jserv, line 202: For consistency, use prefix `_sse2neon_` instead of `sse2neon_`. i.e.,
```c
#define _sse2neon_shuffle(type, a, b, ...) \
```

PR #546, opened 2022-10-21
Improve performance of float-to-integer intrinsics

AymenQ:
Improve performance of several of the float-to-integer intrinsics.

The first commit re-writes `_mm_cvtps_pi8` and `_mm_cvtps_pi16` to use
the signed saturating extract-narrow (`sqxtn`) intrinsic instead of
emulating the saturating-narrow behaviour with several masks.

The second commit adds support for using the round-to-integral
intrinsics introduced by `FEAT_FRINTTS` in Armv8.5-A. These match the
out-of-bounds behaviour specified by the SSE float-to-integer
conversions instead of saturating, so this required a test modification.

Sample codegen diffs are provided in the individual commit messages.


jserv:
Thank @AymenQ for contributing!

PR #550, opened 2022-10-30
fix: Align _mm_prefetch behavior to document

### LINKED ISSUE #547: Consolidate `_mm_prefetch`
Current `_mm_prefetch` does not behave as [Intel documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_prefetch) states:
> Fetch the line of data from memory that contains address `p` to a location in the cache heirarchy specified by the locality hint `i`.

We shall consolidate:
1. Refine the function prototype. i.e., `void _mm_prefetch(char const *p, int i)`
2. Provide the corresponding test cases. See [test/x86/sse.c](https://github.com/simd-everywhere/simde/blob/master/test/x86/sse.c) (Function `test_simde_mm_prefetch`)
3. Properly manipulate the locality hint.

The implementation from [SIMDe](https://github.com/simd-everywhere/simde/blob/master/simde/x86/sse.h):
```c
void simde_mm_prefetch (const void* p, int i) {
    switch(i) {
      case SIMDE_MM_HINT_NTA:
        __builtin_prefetch(p, 0, 0);
        break;
      case SIMDE_MM_HINT_T0:
        __builtin_prefetch(p, 0, 3);
        break;
      case SIMDE_MM_HINT_T1:
        __builtin_prefetch(p, 0, 2);
        break;
      case SIMDE_MM_HINT_T2:
        __builtin_prefetch(p, 0, 1);
        break;
      case SIMDE_MM_HINT_ENTA:
        __builtin_prefetch(p, 1, 0);
        break;
      case SIMDE_MM_HINT_ET0:
        __builtin_prefetch(p, 1, 3);
        break;
      case SIMDE_MM_HINT_ET1:
        __builtin_prefetch(p, 1, 2);
        break;
      case SIMDE_MM_HINT_ET2:
        __builtin_prefetch(p, 0, 1);
        break;
    }
}
```

Reference: SIMDe [Issue #897](https://github.com/simd-everywhere/simde/issues/897).

howjmay:
Fixes #547

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -566,16 +566,12 @@ FORCE_INLINE uint8x16x4_t _sse2neon_vld1q_u8_x4(const uint8_t *p)
     +------+------+------+------+------+------+-------------+
  */
 
-/* Constants for use with _mm_prefetch.  */
+/* Constants for use with _mm_prefetch. */
 enum _mm_hint {
-    _MM_HINT_NTA = 0,  /* load data to L1 and L2 cache, mark it as NTA */
-    _MM_HINT_T0 = 1,   /* load data to L1 and L2 cache */
-    _MM_HINT_T1 = 2,   /* load data to L2 cache only */
-    _MM_HINT_T2 = 3,   /* load data to L2 cache only, mark it as NTA */
-    _MM_HINT_ENTA = 4, /* exclusive version of _MM_HINT_NTA */
-    _MM_HINT_ET0 = 5,  /* exclusive version of _MM_HINT_T0 */
-    _MM_HINT_ET1 = 6,  /* exclusive version of _MM_HINT_T1 */
-    _MM_HINT_ET2 = 7   /* exclusive version of _MM_HINT_T2 */
+    _MM_HINT_NTA = 0, /* load data to L1 and L2 cache, mark it as NTA */
+    _MM_HINT_T0 = 1,  /* load data to L1 and L2 cache */
+    _MM_HINT_T1 = 2,  /* load data to L2 cache only */
+    _MM_HINT_T2 = 3,  /* load data to L2 cache only, mark it as NTA */
 };
 
 // The bit field mapping to the FPCR(floating-point control register)
@@ -2303,12 +2299,25 @@ FORCE_INLINE __m128 _mm_or_ps(__m128 a, __m128 b)
 // https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_m_pmulhuw
 #define _m_pmulhuw(a, b) _mm_mulhi_pu16(a, b)
 
-// Loads one cache line of data from address p to a location closer to the
-// processor. https://msdn.microsoft.com/en-us/library/84szxsww(v=vs.100).aspx
-FORCE_INLINE void _mm_prefetch(const void *p, int i)
+// Fetch the line of data from memory that contains address p to a location in
+// the cache heirarchy specified by the locality hint i.
+// https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_prefetch
+FORCE_INLINE void _mm_prefetch(char const *p, int i)
 {
-    (void) i;
-    __builtin_prefetch(p);
+    switch (i) {
+    case _MM_HINT_NTA:
+        __builtin_prefetch(p, 0, 0);
+        break;
+    case _MM_HINT_T0:
+        __builtin_prefetch(p, 0, 3);
+        break;
+    case _MM_HINT_T1:
+        __builtin_prefetch(p, 0, 2);
+        break;
+    case _MM_HINT_T2:
+        __builtin_prefetch(p, 0, 1);
+        break;
+    }
 }
 
 // Compute the absolute differences of packed unsigned 8-bit integers in a and
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -2491,7 +2491,53 @@ result_t test_m_pmulhuw(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_prefetch(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    typedef struct {
+        __m128 a;
+        float r[4];
+    } prefetch_test_t;
+    prefetch_test_t test_vec[8] = {
+        {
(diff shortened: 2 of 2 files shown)


sse2neon.h
```diff
@@ -566,7 +566,12 @@ FORCE_INLINE uint8x16x4_t _sse2neon_vld1q_u8_x4(const uint8_t *p)
     +------+------+------+------+------+------+-------------+
  */
 
-/* Constants for use with _mm_prefetch.  */
+/* 
+Constants for use with _mm_prefetch.
+See https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/xmmintrin.h#L39
```
jserv, line 571: Don't reference code on GitHub since the line number might be changed. Instead, always cite Intel materials.

sse2neon.h
```diff
@@ -2305,10 +2310,34 @@ FORCE_INLINE __m128 _mm_or_ps(__m128 a, __m128 b)
 
 // Loads one cache line of data from address p to a location closer to the
 // processor. https://msdn.microsoft.com/en-us/library/84szxsww(v=vs.100).aspx
-FORCE_INLINE void _mm_prefetch(const void *p, int i)
+FORCE_INLINE void _mm_prefetch(char const *p, int i)
 {
-    (void) i;
-    __builtin_prefetch(p);
+    switch(i) {
```
jserv, line 2315: Run `clang-format -i` for consistent style.

sse2neon.h
```diff
@@ -2305,10 +2310,34 @@ FORCE_INLINE __m128 _mm_or_ps(__m128 a, __m128 b)
 
 // Loads one cache line of data from address p to a location closer to the
 // processor. https://msdn.microsoft.com/en-us/library/84szxsww(v=vs.100).aspx
```
jserv, line 2312: Replace the description and hyperlink with Intel website.

tests/impl.cpp
```diff
@@ -2491,7 +2491,37 @@ result_t test_m_pmulhuw(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_prefetch(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    struct prefetch_test_t {
+    __m128 a;
+    float r[4];
+  } ;
+    struct prefetch_test_t test_vec[8] = {
```
jserv, line 2498: Re-generate the test vectors with simplified values.

sse2neon.h
howjmay: I think we should update clang-format to v12, since it is quite hard for mac user install clang-format-12 which doesn't exist on brew. Maybe update to clang-format-14 is a better option.

see https://formulae.brew.sh/formula/clang-format

sse2neon.h
```diff
@@ -2304,11 +2304,36 @@ FORCE_INLINE __m128 _mm_or_ps(__m128 a, __m128 b)
 #define _m_pmulhuw(a, b) _mm_mulhi_pu16(a, b)
 
 // Loads one cache line of data from address p to a location closer to the
```
jserv, line 2306: Use Intel's description:
> Fetch the line of data from memory that contains address p to a location in the cache hierarchy specified by the locality hint i.

tests/impl.cpp
howjmay: done

tests/impl.cpp
```diff
@@ -2491,7 +2491,32 @@ result_t test_m_pmulhuw(const SSE2NEONTestImpl &impl, uint32_t iter)
 
 result_t test_mm_prefetch(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
-    return TEST_UNIMPL;
+    struct prefetch_test_t {
```
jserv, line 2494: Use `typedef` for suffix `_t`. That is.
```c
typedef struct {
    __m128 a;
    float r[4];
} prefetch_test_t;
```

sse2neon.h
jserv: version 12 is the minimal requirement indeed.

jserv: See https://github.com/DLTcollab/sse2neon/blob/master/CONTRIBUTING.md

howjmay: I was thinking about the `make indent` command, but running for each file in our case is pretty ok, too.

tests/impl.cpp
```diff
@@ -2491,7 +2491,32 @@ result_t test_m_pmulhuw(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+        {_mm_set_ps(0.5f, 0.6f, -0.7f, -0.8f), {-0.8f, -0.7f, 0.6f, 0.5f}},
+        {_mm_set_ps(0.9f, 0.10f, -0.11f, 0.12f), {0.12f, -0.11f, 0.10f, 0.9f}},
+        {_mm_set_ps(-1.1f, -2.1f, -3.1f, -4.1f), {-4.1f, -3.1f, -2.1f, -1.1f}},
+        {_mm_set_ps(100.0f, -110.0f, 120.0f, -130.0f),
+         {-130.0f, 120.0f, -110.0f, 100.0f}},
+        {_mm_set_ps(200.5f, 210.5f, -220.5f, 230.5f),
+         {995.74f, -93.04f, 144.03f, 902.50f}},
+        {_mm_set_ps(10.11f, -11.12f, -12.13f, 13.14f),
+         {13.14f, -12.13f, -11.12f, 10.11f}},
+        {_mm_set_ps(10.1f, -20.2f, 30.3f, 40.4f),
```
jserv, line 2509: Add a comma (`,`) right after `{_mm_set_ps(10.1f, -20.2f, 30.3f, 40.4f),
         {40.4f, 30.3f, -20.2f, 10.1f}}`, so that `clang-format` can indent better.

howjmay:
@jserv I removed some hints, since they are Knights Corner instead of SSE.
The followings are removed hints
```
    _MM_HINT_ENTA = 4, /* exclusive version of _MM_HINT_NTA */
    _MM_HINT_ET0 = 5,  /* exclusive version of _MM_HINT_T0 */
    _MM_HINT_ET1 = 6,  /* exclusive version of _MM_HINT_T1 */
    _MM_HINT_ET2 = 7   /* exclusive version of _MM_HINT_T2 */
```

PR #555, opened 2022-12-11
Implement five `_mm_cmpistr`

### LINKED ISSUE #308: Implement SSE4.2 text processing intrinsics
[Apache Impala](https://impala.apache.org/) adopts SSE2NEON along with partial SSE4.2 text processing intrinsics support in file [be/src/util/sse-util.h](https://github.com/apache/impala/blob/master/be/src/util/sse-util.h):
* cmpestrm
* cmpestri

However, the above were neither optimized nor generic. Instead, we can provide the full-functioned instrinsics in SSE2NEON.

Cuda-Chen:
Implement `_mm_cmpistra`, `_mm_cmpistrc`, `_mm_cmpistro`, `_mm_cmpistrs`, and `_mm_cmpistrz`.

Close #308.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -8998,12 +8998,12 @@ FORCE_INLINE int _sse2neon_ctzll(unsigned long long x)
 #define SSE2NEON_CMPSTR_SET_UPPER(var, imm) \
     const int var = (imm & 0x01) ? 8 : 16
 
-#define SSE2NEON_CMPESTRX_GET_LEN(a, b, la, lb) \
-    int tmp1 = la ^ (la >> 31);                 \
-    la = tmp1 - (la >> 31);                     \
-    int tmp2 = lb ^ (lb >> 31);                 \
-    lb = tmp2 - (lb >> 31);                     \
-    la = SSE2NEON_MIN(la, bound);               \
+#define SSE2NEON_CMPESTRX_LEN_PAIR(a, b, la, lb) \
+    int tmp1 = la ^ (la >> 31);                  \
+    la = tmp1 - (la >> 31);                      \
+    int tmp2 = lb ^ (lb >> 31);                  \
+    lb = tmp2 - (lb >> 31);                      \
+    la = SSE2NEON_MIN(la, bound);                \
     lb = SSE2NEON_MIN(lb, bound)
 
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -9139,34 +9139,51 @@ FORCE_INLINE int _mm_cmpestrz(__m128i a,
...
+            uint16x8_t equal_mask_##str = vreinterpretq_u16_u8(          \
+                vceqq_u8(vreinterpretq_u8_m128i(str), vdupq_n_u8(0)));   \
+            uint8x8_t res_##str = vshrn_n_u16(equal_mask_##str, 4);      \
+            uint64_t matches_##str =                                     \
+                vget_lane_u64(vreinterpret_u64_u8(res_##str), 0);        \
+            len = _sse2neon_ctzll(matches_##str) >> 2;                   \
+        }                                                                \
     } while (0)
 
+#define SSE2NEON_CMPISTRX_GET_LEN(a, b, la, lb)     \
```
jserv, line 9161: It is confusing to have `SSE2NEON_CMPISTRX_GET_LEN` and `SSE2NEON_CMPISTRX_GET_STR_LEN` since they look similar. Can you clarify?

Cuda-Chen: `SSE2NEON_CMPISTRX_GET_LEN` is used to get the length of string `a` and `b`.
While `SSE2NEON_CMPISTRX_GET_STR_LEN` is used to get the length of one string (`_mm_cmpistrs` and `_mm_cmpistrz` just need to get the length of string `a` and `b`, respectively).

jserv: The name `SSE2NEON_CMPISTRX_GET_STR_LEN` contains two `STR` identifiers. It looks a bit weird. We can rename `SSE2NEON_CMPISTRX_GET_LEN` to `SSE2NEON_CMPISTRX_LEN_PAIR`, which retrieves the length of two strings.

jserv: Similarly, `SSE2NEON_CMPISTRX_LENGTH` looks slightly understandable.

Cuda-Chen: Sounds great. I will make another commit to include this suggestion.

PR #556, opened 2022-12-12
Add clang/Windows build

invertego:
Add Windows CI build using [llvm-mingw](https://github.com/mstorsjo/llvm-mingw), a standalone clang/mingw-w64 toolchain that supports cross compiling.

- Windows builds are produced for x86_64, armv7, and aarch64.
- Tests are only run on x86_64. GitHub does not provide public Windows on ARM64 runners, and there is no qemu-user equivalent on Windows.
- The test framework is updated to support Windows aligned malloc/free APIs.
- The sqrt_sd/pd tests now use an epsilon for comparing results just like the sqrt_ss/ps tests. This is necessary because mingw-w64 implements C sqrt() using x87 rather than SSE.
- In case you are wondering why clang but not GCC: there are currently no builds of GCC that target Windows on ARM64.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -19,6 +19,37 @@ jobs:
           make check
           sh .ci/cross-check.sh
 
+  host_win:
+    runs-on: windows-2022
+    strategy:
+      matrix:
+        arch:
+          - x86_64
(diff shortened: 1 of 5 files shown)


jserv:

Issue #477, opened 2021-08-23
Absent `_mm_aesdec_si128` and `_mm_aesdeclast_si128`


norwend:
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aesdec_si128&expand=262
https://developer.arm.com/architectures/instruction-sets/intrinsics/vaesdq_u8

jserv:
Check [aes-brute-force/src/aes_ni_botan.cpp](https://github.com/GeneralZero/aes-brute-force/blob/dev/src/aes_ni_botan.cpp) at first glance.

- [ ] AES extension
```c
#define AES_DEC_4_ROUNDS(K) \
    do { \
        B0 = _mm_aesdec_si128(B0, K); \
        B1 = _mm_aesdec_si128(B1, K); \
        B2 = _mm_aesdec_si128(B2, K); \
        B3 = _mm_aesdec_si128(B3, K); \
    } while(0)
```

- [ ] Armv8-A Crypto extension
```c
#define AES_DEC_4_ROUNDS(K) \
    do { \
        B0 = vaesimcq_u8(vaesdq_u8(B0, K)); \
        B1 = vaesimcq_u8(vaesdq_u8(B1, K)); \
        B2 = vaesimcq_u8(vaesdq_u8(B2, K)); \
       B3 = vaesimcq_u8(vaesdq_u8(B3, K)); \
    } while(0)
```

jserv:
Another AES-Crypto mapping example: https://gist.github.com/mmozeiko/f9c999dda7dbb03722409854a1c39cc2

jserv:
IIRC, @wangxiao1254 implemented `_mm_aesimc_si128`, `_mm_aesdec_si128`, and `_mm_aesdeclast_si128` with ARMv8 Cryptography Extensions. See https://github.com/f1ed/emp/blob/sm4/emp-tool/utils/block.h

However, for SSE2NEON, we need ARMv7/non-crypto-ext counterparts.

jserv:
Drop-in implementations with ARMv8 Cryptography Extensions:
```c
__m128i _mm_aesdec_si128 (__m128i a, __m128i RoundKey) {
    return vaesimcq_u8(vaesdq_u8(a, (__m128i){})) ^ RoundKey;
}
__m128i _mm_aesdeclast_si128 (__m128i a, __m128i RoundKey) {
    return vaesdq_u8(a, (__m128i){}) ^ RoundKey;
}
[...]

PR #559, opened 2022-12-22
feat: Add _mm_aesdec_si128

howjmay:
This PR solve solve `_mm_aesdec_si128 ` of #477

This PR referred the following resources.
* https://www.linaro.org/blog/accelerated-aes-for-the-arm64-linux-kernel/
* https://github.com/torvalds/linux/blob/master/arch/arm64/crypto/aes-neon.S#L80-96
* https://crypto.stackexchange.com/questions/59311/how-is-multiplication-in-field-gf28-done
* https://cs.stackexchange.com/questions/124048/why-is-addition-in-gf28-the-same-as-xor
* https://www.workofard.com/2017/01/accelerated-aes-for-the-arm64-linux-kernel/


sse2neon.h
```diff
@@ -9533,6 +9533,17 @@ FORCE_INLINE __m128i _mm_aesenc_si128(__m128i EncBlock, __m128i RoundKey)
 #endif
 }
 
+FORCE_INLINE __m128i _mm_aesdec_si128(__m128i a, __m128i RoundKey)
+{
+#if defined(__aarch64__)
```
jserv, line 9538: You shall take the following into consideration:
1. Armv8-A without Cryptographic extension
2. [Armv8-A and A32 with Cryptographic extension](https://developer.arm.com/documentation/100242/0100/functional-description/about-the-cryptographic-extension)
3. Armv7-A

howjmay: It seems we don't have tests for function guarded by "__ARM_FEATURE_CRYPTO" macro. I think maybe I should add another makefile target to run test with `+crypto`?

jserv: Exactly, we shall have conditional `+crypto` in feasible permutations.

sse2neon.h
```diff
@@ -9533,6 +9583,78 @@ FORCE_INLINE __m128i _mm_aesenc_si128(__m128i EncBlock, __m128i RoundKey)
...
+    // mix columns
+    w = (v << 1) ^ (uint8x16_t) (((int8x16_t) v >> 7) &
+                                 0x1b);  // muliplying 'v' by 2 in GF(2^8)
+    w ^= (uint8x16_t) vrev32q_u16((uint16x8_t) v);
+    w ^= vqtbl1q_u8(v ^ w, vld1q_u8(ror32by8));
+
+    //  add round key
+    return vreinterpretq_m128i_u8(w) ^ RoundKey;
+#else /* ARMv7-A NEON implementation */
+/* FIXME: optimized for NEON */
```
howjmay, line 9625: @jserv the tables for table-based AES algorithm are way too many and huge. Therefore, I am using plain pure C implementation here. I am trying to optimize it.

howjmay: The implementation of both `helper_aesenc ` and `helper_aesdec` are table-based AES. Here are its tables https://github.com/qemu/qemu/blob/266469947161aa10b1d36843580d369d5aa38589/crypto/aes.c

tests/impl.cpp
```diff
@@ -581,6 +581,31 @@ static const uint8_t crypto_aes_sbox[256] = {
...
+    0x96, 0xac, 0x74, 0x22, 0xe7, 0xad, 0x35, 0x85, 0xe2, 0xf9, 0x37, 0xe8,
+    0x1c, 0x75, 0xdf, 0x6e, 0x47, 0xf1, 0x1a, 0x71, 0x1d, 0x29, 0xc5, 0x89,
+    0x6f, 0xb7, 0x62, 0x0e, 0xaa, 0x18, 0xbe, 0x1b, 0xfc, 0x56, 0x3e, 0x4b,
+    0xc6, 0xd2, 0x79, 0x20, 0x9a, 0xdb, 0xc0, 0xfe, 0x78, 0xcd, 0x5a, 0xf4,
+    0x1f, 0xdd, 0xa8, 0x33, 0x88, 0x07, 0xc7, 0x31, 0xb1, 0x12, 0x10, 0x59,
+    0x27, 0x80, 0xec, 0x5f, 0x60, 0x51, 0x7f, 0xa9, 0x19, 0xb5, 0x4a, 0x0d,
+    0x2d, 0xe5, 0x7a, 0x9f, 0x93, 0xc9, 0x9c, 0xef, 0xa0, 0xe0, 0x3b, 0x4d,
+    0xae, 0x2a, 0xf5, 0xb0, 0xc8, 0xeb, 0xbb, 0x3c, 0x83, 0x53, 0x99, 0x61,
+    0x17, 0x2b, 0x04, 0x7e, 0xba, 0x77, 0xd6, 0x26, 0xe1, 0x69, 0x14, 0x63,
+    0x55, 0x21, 0x0c, 0x7d};
```
jserv, line 606: Add a comma (`,`) before `}`, so that `clang-format` can indent better.

tests/impl.cpp
```diff
@@ -11565,6 +11640,11 @@ result_t test_mm_aesenclast_si128(const SSE2NEONTestImpl &impl, uint32_t iter)
     return validate128(resultReference, resultIntrinsic);
 }
 
+result_t test_mm_aesdeclast_si128(const SSE2NEONTestImpl &impl, uint32_t iter)
+{
+    return TEST_UNIMPL;
```
jserv, line 11645: Can you provide the corresponding test?

sse2neon.h
```diff
@@ -9580,7 +9703,7 @@ FORCE_INLINE __m128i _mm_aeskeygenassist_si128(__m128i key, const int rcon)
     return _mm_set_epi32(((X3 >> 8) | (X3 << 24)) ^ rcon, X3,
                          ((X1 >> 8) | (X1 << 24)) ^ rcon, X1);
 }
-#undef SSE2NEON_AES_DATA
+#undef SSE2NEON_AES_SBOX
```
jserv, line 9706: Add `#undef SSE2NEON_AES_RSBOX` as well.

sse2neon.h
```diff
@@ -9500,11 +9543,14 @@ FORCE_INLINE __m128i _mm_aesenc_si128(__m128i EncBlock, __m128i RoundKey)
     SSE2NEON_AES_B2W(p, SSE2NEON_AES_F3(p), SSE2NEON_AES_F2(p), p)
 #define SSE2NEON_AES_U3(p) \
     SSE2NEON_AES_B2W(p, p, SSE2NEON_AES_F3(p), SSE2NEON_AES_F2(p))
+
+    // this create a table which contains all the permutation of shift_rows(),
```
jserv, line 9547: Improve the writing.

sse2neon.h
jserv: Agree. Let's stick on shorter implementation since Armv7 path exits for compatibility purpose.

tests/impl.cpp
howjmay: This hasn't been implemented yet. I want to implement `_mm_aesdeclast_si128` in another PR, since I still study how to implement it on ARM64.

sse2neon.h
jserv: should be
> This generates a table containing every possible permutation of  `shift_rows()` and `sub_bytes()` with `mix columns()`.

jserv (changes requested):
Resolve the conflicts.

PR #560, opened 2022-12-23
fix: Assign architecture in CI job host_x86 

howjmay:
This PR assigns the architecture, x86_64 in GitHub Actions job
'host_x86'.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -7,6 +7,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
+        arch: [x86_64]
         cxx_compiler: [g++-10, clang++-11]
     steps:
       - name: checkout code


howjmay:
@jserv I am curious about the purpose of job `host_x86`. I think it aims to ensure our tests cases align the behavior of SSE intrinsics. However, if we run the test with `cross-check.sh` script, then I think it has little difference from `host_arm`. Or just testing the abilities of cross compile and the toolchains.

jserv:
> @jserv I am curious about the purpose of job `host_x86`. I think it aims to ensure our tests cases align the behavior of SSE intrinsics. However, if we run the test with `cross-check.sh` script, then I think it has little difference from `host_arm`. Or just testing the abilities of cross compile and the toolchains.

`host-x86` exists because both clang and gcc can not perfectly generate machine from intrinsics. Some versions of gcc lack of certain intrinsics, and clang fails to accept some usages. Therefore, it is used to verify host intrinsics compatibility. Later, we can revisit the FIXME/TODO to see if we can resolve.

PR #564, opened 2022-12-24
Add description for table-based sub_bytes() and lowercase sse2neon prefix

howjmay:


sse2neon.h
```diff
@@ -9486,8 +9486,8 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
 
 /* X Macro trick. See https://en.wikipedia.org/wiki/X_Macro */
 #define SSE2NEON_AES_H0(x) (x)
-static const uint8_t SSE2NEON_sbox[256] = SSE2NEON_AES_SBOX(SSE2NEON_AES_H0);
-static const uint8_t SSE2NEON_rsbox[256] = SSE2NEON_AES_RSBOX(SSE2NEON_AES_H0);
+static const uint8_t sse2neon_sbox[256] = SSE2NEON_AES_SBOX(SSE2NEON_AES_H0);
```
jserv, line 9489: Enforce `_sse2neon_` prefix. Thats is, use `_sse2neon_sbox` and `_sse2neon_rsbox`.

The "_" is a POSIX lib convention Please read 7.1.3 Reserved identifiers of [the C Standard](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf):
> All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use. and All identifiers that begin with an underscore are always reserved for use as identifiers with file scope in both the ordinary and tag name spaces. 

Additionally, subclause 7.31 defines many other reserved identifiers for future library directions.

I will add the above notice in `CONTRIBUTING.md`.

sse2neon.h
```diff
@@ -9673,33 +9679,32 @@ FORCE_INLINE __m128i _mm_aesenclast_si128(__m128i a, __m128i RoundKey)
...
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 3)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 8)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 13)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 2)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 7)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 12)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 1)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 6)],
-        SSE2NEON_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 11)],
+        sse2neon_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), 0)],
```
jserv, line 9692: With [X macro](https://en.wikipedia.org/wiki/X_Macro) trick, we can shorten the implementation as following:
```c
#define SSE2NEON_SBOX_INITIALIZERS \
    _(5), _(10), _(15), _(4), ... /* Use a better name though */

uint8_t v[16] = {
#define _(x) _sse2neon_sbox[vgetq_lane_u8(vreinterpretq_u8_m128i(a), x)]
    SSE2NEON_SBOX_INITIALIZERS
#undef _
};
```

PR #574, opened 2022-12-28
test: Build/run with crypto and/or crc

howjmay:
The current test has a lack of tests with crypto flag enabled which cause the tests not cover the total implementation.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -55,7 +55,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        arch: [armv7, aarch64]
+        arch_with_features: [{arch: armv7, feature: none}, {arch: aarch64, feature: none}, {arch: aarch64, feature: crypto+crc}]
         cxx_compiler: [g++-10, clang++-11]
     steps:
       - name: checkout code
@@ -65,7 +65,7 @@ jobs:
         # https://github.com/uraimo/run-on-arch-action
         uses: uraimo/run-on-arch-action@v2.5.0
         with:
-          arch: ${{ matrix.arch }}
+          arch: ${{ matrix.arch_with_features.arch }}
           distro: ubuntu20.04
           env: |
             CXX: ${{ matrix.cxx_compiler }}
@@ -73,7 +73,8 @@ jobs:
             apt-get update -q -y
             apt-get install -q -y "${{ matrix.cxx_compiler }}" make
             apt-get install -q -y gcc
-          run: make check
+          run: |
+            make FEATURE=${{ matrix.arch_with_features.feature }} check
 
   coding_style:
     runs-on: ubuntu-20.04
--- a/Makefile
+++ b/Makefile
@@ -30,13 +30,21 @@ endif
 
 # Follow platform-specific configurations
 ifeq ($(processor),$(filter $(processor),aarch64 arm64))
-    ARCH_CFLAGS = -march=armv8-a+fp+simd+crc
+    ARCH_CFLAGS = -march=armv8-a+fp+simd
 else ifeq ($(processor),$(filter $(processor),i386 x86_64))
     ARCH_CFLAGS = -maes -mpclmul -mssse3 -msse4.2
 else ifeq ($(processor),$(filter $(processor),arm armv7 armv7l))
     ARCH_CFLAGS = -mfpu=neon
 else
     $(error Unsupported architecture)
+endif
+
+FEATURE ?=
+ifneq ($(FEATURE),)
+ifneq ($(FEATURE),none)
+COMMA:= ,
+ARCH_CFLAGS := $(ARCH_CFLAGS)+$(subst $(COMMA),+,$(FEATURE))
+endif
 endif
 
 CXXFLAGS += -Wall -Wcast-qual -I. $(ARCH_CFLAGS) -std=gnu++14
--- a/README.md
+++ b/README.md
@@ -85,6 +85,13 @@ runtime. Use the following commands to perform test cases:
 $ make check
 ```
 
+For running check with enabling features, you can use assign the features with `FEATURE` command.
+If `none` is assigned, then the command will be the same as simply calling `make check`.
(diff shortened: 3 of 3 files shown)


.github/workflows/github_actions.yml
```diff
@@ -69,11 +70,17 @@ jobs:
...
             CXX: ${{ matrix.cxx_compiler }}
+            FEATURE: ${{matrix.feature}}
           install: |
             apt-get update -q -y
             apt-get install -q -y "${{ matrix.cxx_compiler }}" make
             apt-get install -q -y gcc
-          run: make check
+          run: |
+            if [ $FEATURE == "crypto" ]; then
+              make check-crypto
```
jserv, line 80: I didn't see any rules associated with `check-crypto`.

Makefile
```diff
@@ -63,6 +63,12 @@ ifeq ($(processor),$(filter $(processor),aarch64 arm64 arm armv7l))
 endif
 	$(EXEC_WRAPPER) $^
 
+check-crypto: tests/main
```
jserv, line 66: How about accepting a variable `FEATURES` which can be recognized as the corresponding flags?
See https://github.com/sysprog21/rv32emu/blob/master/Makefile#L69

Sample usage:
```shell
make FEATURE=crypto clean all
make FEATURE=crypto,crc clean all
```

jserv (changes requested):
Split the changes against `sse2neon.h` to another pull request.

Makefile
howjmay: it seems we need to exclude `armv7` from testing crypto flag?

jserv: ARMv8 introduced hardware acceleration for cryptography. From the perspective of Armv8, it is possible to use the cores with 32-bit mode only. That is, A32, which allows the use of cryptographic extension as well.

To make it clear, we should consider the following targets:
* Armv7-A: 32-bit. NEON is the necessary feature for SSE2NEON.
* Armv8-A: 64-bit and/or 32-bit. It comes with several extensions such as cryptography and CRC.

See also: https://github.com/ThomasKaiser/sbc-bench/blob/master/results/ARMv8-Crypto-Extensions.md

Makefile
```diff
@@ -63,6 +69,12 @@ ifeq ($(processor),$(filter $(processor),aarch64 arm64 arm armv7l))
...
+ifeq ($(processor),$(filter $(processor),aarch64 arm64 arm armv7l))
+	$(CC) $(ARCH_CFLAGS)+crypto -c sse2neon.h
```
jserv, line 74: The use of `+crypto` looks ugly.

.github/workflows/github_actions.yml
```diff
@@ -57,6 +57,7 @@ jobs:
       matrix:
         arch: [armv7, aarch64]
         cxx_compiler: [g++-10, clang++-11]
+        feature: ["none", "crypto"]
```
howjmay, line 60: It seems `uraimo/run-on-arch-action@v2.5.0` doesn't allow us to set an empty element in the matrix, so I use a `none` element to represent the situation

Makefile
howjmay: I have removed this target

Makefile
```diff
@@ -39,6 +39,12 @@ else
     $(error Unsupported architecture)
 endif
 
+FEATURE ?=
+ifneq ($(FEATURE),)
```
jserv, line 43: We should take `make FEATURE=none` into consideration.

howjmay: I have excluded `FEATURE =none` in GitHub Actions workflow. Or should I move the condition check to here?

jserv: You don't have to exclude `FEATURE=none` in GitHub Actions since we always do `make check`. Just refine `Makefile` instead.

jserv (changes requested):
Rebase the latest `master` branch (force-pushed earlier).

howjmay:
A warning `WARNING: The requested image's platform (linux/arm/v7) does not match the detected host platform (linux/amd64) and no specific platform was requested` is returned, and I don't think the CI has been successfully run. I am investigating it

Makefile
```diff
@@ -39,6 +39,14 @@ else
...
+ifneq ($(FEATURE),none)
+COMMA:= ,
+ARCH_CFLAGS := $(ARCH_CFLAGS)+$(subst $(COMMA),+,$(FEATURE))
```
jserv, line 46: What if `make FEATURE=crypto,crc` or `make FEATURE=crypto+crc` is run?

howjmay: Now the command should be `make FEATURE=crypto,crc check`. However, the bigger issue this PR is facing is that CI  is not successfully triggered on armv7

jserv: Would we rollback older version of `run-on-arch-action`? Meanwhile, we should file a new issue there.

howjmay: I think it was an error caused by I set the workflow yaml in wrong way. If not include this PR, the current CI seems to work fine

jserv: By the way, you should modify `README.md` for the `FEATURE=` information while running `make check`.

howjmay: Information has been added, and the tests work fine now. I made some silly error.

.github/workflows/github_actions.yml
```diff
@@ -55,7 +55,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        arch: [armv7, aarch64]
+        arch_with_features: [{arch: armv7, feature: none}, {arch: aarch64, feature: none}, {arch: aarch64, feature: crypto}]
```
jserv, line 58: I have changed the subject of this pull request and would like to address `crc` feature in CI pipeline as well.

howjmay: Thanks! Btw should we remove the current crc enabling in Makefile line 33?

jserv: Yes, go ahead.

howjmay: done

PR #577, opened 2023-01-04
Improve precision of _mm_{rsqrt,sqrt,rcp,div}_{ps,ss} conversions

### LINKED ISSUE #526: precision issue with rsqrt/sqrt/rcp/div
Hi,
I been developing with SSE for many may years, and now got into porting some code to native ARM.
sse2neon works great, and I'm very happy with it, but I realized there are some precision incompatibility.
To my knowledge there are only 2 non exact instruction in SSE: `rsqrt` and `rcp` (`p` and `s` versions), both got 11 bit accuracy.
Neon got `vrsqrteq_f32` and `vrecpeq_f32`, that got 8 bit accuracy, 1% and the `vrsqrtsq_f32` and `vrecpsq_f32` used to improve accuracy.  one Netwon-Raphson step will get about 16 correct bits and two steps seems to get to full 24 bits.

`_mm_rsqrt_ps`  currently with `SSE2NEON_PRECISE_SQRT=0 `is not doing any nr iterations, and with `SSE2NEON_PRECISE_SQRT=1`
 two, it should to one. now we have options for 8 bits or 24, not the 16 that will match the expected 11 bits on Intel.

`_mm_sqrt_ps` is expected to be exact and do 2 Netwon-Raphson iterations in non `__aarch64__`  path, and use the `__aarch64__ vsqrtq_f32` if available.  currently its ok for `__aarch64__` with default `SSE2NEON_PRECISE_SQRT=0`, but return only 8 bits accuracy for non `__aarch64__` arch.  with `SSE2NEON_PRECISE_SQRT=1` its will use the long `vrsqrteq_f32 + 2*vrsqrtsq_f32` even in `__aarch64__` is available.

`_mm_rcp_ps` is correct with default `SSE2NEON_PRECISE_DIV` off, and doing pointless additional step with `SSE2NEON_PRECISE_DIV` on.  

`_mm_div_ps` is that same situation as in `_mm_sqrt_ps`, it expected to be exact, and use `__aarch64__`  `vdivq_f32` if available, or do 2 Netwon-Raphson iterations.

unit test to bu updated too, 0.1% for `rsqrt`/`rcp`, and probably result*FLT_EPS(maybe *2-3)  for `div`/`sqrt`

Option to use lower precision can be added, but it shouldn't  be on by default, and probably need to be more fine grained, for example `rsqrt_8_BITS` `rcp_8_BITS` `div_16_bits`, and `sqrt16_bits`.  but I don't bee them that useful. 

here is some code from Skia that seems to do it correct
https://chromium.googlesource.com/skia/+/chrome/m69/src/opts/SkNx_neon.h
and see reference
https://chromium.googlesource.com/skia/+/chrome/m69/src/opts/SkNx_sse.h

Cuda-Chen:
1. Improve precision of `_mm_{rsqrt,rcp}_{ps,ss}`. Currently the precision of this part is guaranteed below 0.1% relative error.
2. Improve precision of `_mm_{sqrt,div}_{ps,ss}`. Currently the precision of this part is guaranteed below $10^{-6}$ % relative error.
3. Remove `SSE2NEON_PRECISE_DIV` as it is no longer being used.
    
Close #526.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -11,7 +11,7 @@ jobs:
         cxx_compiler: [g++-10, clang++-11]
     steps:
       - name: checkout code
-        uses: actions/checkout@v3.1.0
+        uses: actions/checkout@v3.2.0
       - name: build artifact
         env:
           CXX: ${{ matrix.cxx_compiler }}
@@ -42,7 +42,7 @@ jobs:
           mv llvm-mingw-* "$HOME/llvm-mingw"
           echo "$HOME/llvm-mingw/bin" >> $GITHUB_PATH
       - name: checkout code
-        uses: actions/checkout@v3.0.2
+        uses: actions/checkout@v3.2.0
       - name: build artifact
         env:
           CXX: ${{ matrix.arch }}-w64-mingw32-clang++
@@ -63,7 +63,7 @@ jobs:
       - name: build artifact
         # The Github Action for non-x86 CPU
         # https://github.com/uraimo/run-on-arch-action
-        uses: uraimo/run-on-arch-action@v2.3.0
+        uses: uraimo/run-on-arch-action@v2.5.0
         with:
           arch: ${{ matrix.arch }}
           distro: ubuntu20.04
--- a/README.md
+++ b/README.md
@@ -70,7 +70,6 @@ but SSE intrinsic `_mm_maddubs_epi16` has to be implemented with 13+ NEON instru
 
 Considering the balance between correctness and performance, `sse2neon` recognizes the following compile-time configurations:
 * `SSE2NEON_PRECISE_MINMAX`: Enable precise implementation of `_mm_min_ps` and `_mm_max_ps`. If you need consistent results such as NaN special cases, enable it.
-* `SSE2NEON_PRECISE_DIV`: Enable precise implementation of `_mm_rcp_ps` and `_mm_div_ps` by additional Netwon-Raphson iteration for accuracy.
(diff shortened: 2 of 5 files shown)


jserv (changes requested):
Rebase the latest `master` branch.

jserv:
> 1. Improve precision of _mm_{rsqrt,rcp}_{ps,ss}. Currently the
>    precision of this part is guaranteed below 0.1% relative error.
> 2. Improve precision of _mm_{sqrt,div}_{ps,ss}. Currently the
>    precision of this part is guaranteed below 10^-6% relative error.
> 3. Remove SSE2NEON_PRECISE_DIV as it is no longer being used.

By default, performance is prioritized over accuracy. You shall lower the performance impact. The reason why `SSE2NEON_PRECISE_DIV` exists is because some applications such as Blender do require precision, but other would expect acceptable performance.

Cuda-Chen:
Re-open the PR for branch tidy.

shord:
Hi,
Good job, but I still see a few issues.
_mm_sqrt_ps need two vrsqrteq_f32 iteration in all cases like in _mm_div_ps.  if some one if fine with 8 bits precision, they need to do its explicetly, or add a non default option for low precition.   probebly will not impact performance mutch in any case

There is no need for the #if SSE2NEON_PRECISE_SQRT block in _mm_rsqrt , just do the one vrsqrtsq_f32 iteration in all cases.  no one will expect 24 bits precision from this function.


_mm_div_ps/_mm_sqrt_ps  will have a 2 bits of imprecision in the non __aarch64__ case , but not much to do about it,  a/b is not the same as a*(1/b).  just good idea to coment about it .  the __aarch64__  is expected to be exact.

jserv:
> Rebase the latest `master` branch.

@Cuda-Chen, you can expose the changes via `git format-patch` and make a new commit manually since you got into trouble with the rebasing.

PR #580, opened 2023-01-06
Improve precision of _mm_{rsqrt,sqrt,rcp,div}_{ps,ss} conversions

### LINKED ISSUE #526: precision issue with rsqrt/sqrt/rcp/div

Cuda-Chen:
1. Improve precision of `_mm_{rsqrt,rcp}_{ps,ss}`. The precision of this part is guaranteed below 0.1% relative error.
2. Improve precision of `_mm_sqrt_{ps,ss}`. The precision of `_mm_sqrt_{ps,ss}` is guaranteed below $10^{-6}$% relative error on ARMv8-A and $10^{-4}$% relative error on ARMv7-A.
3. Add warnings of `_mm_sqrt_ps` and `_mm_div_ps` for ARMv7-A NEON implementation
as the results are not IEEE754-compliant due to lack of precise square
root and division intrinsic.
4. Add deprecation notice of `SSE2NEON_PRECISE_DIV` and `SSE2NEON_PRECISE_SQRT`
as accuracy has precedence of performance in `_mm_{rsqrt,sqrt,rcp,div}_{ps,ss}`.

Close #526.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1665,19 +1665,20 @@ FORCE_INLINE int64_t _mm_cvttss_si64(__m128 a)
 
 // Divide packed single-precision (32-bit) floating-point elements in a by
 // packed elements in b, and store the results in dst.
+// Due to ARMv7-A NEON's lack of a precise division intrinsic, we implement
+// division by multiplying a by b's reciprocal before using the Newton-Raphson
+// method to approximate the results.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_div_ps
 FORCE_INLINE __m128 _mm_div_ps(__m128 a, __m128 b)
 {
-#if defined(__aarch64__) && !SSE2NEON_PRECISE_DIV
+#if defined(__aarch64__)
     return vreinterpretq_m128_f32(
         vdivq_f32(vreinterpretq_f32_m128(a), vreinterpretq_f32_m128(b)));
 #else
     float32x4_t recip = vrecpeq_f32(vreinterpretq_f32_m128(b));
     recip = vmulq_f32(recip, vrecpsq_f32(recip, vreinterpretq_f32_m128(b)));
-#if SSE2NEON_PRECISE_DIV
     // Additional Netwon-Raphson iteration for accuracy
     recip = vmulq_f32(recip, vrecpsq_f32(recip, vreinterpretq_f32_m128(b)));
-#endif
     return vreinterpretq_m128_f32(vmulq_f32(vreinterpretq_f32_m128(a), recip));
 #endif
 }
@@ -1686,6 +1687,8 @@ FORCE_INLINE __m128 _mm_div_ps(__m128 a, __m128 b)
 // lower single-precision (32-bit) floating-point element in b, store the result
 // in the lower element of dst, and copy the upper 3 packed elements from a to
 // the upper elements of dst.
+// Warning: ARMv7-A does not produce the same result compared to Intel and not
+// IEEE-compliant.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_div_ss
 FORCE_INLINE __m128 _mm_div_ss(__m128 a, __m128 b)
 {
@@ -2210,10 +2213,6 @@ FORCE_INLINE __m128 _mm_rcp_ps(__m128 in)
 {
     float32x4_t recip = vrecpeq_f32(vreinterpretq_f32_m128(in));
     recip = vmulq_f32(recip, vrecpsq_f32(recip, vreinterpretq_f32_m128(in)));
-#if SSE2NEON_PRECISE_DIV
-    // Additional Netwon-Raphson iteration for accuracy
-    recip = vmulq_f32(recip, vrecpsq_f32(recip, vreinterpretq_f32_m128(in)));
-#endif
     return vreinterpretq_m128_f32(recip);
 }
 
@@ -2234,13 +2233,8 @@ FORCE_INLINE __m128 _mm_rcp_ss(__m128 a)
 FORCE_INLINE __m128 _mm_rsqrt_ps(__m128 in)
 {
     float32x4_t out = vrsqrteq_f32(vreinterpretq_f32_m128(in));
-#if SSE2NEON_PRECISE_SQRT
-    // Additional Netwon-Raphson iteration for accuracy
     out = vmulq_f32(
         out, vrsqrtsq_f32(vmulq_f32(vreinterpretq_f32_m128(in), out), out));
-    out = vmulq_f32(
-        out, vrsqrtsq_f32(vmulq_f32(vreinterpretq_f32_m128(in), out), out));
-#endif
     return vreinterpretq_m128_f32(out);
 }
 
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Rebase the latest `master` branch where obsolete Microsoft hyperlinks were replaced.

jserv:
> 3. Enable `SSE2NEON_PRECISE_SQRT` by default suggested by https://github.com/DLTcollab/sse2neon/pull/577#issuecomment-1372241730.

Can you show some measurement results regarding `SSE2NEON_PRECISE_SQRT`?

jserv:
Cc. @shord, please help review.

Cuda-Chen:
> Can you show some measurement results regarding `SSE2NEON_PRECISE_SQRT`?

Currently only `_mm_sqrt_{ps,ss}` is affected with `SSE2NEON_PRECISE_SQRT`. Therefore, I will just write the measurement result of `_mm_sqrt_ss`.

The measurement shows the relative error between ARMv7-A NEON implementation and C implementation with/without enabling `SSE2NEON_PRECISE_SQRT`. The input data is generated by test program of this project, and the below shows the measurement:
- `SSE2NEON_PRECISE_SQRT=0`
  - mean: 1.0294616167934348e-6
  - std: 2.0487346409843377e-6
- `SSE2NEON_PRECISE_SQRT=1`
  - mean: 2.475134107285829e-8
  - 4.406416732048071e-8

shord:
there are still a few problems
sse2neon as a simple drop in h file is expected to generate code that will not create unexpected regression of the program results.  Later, if someone has performance issues they can profile and optimize  the arm code, and decide if precision can be reduced.

_mm_div_ps and _mm_sqrt_ps are exact and slow instructions on x86, with a latency of 12 cycles and low throughput. If someone can use an approximation they can use rcp and rsqrt combination in the original x86 code to get high throughput with about the same latency.    
For the non __aarch64__ even with two refinement iteration _mm_div_ps and _mm_sqrt_ps is not exact, it will have 1 bit error do to rounding before multiplying, but it seems a reasonable compromise, but probably needs to be warned about in the documentation.

SSE2NEON_PRECISE_SQRT and SSE2NEON_PRECISE_DIV can be defined to use scalar instruction for full IEEE compliance for non __aarch64__ case.

1.  Remove SSE2NEON_PRECISE_DIV  test  from _mm_rcp_ps, do 1 vrecpsq_f32 iteration
2.  Remove SSE2NEON_PRECISE_SQRT test from _mm_sqrt_ps and to 2 vrsqrtsq_f32 iteration
3.   add warning that _mm_div_(p/s)s _mm_sqrt_(p/s)s are not the same as an intel cpu and not IEEE compliant for arm32

Later we can add SSE2NEON_PRECISE_SQRT and SSE2NEON_PRECISE_DIV to have fully compliant _mm_div_(p/s)sand  _mm_sqrt_(p/s)s.

Maybe for the scalar case  _mm_div_ss and _mm_sqrt_ss will be faster with scalar code?  Now they call the ps version and use only one result.

Cuda-Chen:
> Maybe for the scalar case `_mm_div_ss` and `_mm_sqrt_ss` will be faster with scalar code? Now they call the `ps` version and use only one result.

I will create a separated PR to handle this.

sse2neon.h
```diff
@@ -65,7 +65,7 @@
 #endif
 /* _mm_sqrt_ps and _mm_rsqrt_ps */
 #ifndef SSE2NEON_PRECISE_SQRT
-#define SSE2NEON_PRECISE_SQRT (0)
+#define SSE2NEON_PRECISE_SQRT (1)
```
jserv, line 68: If you change the default configurations, there must be a strong reason. Think of the compatibility and consistency.

sse2neon.h
```diff
@@ -1665,19 +1665,19 @@ FORCE_INLINE int64_t _mm_cvttss_si64(__m128 a)
 
 // Divide packed single-precision (32-bit) floating-point elements in a by
 // packed elements in b, and store the results in dst.
+// Warning: ARMv7-A does not produce the same result compared to Intel and not
```
jserv, line 1668: Improve English writing and mention the reason.

Cuda-Chen:
Hi @jserv ,
May I add @shord as a reviewer of this PR for notifying him to review once any changes are commited?

sse2neon.h
```diff
@@ -1665,19 +1665,20 @@ FORCE_INLINE int64_t _mm_cvttss_si64(__m128 a)
 
 // Divide packed single-precision (32-bit) floating-point elements in a by
 // packed elements in b, and store the results in dst.
+// Warning: as ARMv7-A NEON does not have exact division function, we implement
```
jserv, line 1668: Rephrase as following:
> Due to ARMv7-A NEON's lack of a precise division intrinsic, we implement division by multiplying a by b's reciprocal before using the Newton-Raphson method to approximate the results.

sse2neon.h
```diff
@@ -2550,10 +2544,16 @@ FORCE_INLINE void _mm_lfence(void)
 
 // Compute the square root of packed single-precision (32-bit) floating-point
 // elements in a, and store the results in dst.
+// Warning: as ARMv7-A NEON does not have exact square root function, we
```
jserv, line 2547: Ditto. Rephrase.

tests/common.cpp
```diff
@@ -316,6 +316,23 @@ result_t validateFloatEpsilon(__m128 a,
     float df1 = fabsf(t[1] - f1);
     float df2 = fabsf(t[2] - f2);
     float df3 = fabsf(t[3] - f3);
+
+    if ((std::isnan(t[0]) && std::isnan(f0)) || (t[0] == 0 && f0 == 0)) {
```
jserv, line 320: Add some comments which describe the motivation.

jserv (changes requested):
Mention the exclusion of macro "SSE2NEON_PRECISE_DIV" in git commit messages.

sse2neon.h

sse2neon.h

tests/common.cpp
Cuda-Chen: Done.

jserv:

shord:
Good job,  Everything looks ok!
Just need a few more cleanups: 
Remove all left overs of SSE2NEON_PRECISE_DIV SSE2NEON_PRECISE_SQRT from h file read me and testing, its not deprecated, it's just not used.

PR #584, opened 2023-01-11
tests: Cover all immediate numbers

### LINKED ISSUE #582: Cover complete immediate numbers in tests
For intrinsics that take immediate numbers, we are only testing few of the immediate numbers in the valid range.
Maybe we should use macro to tests all the numbers within valid range?

howjmay:
This PR use X-Macro skill to iterate all the possible immediate numbers for the intrinsics that allow immediate numbers only.

closes #582

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -4824,21 +4824,38 @@ result_t test_mm_extract_epi16(const SSE2NEONTestImpl &impl, uint32_t iter)
     return TEST_SUCCESS;
 }
 
+#define IMM_8_ITER \
+    TEST(0)        \
+    TEST(1)        \
+    TEST(2)        \
+    TEST(3)        \
+    TEST(4)        \
+    TEST(5)        \
+    TEST(6)        \
+    TEST(7)
+#define ASSERT_VALIDATE_INT16(a, b)                                            \
+    assert(validateInt16(a, b[0], b[1], b[2], b[3], b[4], b[5], b[6], b[7]) == \
+           TEST_SUCCESS)
 result_t test_mm_insert_epi16(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
     const int16_t *_a = (const int16_t *) impl.mTestIntPointer1;
     const int16_t insert = (int16_t) *impl.mTestIntPointer2;
-    const int imm8 = 2;
 
-    int16_t d[8];
-    for (int i = 0; i < 8; i++) {
-        d[i] = _a[i];
-    }
-    d[imm8] = insert;
+#define TEST(IDX)                                           \
+    int16_t d##IDX[8];                                      \
+    for (int i = 0; i < 8; i++) {                           \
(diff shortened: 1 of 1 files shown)


howjmay:
@jserv Do you think this PR is a good idea? I think even though this PR covers all the possible immediate numbers, the implementation become less readable and harder to implement. 
However, in the meantime, I don't think the tests will be ones that changed frequently, or even won't be changed at all in the future.

tests/impl.cpp
```diff
@@ -4824,21 +4824,37 @@ result_t test_mm_extract_epi16(const SSE2NEONTestImpl &impl, uint32_t iter)
...
+#define TEST(IDX)                                                            \
+    int16_t d##IDX[8];                                                       \
+    for (int i = 0; i < 8; i++) {                                            \
+        d##IDX[i] = _a[i];                                                   \
+    }                                                                        \
+    d##IDX[IDX] = insert;                                                    \
+                                                                             \
+    __m128i a##IDX = load_m128i(_a);                                         \
+    __m128i b##IDX = _mm_insert_epi16(a##IDX, insert, IDX);                  \
+    assert(validateInt16(b##IDX, d##IDX[0], d##IDX[1], d##IDX[2], d##IDX[3], \
```
jserv, line 4851: Can you simplify this line as well?

howjmay: Ok

jserv:
> @jserv Do you think this PR is a good idea? I think even though this PR covers all the possible immediate numbers, the implementation become less readable and harder to implement. However, in the meantime, I don't think the tests will be ones that changed frequently, or even won't be changed at all in the future.

Alternatively, we can even generate test cases via shell scripts with more flexibility.

howjmay:
> > @jserv Do you think this PR is a good idea? I think even though this PR covers all the possible immediate numbers, the implementation become less readable and harder to implement. However, in the meantime, I don't think the tests will be ones that changed frequently, or even won't be changed at all in the future.
> 
> Alternatively, we can even generate test cases via shell scripts with more flexibility.

May I ask for more details?

jserv:
> > Alternatively, we can even generate test cases via shell scripts with more flexibility.
> May I ask for more details?

Check this: https://github.com/IBM/microprobe/tree/master/targets/riscv/examples
There are various template-based generators.

howjmay:
> > > Alternatively, we can even generate test cases via shell scripts with more flexibility.
> > > May I ask for more details?
> 
> Check this: https://github.com/IBM/microprobe/tree/master/targets/riscv/examples There are various template-based generators.

I didn't see the  templates for SSE. I saw only PowerPC and RISCV only. Not sure how can we generate tests by this. I may need more research

howjmay:
So the current change is an acceptable one or I should discard it?

howjmay:
I will extend the same implementation to the other instrinsics with immediate numbers

PR #588, opened 2023-03-17
Fix MSVC compile issues

toxieainc:


jserv:
Thank @toxieainc for contributing!

PR #589, opened 2023-03-17
Fix a whole couple of 64bit-MSVC/clang and MSVC/non-clang issues

toxieainc:


jserv:
@toxieainc, I appreciate your efforts very much. Can you add MSVC related rules for GitHub Actions? Thus, we can automate the validation process.

toxieainc:
@jserv unfortunately i have no clue about that. Also it will not build completely (still) with non-clang MSVC, due to a lot of errors where fixing these in detail goes over my head. But at least the clang MSVC builds will be good.

PR #593, opened 2023-04-14
Add notice that NEON handles certain IEEE single-precision values

### LINKED ISSUE #527: `SSE2NEON_PRECISE_MINMAX`  clarification
Just looked into other of the library setting, and its not clear if `SSE2NEON_PRECISE_MINMAX`  is needed for correct handling of infinity
It's probably fine, but there is not much explanation about the problems its solve in the header file.
I don't see any inf in the testing code too.

Cuda-Chen:
Add notice to remind user how NEON handles denormal or NaN values in IEEE single-precision format.

Close #527.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,9 @@ but SSE intrinsic `_mm_maddubs_epi16` has to be implemented with 13+ NEON instru
 
 ## Compile-time Configurations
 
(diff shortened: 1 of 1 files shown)


jserv:
Thank @Cuda-Chen for contributing.

PR #594, opened 2023-04-24
Add infinity test in `test_mm_{max,min}_{pd,sd}`

Cuda-Chen:
There are no any test cases for testing infinity in `test_mm_{max,min}_{pd,sd}`, which will let user feel confused whether using infinity will work on these conversions or not.

As a result, this PR adds positive/negative infinity test cases for these four intrinsic conversions.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -145,6 +145,22 @@ public:
                 }
             }
 
+            if (test == it_mm_max_pd || test == it_mm_max_sd ||
+                test == it_mm_min_pd || test == it_mm_min_sd) {
+                // Make sure the positive/negative inifinity values are included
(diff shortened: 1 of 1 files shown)


tests/impl.cpp
```diff
@@ -5096,7 +5112,6 @@ result_t test_mm_max_pd(const SSE2NEONTestImpl &impl, uint32_t iter)
     __m128d a = load_m128d(_a);
     __m128d b = load_m128d(_b);
     __m128d c = _mm_max_pd(a, b);
-
```
jserv, line 5099: Don't do this. Minimize the necessary changes.

Cuda-Chen: Fixed.

Issue #384, opened 2021-04-13
MSVC support


Xottab-DUTY:
MSVC support will be nice.

Also, is it worth the effort, if I make a pull request with some changes that will improve the compatibility with MSVC? (I know almost nothing about ARM, but I can try to fix some macros, etc, maybe small things, if no one mind about the idea)

jserv:
> MSVC support will be nice.
> Also, is it worth the effort, if I make a pull request with some changes that will improve the compatibility with MSVC? (I know almost nothing about ARM, but I can try to fix some macros, etc, maybe small things, if no one mind about the idea)

Welcome! Can you enable GitHub Actions integration with Visual Studio?
See https://devblogs.microsoft.com/visualstudio/whats-new-with-github-actions-tooling-in-visual-studio/

aaronfranke:
When I try to compile sse2neon for ARM Windows using MSVC, it triggers the `#error "Macro name collisions may happen with unsupported compiler."` line in the below code, which fails the build:

https://github.com/DLTcollab/sse2neon/blob/master/sse2neon.h#L83

```cpp
#if defined(__GNUC__) || defined(__clang__)
#pragma push_macro("FORCE_INLINE")
#pragma push_macro("ALIGN_STRUCT")
#define FORCE_INLINE static inline __attribute__((always_inline))
#define ALIGN_STRUCT(x) __attribute__((aligned(x)))
#ifndef likely
#define likely(x) __builtin_expect(!!(x), 1)
#endif
#ifndef unlikely
#define unlikely(x) __builtin_expect(!!(x), 0)
#endif
#else
#error "Macro name collisions may happen with unsupported compiler."
#ifdef FORCE_INLINE
#undef FORCE_INLINE
#endif
#define FORCE_INLINE static inline
#ifndef ALIGN_STRUCT
#define ALIGN_STRUCT(x) __declspec(align(x))
#endif
#endif
#ifndef likely
#define likely(x) (x)
#endif
#ifndef unlikely
#define unlikely(x) (x)
#endif
```

jserv:
Line 83 is indeed `#warning` rather than `#error`. At present, SSE2NEON is built with clang and gcc.
@aaronfranke, Can you remove the part from Line 105 to Line 120 and build with MSVC?

aaronfranke:
@jserv When I comment out the error line and lines 105 to 120, I get hundreds of errors. Here are some of the first ones:

```
[...]

PR #596, opened 2023-05-17
MSVC support

anthony-linaro:
This resolves issue #384.

Much of this is based on the initial effort done by @invertego in said issue, however it adds some extras, and ensures that all the tests pass correctly.

This was tested on my Thinkpad X13s running Windows 11 22H2 (22621.1702), and using SDK 10.0.22621.


sse2neon.sln
```diff
@@ -0,0 +1,39 @@
...
+		Debug|ARM = Debug|ARM
+		Debug|ARM64 = Debug|ARM64
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|ARM = Release|ARM
+		Release|ARM64 = Release|ARM64
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
```
invertego, line 19: The x86/x64 solution platforms should either build x86/x64 or they shouldn't exist. Right now, they are building the ARM project platform.

tests/impl.cpp
```diff
@@ -11797,8 +11829,15 @@ result_t test_mm_set_denormals_zero_mode(const SSE2NEONTestImpl &impl,
 result_t test_rdtsc(const SSE2NEONTestImpl &impl, uint32_t iter)
 {
     uint64_t start = _rdtsc();
-    for (int i = 0; i < 100000; i++)
+    for (int i = 0; i < 100000; i++) {
+#if defined(_MSC_VER)
+        __nop();
```
invertego, line 11834: These nops are left over from me messing around with preventing the optimizer from eliminating this loop. The _ReadWriteBarrier does the trick and is probably the most robust; the nops can go.

sse2neon.h
```diff
@@ -2745,12 +2803,22 @@ FORCE_INLINE __m128i _mm_undefined_si128(void)
 #if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wuninitialized"
+#endif
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable : 4700)
```
invertego, line 2809: This doesn't actually work, so I gave up and just used setzero below. However, it looks like it *does* work if you move the #pragma above the function declaration.

The other thing to watch out for is runtime uninitialized variable checks. If __MSVC_RUNTIME_CHECKS is defined, the setzero needs to be there or those checks will cause a crash at runtime.

tests/impl.cpp
```diff
@@ -3139,12 +3147,19 @@ result_t test_mm_ucomineq_ss(const SSE2NEONTestImpl &impl, uint32_t iter)
     return test_mm_comineq_ss(impl, iter);
 }
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable : 4700)
```
invertego, line 3152: Not actually needed; see comment _mm_undefined_si128().

tests/impl.cpp
```diff
@@ -6752,6 +6767,10 @@ result_t test_mm_ucomineq_sd(const SSE2NEONTestImpl &impl, uint32_t iter)
     return test_mm_comineq_sd(impl, iter);
 }
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable : 4700)
```
invertego, line 6772: Not actually needed; see comment _mm_undefined_si128().

tests/impl.cpp
```diff
@@ -2642,8 +2646,12 @@ result_t test_mm_rsqrt_ss(const SSE2NEONTestImpl &impl, uint32_t iter)
     __m128 c = _mm_rsqrt_ss(a);
 
     // Here, we ensure the error rate of "_mm_rsqrt_ps()" is under 0.1% compared
-    // to the C implementation.
+    // to the C implementation on non-MSVC compilers, and under 1% on MSVC.
```
invertego, line 2649: Why is this necessary?

sse2neon.h
```diff
@@ -8828,9 +8917,14 @@ FORCE_INLINE __m128i _mm_aeskeygenassist_si128(__m128i a, const int rcon)
 // for more details.
 FORCE_INLINE __m128i _mm_aesenc_si128(__m128i a, __m128i b)
 {
+#ifdef _MSC_VER
```
invertego, line 8920: This shouldn't be strictly necessary. It should be possible to tweak the code in the #else block slightly to work with MSVC by avoiding GNU extensions (e.g. replacing operator ^ with veorq_u8).

sse2neon.h
```diff
@@ -81,7 +82,7 @@
 #define _sse2neon_likely(x) __builtin_expect(!!(x), 1)
 #define _sse2neon_unlikely(x) __builtin_expect(!!(x), 0)
 #else /* non-GNU / non-clang compilers */
-#warning "Macro name collisions may happen with unsupported compiler."
+#pragma message("Macro name collisions may happen with unsupported compiler.")
```
invertego, line 85: Probably don't actually want to print this with MSVC.

sse2neon.h
```diff
@@ -6018,6 +6090,7 @@ FORCE_INLINE __m64 _mm_abs_pi8(__m64 a)
 // Concatenate 16-byte blocks in a and b into a 32-byte temporary result, shift
 // the result right by imm8 bytes, and store the low 16 bytes in dst.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_alignr_epi8
+#if defined(__GNUC__) && !defined(__clang__)
```
invertego, line 6093: Why does GCC need special treatment here?

sse2neon.h
```diff
@@ -112,18 +113,60 @@
...
+
+#define _sse2neon_init(...) \
+    {                       \
+        __VA_ARGS__         \
+    }
+
 /* Compiler barrier */
+#if defined(_MSC_VER)
+#define SSE2NEON_BARRIER()                     \
+    __dmb(_ARM64_BARRIER_ISHST);               \
```
invertego, line 161: The equivalent thing here would be a compiler barrier, e.g. _ReadWriteBarrier().

sse2neon.h
```diff
@@ -2442,7 +2499,11 @@ FORCE_INLINE __m128 _mm_setzero_ps(void)
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_sfence
 FORCE_INLINE void _mm_sfence(void)
 {
+#if defined(_MSC_VER)
+    __dmb(_ARM64_BARRIER_SY);
```
invertego, line 2503: I don't know if this actually the best choice. This just happens to be how the MemoryBarrier() macro in winnt.h is defined.

To reproduce what _sse2neon_smp_mb() does, 
 the barrier type would be _ARM64_BARRIER_ISH. Again, I don't know if this is actually the best choice.

sse2neon.h
```diff
@@ -2442,7 +2499,11 @@ FORCE_INLINE __m128 _mm_setzero_ps(void)
...
+#else
     _sse2neon_smp_mb();
```
invertego, line 2505: Leaving a comment on this line because GitHub doesn't let you comment on unchanged lines...

This function is used by _mm_mfence() and _mm_lfence() below. MSVC support for stdatomic.h is still experimental and __STDC_NO_ATOMICS__ is still defined.
https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/

Furthermore, __STDC_VERSION__ isn't defined at all when compiling as C++.

In both cases, _sse2neon_smp_mb() falls through to the empty "FIXME: MSVC support" else block. At a minimum, there should be a MSVC case in _sse2neon_smp_mb(). that emits the same code GCC/Clang does. Maybe these functions ought to be using more precise barriers, though.

invertego, line 2505: Leaving a comment on this line because GitHub only lets you comment on the lines immediately surrounding changed lines...

This function is used by _mm_mfence() and _mm_lfence() below. MSVC support for stdatomic.h is still experimental and STDC_NO_ATOMICS is still defined.
https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/

Furthermore, STDC_VERSION isn't defined at all when compiling as C++.

In both cases, _sse2neon_smp_mb() falls through to the empty "FIXME: MSVC support" else block. At a minimum, there should be a MSVC case in _sse2neon_smp_mb(). that emits the same code GCC/Clang does. Maybe these functions ought to be using more precise barriers, though.

sse2neon.sln
anthony-linaro: Correct - I deleted them, but these appear to have been kept for some reason by VS - will manually remove

tests/impl.cpp
anthony-linaro: ACK, will remove next commit

sse2neon.h
```diff
@@ -4706,7 +4783,12 @@ FORCE_INLINE __m128i _mm_packus_epi16(const __m128i a, const __m128i b)
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_pause
 FORCE_INLINE void _mm_pause()
 {
+#if defined(_MSC_VER)
+    __dmb(_ARM64_BARRIER_ISHST);
```
invertego, line 4787: This dmb + yield combo is based on the definition of YieldProcessor() in winnt.h. I can make no guarantees as to the wisdom this implementation choice. Plainly it does not match the inline assembly below (but I suspect that's not ideal either...).

tests/impl.cpp
anthony-linaro: This test failed with the tighter tolerance with MSVC - I just went back to the old method from a few months ago which was 1%

sse2neon.h
```diff
@@ -2177,6 +2232,10 @@ FORCE_INLINE __m128 _mm_or_ps(__m128 a, __m128 b)
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_prefetch
 FORCE_INLINE void _mm_prefetch(char const *p, int i)
 {
+    (void) i;
+#if defined(_MSC_VER)
+    __prefetch(p);
```
invertego, line 2237: This emits `prfm PLDL1KEEP,[x0]`, which seems to be the equivalent of `__builtin_prefetch(p, 0, 3)`. There may be MSVC intrinsics to get the other types of prefetch instructions but I don't think I ever looked into it.

sse2neon.h
anthony-linaro: Honestly, I think I'm okay with setzero in all cases - avoids the headaches with `__MSVC_RUNTIME_CHECKS`

sse2neon.h
anthony-linaro: ACK, fixed in next commit

sse2neon.vcxproj
```diff
@@ -0,0 +1,151 @@
...
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
+    <ClCompile>
+      <WarningLevel>Level3</WarningLevel>
+      <SDLCheck>true</SDLCheck>
+      <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <ConformanceMode>true</ConformanceMode>
+      <AdditionalIncludeDirectories>.;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalOptions>/Zc:preprocessor</AdditionalOptions>
```
invertego, line 110: I'm assuming `/Zc:preprocessor` is why the intrinsic implementations that heavily abuse the preprocessor did not require any changes. Because this is not a default compiler flag and sse2neon is a header-only library, I think we need to check _MSVC_TRADITIONAL and either A) disable the stuff that doesn't work with the legacy preprocessor or B) outright refuse to compile at until user sets the flag.

https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170

sse2neon.h
anthony-linaro: IIRC (I didn't actually write this particular bit, someone else did), the updated version didn't work in GCC

sse2neon.h
```diff
@@ -3026,7 +3104,7 @@ FORCE_INLINE __m128i _mm_castps_si128(__m128 a)
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_castsi128_pd
 FORCE_INLINE __m128d _mm_castsi128_pd(__m128i a)
 {
-#if defined(__aarch64__)
+#if defined(__aarch64__) || defined(_M_ARM64)
     return vreinterpretq_m128d_f64(vreinterpretq_f64_m128i(a));
 #else
     return vreinterpretq_m128d_f32(vreinterpretq_f32_m128i(a));
```
invertego, line 3110: Again, I'm placing a comment on an unrelated line due to GitHub limitations...

Just below this, _mm_clflush() is currently doing nothing on MSVC (with another FIXME: MSVC comment).

sse2neon.h
anthony-linaro: I'll swap to _ARM64_BARRIER_ISH to maintain consistency - if it needs to be changed for both implementations, that can be a different discussion elsewhere

sse2neon.h
anthony-linaro: Looks like there is a `__prefetch2` https://learn.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170

Although to quote the doc:
```However, from the C++ language point of view, the function has no observable effect, and may do nothing at all.```

So is this worth it at all?

Comment (invertego) [reply sse2neon.h:2237]: __prefetch2 seems to be it. A few macros exist in winnt.h that demonstrate using it for different types of prefetches.

File: sse2neon.vcxproj (hunk 15 above)
Comment (anthony-linaro) [reply sse2neon.vcxproj:110]: I decided with option B - MSVC WoA support is patchy prior to 2019 anyway, so most people should use a compiler compatible with this flag

File: sse2neon.h (hunk 16 above)
Comment (anthony-linaro) [reply sse2neon.h:3110]: ACK, will use the implementation from `compiler-rt`: https://codebrowser.dev/llvm/compiler-rt/lib/builtins/clear_cache.c.html#66

File: sse2neon.h [hunk 17]
```diff
@@ -2468,7 +2529,7 @@ FORCE_INLINE void _mm_lfence(void)
 
 // FORCE_INLINE __m128 _mm_shuffle_ps(__m128 a, __m128 b, __constrange(0,255)
 // int imm)
-#ifdef _sse2neon_shuffle
+#if defined(_sse2neon_shuffle) && defined(SSE2NEON_STATEMENT_EXPRESSION)
```
invertego, line 2532: I added the `defined(SSE2NEON_STATEMENT_EXPRESSION)` checks so I could force compilers down the other path for testing. These checks should be deleted now, as nothing is defining SSE2NEON_STATEMENT_EXPRESSION. I think we definitely want to keep these separate implementations for compilers that support __builtin_shuffle(vector).

sse2neon.h
```diff
@@ -112,18 +113,60 @@
 /* If using MSVC */
```
invertego, line 113: My original change had this check:

```
#if defined(__cplusplus)
#else
#error sse2neon only supports C++ compilation with this compiler
#endif
```

It'd still be good to have this message so users aren't hit with random compiler errors when compiling as C with MSVC.

sse2neon.h
anthony-linaro: ACK, will remove

sse2neon.h
```diff
@@ -112,18 +113,60 @@
...
+#include <malloc.h>
+#endif
+
 #if (defined(_M_AMD64) || defined(__x86_64__)) || \
     (defined(_M_ARM64) || defined(__arm64__))
 #define SSE2NEON_HAS_BITSCAN64
 #endif
 #endif
 
+#if defined(__GNUC__) || defined(__clang__) || 0
```
invertego, line 127: `|| 0` leftover from testing

sse2neon.h
```diff
@@ -112,18 +113,60 @@
...
+        body                             \
+    })
+#define _sse2neon_define2(type, a, b, body) \
+    __extension__({                         \
+        type _a = (a), _b = (b);            \
+        body                                \
+    })
+#define _sse2neon_return(ret) (ret)
+#else
+// #define _sse2neon_define(type, a, body) [_a = (a)](){body}()
```
invertego, line 145: Dead code

sse2neon.vcxproj
invertego: Something to consider: the new preprocessor can break existing code that worked with the old one. Users may specifically want to avoid turning it on.

anthony-linaro: Understood, but I think the benefits outweigh the risks here - we are in luck that WoA is a relatively young platform, so might not have acquired quite as much technical debt as other platforms.

A proposal could be to create an issue to keep track of of anyone is affected by this

sse2neon.h
```diff
@@ -8339,6 +8420,8 @@ FORCE_INLINE uint64_t _mm_crc32_u64(uint64_t crc, uint64_t v)
     __asm__ __volatile__("crc32cx %w[c], %w[c], %x[v]\n\t"
                          : [c] "+r"(crc)
                          : [v] "r"(v));
+#elif defined(_M_ARM64)
+    crc = __crc32cd(crc, v);
```
invertego, line 8424: The first parameter should be cast to uint32_t to avoid the following warning:
`warning C4244: 'argument': conversion from 'uint64_t' to 'unsigned int', possible loss of data`

sse2neon.h
```diff
@@ -111,19 +116,64 @@
 
 /* If using MSVC */
 #ifdef _MSC_VER
+#include <windows.h>
```
invertego, line 119: If/when to include windows.h is a decision best left to the application. Including it here creates headaches for the project that was my whole motivation for looking at sse2neon in the first place ([ares-emulator/ares](https://github.com/ares-emulator/ares)). It avoids including windows.h in most translation units because doing so pollutes the global namespace with names that clash with application code.

The macros used in _mm_prefetch() could easily be replaced with constants. FlushInstructionCache() is the real problem case. I see three options:

1. Reproduce the FlushInstructionCache() declaration here. Ugly, a bit evil, but unlikely to cause issues in practice.
2. Guard this behind something the application must #define to opt in. I suspect most users would never even notice the absence of _mm_clflush().
3. Leave it like this and force users to deal with the side effects of including windows.h.

Just speaking personally, 3 is my least favorite option.

sse2neon.sln
invertego: There are still some vestigial x86/x64 lines under postSolution in the latest commit, though they are likely harmless.

sse2neon.vcxproj
invertego: I don't feel *that* strongly about this, but I brought it up because sse2neon is aimed at users porting existing code to ARM and not users writing new applications. One could argue the very purpose of this library is to stave off paying down accumulated technical debt. ;)

sse2neon.h
invertego: Still present in the latest commit

sse2neon.h
anthony-linaro: Option 1 feels a bit rough to me - duplicating Win API functions feels icky. Option 3 I can see your point, so I suppose this leaves us with option 2 - will implement next commit

anthony-linaro: Option 2 implemented - does that look acceptable?

sse2neon.h
```diff
@@ -4950,7 +5040,7 @@ FORCE_INLINE __m128i _mm_setzero_si128(void)
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_shuffle_epi32
 // FORCE_INLINE __m128i _mm_shuffle_epi32(__m128i a,
 //                                        __constrange(0,255) int imm)
-#ifdef _sse2neon_shuffle
+#if defined(_sse2neon_shuffle) && defined(SSE2NEON_STATEMENT_EXPRESSION)
```
invertego, line 5043: Still an instance of `&& defined(SSE2NEON_STATEMENT_EXPRESSION)` to be removed here and two more below.

sse2neon.h
invertego: Fine by me!

invertego:
It's not a huge deal, it'd be nice if this built cleanly at /W4. Here's what's remains:

```
sse2neon.h(6468,9): warning C4310: cast truncates constant value
sse2neon.h(8258,39): warning C4100: 'a': unreferenced formal parameter
sse2neon.h(8260,39): warning C4100: 'b': unreferenced formal parameter
sse2neon.h(8261,35): warning C4100: 'lb': unreferenced formal parameter
sse2neon.h(8271,39): warning C4100: 'a': unreferenced formal parameter
sse2neon.h(8272,35): warning C4100: 'la': unreferenced formal parameter
sse2neon.h(8273,39): warning C4100: 'b': unreferenced formal parameter
sse2neon.h(8356,50): warning C4100: 'b': unreferenced formal parameter
sse2neon.h(8367,39): warning C4100: 'a': unreferenced formal parameter

**Comment by anthony-linaro** [comment]
@jserv Done!

**Comment by jserv** [comment]
The proposed change caused MinGW breakage:
```shell
$ aarch64-w64-mingw32-clang++ -o tests/common.o -Wall -Wcast-qual -I. -march=armv8-a+fp+simd -std=gnu++14 -c -MMD -MF tests/common.o.d tests/common.cpp
In file included from tests/common.cpp:1:
In file included from tests/common.h:5:
./sse2neon.h:838[9](https://github.com/DLTcollab/sse2neon/actions/runs/5046362677/jobs/9063430352?pr=596#step:4:10):[11](https://github.com/DLTcollab/sse2neon/actions/runs/5046362677/jobs/9063430352?pr=596#step:4:12): error: use of undeclared identifier '__crc32ch'
    crc = __crc32ch(crc, v);
```

@invertego and @anthony-linaro, can you check?

invertego:
This change assumes the availability of CRC and AES intrinsics on aarch64, but clang's arm_neon.h only declares them if the requisite architectural features are enabled. Adding "+crc+aes" to ARCH_CFLAGS in the Makefile will make these errors go away. I suppose that's sufficient so long as we don't actually care about building sse2neon (specifically for Windows) with these features disabled. Otherwise, it will be necessary to check __clang__ in addition to _M_ARM64 to prevent this assumption from applying to clang.

anthony-linaro:
As far as I can tell (I did add it in a comment) there does not exist a
supported WoA platform without crypto extensions (people have it running on
some unsupported ones like old Windows Phones such as the Lumia 950, but
those are rare/unsupported, and we can discount them).

I will modify the makefile for Windows as suggested.

anthony-linaro:
@jserv This should now (hopefully) be resolved

jserv:
> @jserv This should now (hopefully) be resolved

It seems that crypto extension was not properly handled.
```shell
$ aarch64-w64-mingw32-clang++ -o tests/common.o -Wall -Wcast-qual -I. -march=armv8-a+fp+simd -std=gnu++14 -c -MMD -MF tests/common.o.d tests/common.cpp
In file included from tests/common.cpp:1:
In file included from tests/common.h:5:
./sse2neon.h:8[9](https://github.com/DLTcollab/sse2neon/actions/runs/5055092682/jobs/9070866019?pr=596#step:4:10)16:20: error: use of undeclared identifier 'vaeseq_u8'
        vaesmcq_u8(vaeseq_u8(vreinterpretq_u8_m[12](https://github.com/DLTcollab/sse2neon/actions/runs/5055092682/jobs/9070866019?pr=596#step:4:13)8i(a), vdupq_n_u8(0))),
                   ^
```

jserv:
Thank @anthony-linaro and @invertego for this great effort!

PR #597, opened 2023-05-18
Remove Kahan algorithm in `_mm_dp_ps`

### LINKED ISSUE #595: _mm_dp_ps does not always match x86_64
The `_mm_dp_ps` implementation  will give a perfect match for 0xFF and 0x7F modes, but the other modes will use the Kahan algorithm to get a little bit more precision. This means that changing which lanes you write the result to can produce slightly different results, which is likely unwanted and surprising. It also won't match what x86_64 produces.

Note that the Intel® SSE4 Programming Reference states that this instruction should produce the same result that you get from a standard non-Kahan algorithm implementation.
> Dot Product operations are not specified in IEEE-754. When neither FTZ nor DAZ are enabled, the dot product instructions resemble sequences of IEEE-754 multiplies and adds (with rounding at each stage)

Either this enhanced precision code path should be removed (my preference and it also simplifies the code) or made optional using an `SSE2NEON_PRECISE_*` define.

Cuda-Chen:
Remove Kahan algorithm in `_mm_dp_ps` to align conversion result with SSE.

Also, apply shortcut when immediate is `0xXF` and target is ARMv8-A.

Last, add more tests for testing possible combinations of immediate
including `0xXF`.

Close #595.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -779,16 +779,6 @@ FORCE_INLINE __m128 _mm_shuffle_ps_2032(__m128 a, __m128 b)
     return vreinterpretq_m128_f32(vcombine_f32(a32, b20));
 }
 
-// Kahan summation for accurate summation of floating-point numbers.
-// http://blog.zachbjornson.com/2019/08/11/fast-float-summation.html
-FORCE_INLINE void _sse2neon_kadd_f32(float *sum, float *c, float y)
-{
-    y -= *c;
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Consolidate the reason to drop Kahan algorithm in git commit messages.
Check https://cbea.ms/git-commit/ for general rules.

jserv:

PR #598, opened 2023-06-06
Add test running for MSVC to CI

anthony-linaro:
This allows tests to run for Windows MSVC ARM64 platforms


jserv:
Thank @anthony-linaro for contributing!

PR #609, opened 2023-09-06
Fix build for gcc-13 and 32 bit arm systems.

balister:
* gcc-13 still does not support vld1q_u8_x4 for 32 bit arm builds.


jserv:
Thank @balister for contributing!

Issue #571, opened 2022-12-26
Relaxed type casting


aqrit:
From https://github.com/embree/embree/issues/410 

I haven't checked with a compiler but these lines (from 72daa0f3204dd5b5c8506a4295d92ab87228d471) seem lax:

```
sse2neon.h:580:    return vget_lane_u32((uint32x2_t) o, 0); /// uint16_t
sse2neon.h:2486:    return vsetq_lane_f32(vgetq_lane_f32(_mm_rsqrt_ps(in), 0), in, 0); /// __m128
sse2neon.h:3151:    return vld1q_f32((float32_t *) c); /// __m128d
sse2neon.h:3173:    return vld1q_f32((float32_t *) c); /// __m128d
sse2neon.h:3303:    return (__m128i) vrhaddq_u16(vreinterpretq_u16_m128i(a), /// C-style cast
sse2neon.h:4540:    return vld1q_f32((float32_t *) c); /// __m128d
sse2neon.h:5025:    return vgetq_lane_u64(high_bits, 0) | (vgetq_lane_u64(high_bits, 1) << 1); /// int
sse2neon.h:5078:    return vld1q_f32((float32_t *) c); /// __m128d
sse2neon.h:5324:    return (__m128i) vld1q_s8(data); /// C-style cast
sse2neon.h:5481:    return (__m128i) vld1q_s8(data); /// C-style cast
sse2neon.h:5868:    return (__m128i) vshlq_s16((int16x8_t) a, vdupq_n_s16(-count)); /// C-style cast
sse2neon.h:6306:    return vld1q_f32((float32_t *) c); /// __m128d
sse2neon.h:7011:    return vreinterpretq_s64_s16( /// __m128i
sse2neon.h:7034:    return vreinterpret_s64_s16(vqadd_s16(vuzp1_s16(a, b), vuzp2_s16(a, b))); /// __m64
sse2neon.h:7037:    return vreinterpret_s64_s16(vqadd_s16(res.val[0], res.val[1])); /// __m64
```

jserv:
Recent `embree` enforce `-flax-vector-conversions` option which allows implicit conversions between vectors with differing numbers of elements and/or incompatible element types. According to [GCC manual](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html), this option should not be used for new code.

jserv:
Duplicated with #614

PR #614, opened 2023-10-04
disambiguate vector type to avoid errors depending on lax conversion …

JoachimSchurig:
…settings

Typecast explicitly to __m128d * to avoid compiler errors depending on platform and compiler, and the settings of -flax-vector-conversions or its different defaults

The test suite will fail to compile on AppleClang/aarch64 with at least AppleClang 14 and 15 if you add the compiler option -flax-vector-conversions to the makefile. There are projects that include sse2neon.h that have this option set.

Also, the option has differing defaults depending on compiler make and version, therefore it makes sense to not depend on it for the code to be compiled.

See discussions here: [llvm](https://reviews.llvm.org/D67678)  and here [gcc](https://gcc.gnu.org/bugzilla//show_bug.cgi?id=88698)

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -5534,7 +5534,7 @@ FORCE_INLINE void _mm_storeu_si32(void *p, __m128i a)
 FORCE_INLINE void _mm_stream_pd(double *p, __m128d a)
 {
 #if __has_builtin(__builtin_nontemporal_store)
-    __builtin_nontemporal_store(a, (float32x4_t *) p);
+    __builtin_nontemporal_store(a, (__m128d *) p);
 #elif defined(__aarch64__) || defined(_M_ARM64)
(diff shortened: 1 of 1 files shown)


jserv:
Thank @JoachimSchurig for contributing. Can you check #571 as well?

JoachimSchurig:
Thank _you_ for this wonderful project!

As for #571, all tests compile and pass on AppleClang when compiled with `-flax-vector-conversions` or `-flax-vector-conversions=none` or `-flax-vector-conversions=integer` or `-fno-lax-vector-conversions` (or without any of those).
I think the difference to the above is that they operate on value types and not on pointers, and not through the `__builtin_nontemporal_store` builtin. That may trigger the error at `_mm_stream_pd`.

PR #615, opened 2023-10-06
Introduce fast and deterministic RNG

### LINKED ISSUE #581: Generate more random test inputs
The current test inputs are listed as following.
```c
    float *mTestFloatPointer1;
    float *mTestFloatPointer2;
    int32_t *mTestIntPointer1;
    int32_t *mTestIntPointer2;
```
It seems the generated numbers are not quite random, which may reduce the coverage of tests.

Cuda-Chen:
Use SplitMix64 for fast and deterministic RNG for cross compiler/platform support.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -350,10 +350,27 @@ static inline double bankersRounding(double val)
     return ret;
 }
 
-static float ranf(void)
+// SplitMix64 PRNG by Sebastiano Vigna, see:
+// <https://xoshiro.di.unimi.it/splitmix64.c>
+static uint64_t state;  // the state of SplitMix64 PRNG
+const double TWOPOWER64 = pow(2, 64);
+
+#define SSE2NEON_INIT_RNG(seed) \
+    do {                        \
+        state = seed;           \
+    } while (0)
+
+static double next()
 {
-    uint32_t ir = rand() & 0x7FFF;
-    return (float) ir * (1.0f / 32768.0f);
+    uint64_t z = (state += 0x9e3779b97f4a7c15);
+    z = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9;
+    z = (z ^ (z >> 27)) * 0x94d049bb133111eb;
+    return z ^ (z >> 31);
+}
+
+static float ranf()
+{
+    return next() / TWOPOWER64;
 }
 
 static float ranf(float low, float high)
@@ -11796,7 +11813,7 @@ SSE2NEONTestImpl::SSE2NEONTestImpl(void)
     mTestFloatPointer2 = (float *) platformAlignedAlloc(sizeof(__m128));
     mTestIntPointer1 = (int32_t *) platformAlignedAlloc(sizeof(__m128i));
     mTestIntPointer2 = (int32_t *) platformAlignedAlloc(sizeof(__m128i));
-    srand(0);
+    SSE2NEON_INIT_RNG(123456);
     for (uint32_t i = 0; i < MAX_TEST_VALUE; i++) {
         mTestFloats[i] = ranf(-100000, 100000);
         mTestInts[i] = (int32_t) ranf(-100000, 100000);


tests/impl.cpp
```diff
@@ -4059,6 +4083,11 @@ result_t test_mm_cmplt_sd(const SSE2NEONTestImpl &impl, uint32_t iter)
     __m128d b = load_m128d(_b);
     __m128d c = _mm_cmplt_sd(a, b);
 
+    /*double res[2];
```
jserv, line 4086: Don't leave the dead code.

Cuda-Chen: Done.

tests/impl.cpp
```diff
@@ -350,10 +350,34 @@ static inline double bankersRounding(double val)
     return ret;
 }
 
-static float ranf(void)
-{
-    uint32_t ir = rand() & 0x7FFF;
-    return (float) ir * (1.0f / 32768.0f);
+/*  Written in 2015 by Sebastiano Vigna (vigna@acm.org)
```
jserv, line 353: Since the code is released as public domain, you can simply skip the copyright notice. Simply mention the name as https://github.com/microsoft/mimalloc/blob/master/include/mimalloc/internal.h#L768

Cuda-Chen: Done.

tests/impl.cpp
```diff
@@ -350,10 +350,21 @@ static inline double bankersRounding(double val)
     return ret;
 }
 
-static float ranf(void)
+// by Sebastiano Vigna, see: <https://xoshiro.di.unimi.it/splitmix64.c>
+static uint64_t x;
+const double TWOPOWER64 = pow(2, 64);
+
+static double next()
```
jserv, line 357: How do we specify the initial value?

tests/impl.cpp
```diff
@@ -11796,7 +11807,7 @@ SSE2NEONTestImpl::SSE2NEONTestImpl(void)
     mTestFloatPointer2 = (float *) platformAlignedAlloc(sizeof(__m128));
     mTestIntPointer1 = (int32_t *) platformAlignedAlloc(sizeof(__m128i));
     mTestIntPointer2 = (int32_t *) platformAlignedAlloc(sizeof(__m128i));
-    srand(0);
+    x = 123456;
```
jserv, line 11810: Avoid such risky operation, which might be removed by other developers accidentally. Introduce a helper function instead.

tests/impl.cpp
```diff
@@ -350,10 +350,26 @@ static inline double bankersRounding(double val)
     return ret;
 }
 
-static float ranf(void)
+// by Sebastiano Vigna, see: <https://xoshiro.di.unimi.it/splitmix64.c>
+static uint64_t x;
+const double TWOPOWER64 = pow(2, 64);
+
+#define SSE2NEON_INIT_TEST_CASES_GENERATOR(seed) \
```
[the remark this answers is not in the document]
Cuda-Chen: How about use `state` to represent the seed and `SSE2NEON_INIT_STATE` to give the seed to SplitMix64?
For example:

```cpp
static uint64_t state; // the state of SplitMix64

#define SSE2NEON_INIT_STATE(seed) state = seed
```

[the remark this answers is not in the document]
jserv: Why not `SSE2NEON_INIT_RNG`?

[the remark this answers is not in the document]
Cuda-Chen: > Why not `SSE2NEON_INIT_RNG`?

That sounds great. I will apply this in next commit.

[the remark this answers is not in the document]
Cuda-Chen: Done.

tests/impl.cpp
Cuda-Chen: Done.

howjmay:
I am curious. Doesn't it still generate the same random number sequence?
I concerned about we may not cover all the possible values with the previous RNG function.
But maybe having a consistent random number sequence is better?

Cuda-Chen:
> I am curious. Doesn't it still generate the same random number sequence? I concerned about we may not cover all the possible values with the previous RNG function. But maybe having a consistent random number sequence is better?

After searching, I realize that "consistent random number sequence" means the RNG always produces the same random number sequence every time the test program is run. Is that what we are going to discuss about?

howjmay:
> > I am curious. Doesn't it still generate the same random number sequence? I concerned about we may not cover all the possible values with the previous RNG function. But maybe having a consistent random number sequence is better?
> 
> After searching, I realize that "consistent random number sequence" means the RNG always produces the same random number sequence every time the test program is run. Is that what we are going to discuss about?

Yes that is exactly what I mentioned. So do we need to have a consistent random number sequence or maybe not?

jserv:
> > After searching, I realize that "consistent random number sequence" means the RNG always produces the same random number sequence every time the test program is run. Is that what we are going to discuss about?
> 
> Yes that is exactly what I mentioned. So do we need to have a consistent random number sequence or maybe not?

For the purpose of CI pipeline, we shall stick to deterministic random number generation.

howjmay:
> > > After searching, I realize that "consistent random number sequence" means the RNG always produces the same random number sequence every time the test program is run. Is that what we are going to discuss about?
> > 
> > 
> > Yes that is exactly what I mentioned. So do we need to have a consistent random number sequence or maybe not?
> 
> For the purpose of CI pipeline, we shall stick to deterministic random number generation.

Got it. 
Then I am curious the purpose of this PR. The old implementation has already presented a deterministic random number generation. Maybe we should just close the original issue?

Cuda-Chen:
> > > > After searching, I realize that "consistent random number sequence" means the RNG always produces the same random number sequence every time the test program is run. Is that what we are going to discuss about?
> > > 
> > > 
> > > Yes that is exactly what I mentioned. So do we need to have a consistent random number sequence or maybe not?
> > 
> > 
> > For the purpose of CI pipeline, we shall stick to deterministic random number generation.
> 
> Got it. Then I am curious the purpose of this PR. The old implementation has already presented a deterministic random number generation. Maybe we should just close the original issue?

I agree. Just close the original issue with a comment mentioning about sticking to deterministic random number generation of CI pipelines.

jserv (changes requested):
Refine the subject per discussions.

howjmay:
issue was closed

Cuda-Chen:
As the `rand()` usually acts good for deterministic random number generation, I would like to close this PR if we do not need fast random number generation.

jserv:
> As the `rand()` usually acts good for deterministic random number generation, I would like to close this PR if we do not need fast random number generation.

The proposed RNG still makes sense for cross-platform consideration. In particular, this project supports GCC, Clang, and MSVC in combination with Linux, macOS, and Windows. Therefore, a fast and deterministic RNG can minimize potential issues while performing tests.

@Cuda-Chen, please refine the commit messages.

PR #620, opened 2023-11-22
Add A32 support in CI

Cuda-Chen:
Add A32 support in CI.

For @jonathanhue, if possible, kindly leave some suggestions as I have little experience on A32 binary running on Armv8-A.

### DIFF
--- a/.github/workflows/github_actions.yml
+++ b/.github/workflows/github_actions.yml
@@ -55,7 +55,12 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        arch_with_features: [{arch: armv7, feature: none}, {arch: aarch64, feature: none}, {arch: aarch64, feature: crypto+crc}]
+        arch_with_features: [
+          {arch: armv7, feature: none, arch_cflags: none},
+          {arch: aarch64, feature: none, arch_cflags: none},
+          {arch: aarch64, feature: crypto+crc, arch_cflags: none},
+          {arch: armv7, feature: none, arch_cflags: '-mcpu=cortex-a32 -mfpu=neon-fp-armv8'}
+        ]
         cxx_compiler: [g++-10, clang++-11]
     steps:
       - name: checkout code
@@ -69,6 +74,7 @@ jobs:
           distro: ubuntu20.04
           env: |
             CXX: ${{ matrix.cxx_compiler }}
+            ARCH_CFLAGS: ${{ matrix.arch_with_features.arch_cflags }}
           install: |
             apt-get update -q -y
             apt-get install -q -y "${{ matrix.cxx_compiler }}" make
--- a/Makefile
+++ b/Makefile
@@ -29,14 +29,24 @@ EXEC_WRAPPER = qemu-$(processor)
 endif
 
 # Follow platform-specific configurations
-ifeq ($(processor),$(filter $(processor),aarch64 arm64))
-    ARCH_CFLAGS = -march=armv8-a+fp+simd
-else ifeq ($(processor),$(filter $(processor),i386 x86_64))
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -34,7 +34,11 @@ ifeq ($(processor),$(filter $(processor),aarch64 arm64))
 else ifeq ($(processor),$(filter $(processor),i386 x86_64))
     ARCH_CFLAGS = -maes -mpclmul -mssse3 -msse4.2
 else ifeq ($(processor),$(filter $(processor),arm armv7 armv7l))
-    ARCH_CFLAGS = -mfpu=neon
+	ifdef A32
+		ARCH_CFLAGS = -mcpu=cortex-a32 -mfpu=neon-fp-armv8
```
jserv, line 38: Enforce consistent indention.

README.md
```diff
@@ -126,6 +126,10 @@ or
 ```shell
 $ make CROSS_COMPILE=arm-linux-gnueabihf- check # ARMv7-A
 ```
+or
+``` shell
+$ make CROSS_COMPILE=arm-linux-gnueabihf- A32=1 check # ARMv8-A running in 32-bit mode, i.e., A32 instruction set
```
jserv, line 131: In this case, A32 is always 1, right?

Cuda-Chen: Correct. A32 always sets to 1 if we want to run test program compiled under A32 instruction set.

Makefile
Cuda-Chen: Fixed.

README.md
jserv: If toolchain is set to `arm-linux-gnueabihf-`, `A32` should not appear since the former implies the latter.

Cuda-Chen: The goal of using `A32` for indicating A32 instruction set build is that we can not only test our code on AArch32 but also check whether the NEON optimization on A32 will create any impacts or not.

Nevertheless, we may set the `-mcpu` flag in other Makefile input parameters such as `CFLAGS`, and I consider I can start to implement from this.

README.md
```diff
@@ -126,6 +126,13 @@ or
 ```shell
 $ make CROSS_COMPILE=arm-linux-gnueabihf- check # ARMv7-A
 ```
+or
+``` shell
+$ make \
+  CROSS_COMPILE=arm-linux-gnueabihf- \
+  ARCH_CFLAGS="-mcpu=cortex-a32 -mfpu=neon-fp-armv8" \
+  check # ARMv8-A running in 32-bit mode, i.e., A32 instruction set
```
jserv, line 134: Move the comment out of code block.

.github/workflows/github_actions.yml
```diff
@@ -55,7 +55,12 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        arch_with_features: [{arch: armv7, feature: none}, {arch: aarch64, feature: none}, {arch: aarch64, feature: crypto+crc}]
+        arch_with_features: [
+          {arch: armv7, feature: none, arch_cflags: none},
```
jserv, line 59: At least one feature should be specified. That is `simd` (or NEON), right?

Cuda-Chen: Correct. 

Based on my experiment, `uraimo/run-on-arch-action` does not permit use to expand variables in `env` via any `if-else` expression. As a consequence, I use `none` to let `make` to fill the default value of `ARCH_CFLAGS` compile flags, just like the `FEATURE` variable in Makefile does.

jserv: Add necessary comments to describe the features.

Cuda-Chen: Comments to describe the usage of `ARCH_CLFAGS` are added in README.

jserv:

PR #621, opened 2023-12-02
Fix _mm_test_mix_ones_zeros and _mm_testnzc_si128

aqrit:
Bug: `_mm_test_mix_ones_zeros` always returned true.
The function wasn't reducing `zf` and `cf` to a bool before combining them.

The fix proposed here isn't the most efficient, but at least it is correct. 

` _mm_testnzc_si128` is an alias for `_mm_test_mix_ones_zeros`.

Note(s):
The arguments are named incorrectly in the `_mm_test_mix_ones_zeros` documentation[0].
The second argument is the mask, as per the behavior of `_mm_test_mix_ones_zeros` with gcc and clang.
This naming error seems to have propagated through both gcc[1] and llvm[2] headers but not to rust[3] headers or sse2neon[4].

[0] https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=ptest&techs=SSE_ALL&ig_expand=6902
[1] https://github.com/gcc-mirror/gcc/blob/27ce74fa23c93c1189c301993cd19ea766e6bdb5/gcc/config/i386/smmintrin.h#L94
[2] https://github.com/llvm/llvm-project/blob/70535f5e609f747c28cfef699eefb84581b0aac0/clang/lib/Headers/smmintrin.h#L1130
[3] https://github.com/rust-lang/stdarch/blob/f4528dd6e85d97bb802240d7cd048b6e1bf72540/crates/core_arch/src/x86/sse41.rs#L1149
[4] https://github.com/DLTcollab/sse2neon/blob/243e90f654193c97a691b1a53213d091e02eb631/sse2neon.h#L7595

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -7593,7 +7593,22 @@ FORCE_INLINE int _mm_test_all_zeros(__m128i a, __m128i mask)
 // otherwise return 0.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=mm_test_mix_ones_zero
 // Note: Argument names may be wrong in the Intel intrinsics guide.
-#define _mm_test_mix_ones_zeros(a, mask) _mm_testnzc_si128(a, mask)
+FORCE_INLINE int _mm_test_mix_ones_zeros(__m128i a, __m128i mask)
+{
+    uint64x2_t v = vreinterpretq_u64_m128i(a);
+    uint64x2_t m = vreinterpretq_u64_m128i(mask);
+
+    // find ones (set-bits) and zeros (clear-bits) under clip mask
+    uint64x2_t ones = vandq_u64(m, v);
+    uint64x2_t zeros = vbicq_u64(m, v);
+
+    // If both 128-bit variables are populated (non-zero) then return 1.
+    // For comparision purposes, first compact each var down to 32-bits.
+    uint32x2_t reduced = vpmax_u32(vqmovn_u64(ones), vqmovn_u64(zeros));
+
+    // if folding minimum is non-zero then both vars must be non-zero
+    return (vget_lane_u32(vpmin_u32(reduced, reduced), 0) != 0);
+}
 
 // Compute the bitwise AND of 128 bits (representing integer data) in a and b,
 // and set ZF to 1 if the result is zero, otherwise set ZF to 0. Compute the
@@ -7613,10 +7628,7 @@ FORCE_INLINE int _mm_testc_si128(__m128i a, __m128i b)
 // otherwise set CF to 0. Return 1 if both the ZF and CF values are zero,
 // otherwise return 0.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_testnzc_si128
-FORCE_INLINE int _mm_testnzc_si128(__m128i a, __m128i b)
-{
-    return !(_mm_test_all_zeros(a, b) | _mm_testc_si128(a, b));
-}
+#define _mm_testnzc_si128(a, b) _mm_test_mix_ones_zeros(a, b)
 
 // Compute the bitwise AND of 128 bits (representing integer data) in a and b,
 // and set ZF to 1 if the result is zero, otherwise set ZF to 0. Compute the
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -7592,14 +7592,10 @@ FORCE_INLINE int _mm_test_all_zeros(__m128i a, __m128i mask)
 // zero, otherwise set CF to 0. Return 1 if both the ZF and CF values are zero,
 // otherwise return 0.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=mm_test_mix_ones_zero
+// Note: Argument names may be wrong in the Intel intrinsics guide.
 FORCE_INLINE int _mm_test_mix_ones_zeros(__m128i a, __m128i mask)
 {
-    uint64x2_t zf =
-        vandq_u64(vreinterpretq_u64_m128i(mask), vreinterpretq_u64_m128i(a));
-    uint64x2_t cf =
-        vbicq_u64(vreinterpretq_u64_m128i(mask), vreinterpretq_u64_m128i(a));
-    uint64x2_t result = vandq_u64(zf, cf);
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -7592,14 +7592,10 @@ FORCE_INLINE int _mm_test_all_zeros(__m128i a, __m128i mask)
...
+// Note: Argument names may be wrong in the Intel intrinsics guide.
 FORCE_INLINE int _mm_test_mix_ones_zeros(__m128i a, __m128i mask)
 {
-    uint64x2_t zf =
-        vandq_u64(vreinterpretq_u64_m128i(mask), vreinterpretq_u64_m128i(a));
-    uint64x2_t cf =
-        vbicq_u64(vreinterpretq_u64_m128i(mask), vreinterpretq_u64_m128i(a));
-    uint64x2_t result = vandq_u64(zf, cf);
-    return !(vgetq_lane_u64(result, 0) | vgetq_lane_u64(result, 1));
+    return !(_mm_testz_si128(a, mask) | _mm_testc_si128(a, mask));
```
aqrit, line 7598: Needs a forward declare, or should I just copy and paste these functions into this functions body?

sse2neon.h
```diff
@@ -7592,15 +7592,8 @@ FORCE_INLINE int _mm_test_all_zeros(__m128i a, __m128i mask)
 // zero, otherwise set CF to 0. Return 1 if both the ZF and CF values are zero,
 // otherwise return 0.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=mm_test_mix_ones_zero
-FORCE_INLINE int _mm_test_mix_ones_zeros(__m128i a, __m128i mask)
-{
-    uint64x2_t zf =
-        vandq_u64(vreinterpretq_u64_m128i(mask), vreinterpretq_u64_m128i(a));
-    uint64x2_t cf =
-        vbicq_u64(vreinterpretq_u64_m128i(mask), vreinterpretq_u64_m128i(a));
-    uint64x2_t result = vandq_u64(zf, cf);
```
howjmay, line 7601: I think fixed this line with the following code can solve the issue
```diff
-uint64x2_t result = vandq_u64(zf, cf);
+uint64x2_t result = vreinterpretq_u64_u32(vmvnq_u32(vreinterpretq_u32_u64(vorrq_u64(zf, cf))));
```

tests/impl.cpp
```diff
@@ -9228,14 +9228,15 @@ result_t test_mm_test_mix_ones_zeros(const SSE2NEONTestImpl &impl,
...
-    int32_t d2 = !((_a[2]) & _mask[2]) & !((!_a[2]) & _mask[2]);
-    int32_t d3 = !((_a[3]) & _mask[3]) & !((!_a[3]) & _mask[3]);
-    int32_t result = ((d0 & d1 & d2 & d3) == 0) ? 1 : 0;
+    int32_t has_ones = 0;
+    int32_t has_zeros = 0;
+    for (int i = 0; i < 4; i++) {
+        has_ones |= _a[i] & _mask[i];
+        has_zeros |= ~_a[i] & _mask[i];
+    }
+    int32_t result = ((has_ones != 0) & (has_zeros != 0));
```
howjmay, line 9237: If possible maybe we can add some edge cases here to test the behavior?

sse2neon.h
howjmay: I think we could just expand the implementation or use the fix I suggested in the following

sse2neon.h
aqrit: This suffers from the same problem. We can't `OR`/`AND` zf and cf together, until we've normalized then somehow. I'll see what I can do.

aqrit: something like this for ARMv7-A ?
```
FORCE_INLINE int _mm_test_mix_ones_zeros(__m128i a, __m128i mask)
{
    uint64x2_t v = vreinterpretq_u64_m128i(a);
    uint64x2_t m = vreinterpretq_u64_m128i(mask);
    uint32x2_t has_one = vqmovn_u64(vandq_u64(m, v));
    uint32x2_t has_zero = vqmovn_u64(vbicq_u64(m, v));
    uint32x2_t pack = vpmax_u32(has_one, has_zero);
    uint32x2_t reduced = vpmin_u32(pack, pack);
    return (vget_lane_u32(reduced, 0) != 0);
}
```

aqrit: added. remeber to squash it.

tests/impl.cpp
aqrit: I'm not sure where the values are coming from, I'll leave this to you.

jserv: > If possible maybe we can add some edge cases here to test the behavior?

@howjmay, Can you show relevant code accordingly?

jserv:
@aqrit, Additionally, I would like to acknowledge your significant contributions, including those to [sse2zig](https://github.com/aqrit/sse2zig), in the sse2neon header. Could you please let me know the preferred name or manner in which you'd like to be credited?

sse2neon.h
```diff
@@ -7592,14 +7592,22 @@ FORCE_INLINE int _mm_test_all_zeros(__m128i a, __m128i mask)
 // zero, otherwise set CF to 0. Return 1 if both the ZF and CF values are zero,
 // otherwise return 0.
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=mm_test_mix_ones_zero
+// Note: Argument names may be wrong in the Intel intrinsics guide.
```
howjmay, line 7595: One question about this. Why do you think the argument may be wrong. Just curious about the reason

aqrit: according to the Release notes:

on `01/24/2023` 
`Corrected parameter order of _mm_test_all_zeros and _mm_test_mix_ones_zeros.`

I don't know if this person still maintains this or not:
https://old.reddit.com/r/simd/comments/cck1kh/feedback_on_intel_intrinsics_guide/

I don't have a twitter account so I can't really track them down.

tests/impl.cpp
howjmay: I think if there is a test case all the bits of `mask` is zero would be good. It is a case that force the answer to be zero

sse2neon.h
howjmay: got it!! Thank you for explaining!!!

jserv:
Thank @aqrit for contributing. It is time to make a new release.

PR #623, opened 2024-01-02
Fix Clang showing incorrect GCC version warning

brechtvl:
Clang also defines `__GNUC__` and even in recent Clang versions like 15 this value is 4.


jserv:
Thank @brechtvl for contributing!

PR #626, opened 2024-01-08
Restore options for precision of div/rcp/sqrt/rsqrt

brechtvl:
This restores support for SSE2NEON_PRECISE_DIV and SSE2NEON_PRECISE_SQRT with the same level of precision that they had before deprecation in #580.

That changed enabled higher precision by default, but not as high as was possible with the options before. The options were originally added to make Embree and Blender work correctly on Apple Silicion, and tested to need exactly this level of precision.

---

It would be helpful for Blender not to have to maintain either a patched sse2neon or work with different tolerances on x86 and ARM.


jserv:
Thank @brechtvl for contributing!

Issue #624, opened 2024-01-05
Optimize CRC intrinisics for targets lacking of CRC extension


jserv:
The existing functions `_mm_crc32_u8`, `_mm_crc32_u16`, and `_mm_crc32_u32` are currently executed using generic C programming in the absence of CRC extensions. There is a potential implementation designed for CPUs which are equipped to handle [PMULL](https://developer.arm.com/documentation/ddi0602/2023-12/SIMD-FP-Instructions/PMULL--PMULL2--Polynomial-Multiply-Long-) (Polynomial Multiply Long) instructions but lack support for CRC32 instructions.
 
Reference:
* [libdeflate/lib/arm/crc32_impl.h](https://github.com/ebiggers/libdeflate/blob/master/lib/arm/crc32_impl.h)
* [Fastest CRC32 for x86, Intel and AMD, + comprehensive derivation and discussion of various approaches](https://github.com/komrad36/CRC)
* [Writing better code with help from the compiler](https://www.kdab.com/wp-content/uploads/stories/slides/Day2/ThiagoMacieira_Better_code_help_from_the_compiler.pdf)

PR #627, opened 2024-01-16
Optimize CRC intrinisics for targets lacking of CRC extension

Cuda-Chen:
Close https://github.com/DLTcollab/sse2neon/issues/624.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -8500,12 +8500,47 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
     crc = __crc32cb(crc, v);
 #else
     crc ^= v;
-    for (int bit = 0; bit < 8; bit++) {
-        if (crc & 1)
-            crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
-        else
-            crc = (crc >> 1);
-    }
+#if defined(__ARM_FEATURE_CRYPTO)
+    // Adapted from: https://mary.rs/lab/crc32/
+    // Barrent reduction
+    uint64x2_t orig =
+        vcombine_u64(vcreate_u64((uint64_t) (crc) << 24), vcreate_u64(0x0));
+    uint64x2_t tmp = orig;
+
+    // Polynomial P(x) of CRC32C
+    uint64_t p = 0x105EC76F1;
+    // Barrett Reduction (in bit-reflected form) constant mu_{64} = \lfloor
+    // 2^{64} / P(x) \rfloor = 0x11f91caf6
+    uint64_t mu = 0x1dea713f1;
+
+    // Multiply by mu_{64}
+    tmp = _sse2neon_vmull_p64(vget_low_u64(tmp), vcreate_u64(mu));
+    // Divide by 2^{64} (mask away the unnecessary bits)
+    tmp =
+        vandq_u64(tmp, vcombine_u64(vcreate_u64(0xFFFFFFFF), vcreate_u64(0x0)));
+    // Multiply by P(x) (shifted left by 1 for alignment reasons)
+    tmp = _sse2neon_vmull_p64(vget_low_u64(tmp), vcreate_u64(p));
+    // Subtract original from result
+    tmp = veorq_u64(tmp, orig);
+
+    // Extract the 'lower' (in bit-reflected sense) 32 bits
+    crc = vgetq_lane_u32(vreinterpretq_u32_u64(tmp), 1);
+#else  // Fall back to the generic table lookup approach
+    // Adapted from: https://create.stephan-brumme.com/crc32/
+    // Apply half-byte comparision algorithm for the best ratio between
+    // performance and lookup table.
+
+    // The lookup table just needs to store every 16th entry
+    // of the standard look-up table.
+    static const uint32_t crc32_half_byte_tbl[] = {
+        0x00000000, 0x105ec76f, 0x20bd8ede, 0x30e349b1, 0x417b1dbc, 0x5125dad3,
+        0x61c69362, 0x7198540d, 0x82f63b78, 0x92a8fc17, 0xa24bb5a6, 0xb21572c9,
+        0xc38d26c4, 0xd3d3e1ab, 0xe330a81a, 0xf36e6f75,
+    };
+
+    crc = (crc >> 4) ^ crc32_half_byte_tbl[crc & 0x0F];
+    crc = (crc >> 4) ^ crc32_half_byte_tbl[crc & 0x0F];
+#endif
 #endif
     return crc;
 }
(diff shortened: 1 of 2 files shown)


sse2neon.h
```diff
@@ -8500,12 +8500,8 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
...
 #else
     crc ^= v;
-    for (int bit = 0; bit < 8; bit++) {
-        if (crc & 1)
-            crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
-        else
-            crc = (crc >> 1);
-    }
+    for (int bit = 0; bit < 8; bit++)
+        crc = (crc & 1) ? ((crc >> 1) ^ UINT32_C(0x82f63b78)) : (crc >> 1);
```
jserv, line 8504: Don't do this. The optimizing compilers can generate the identical machine code, and the original code is easier to maintain.

Cuda-Chen: Revert completed.

jserv:
Consider the following changes:
```diff
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -8501,10 +8501,9 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
 #else
     crc ^= v;
     for (int bit = 0; bit < 8; bit++) {
+        crc >>= 1;
         if (crc & 1)
-            crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
-        else
-            crc = (crc >> 1);
+            crc ^= UINT32_C(0x82f63b78);
     }
 #endif
     return crc;
```

sse2neon.h
```diff
@@ -8486,6 +8486,52 @@ FORCE_INLINE uint64_t _mm_crc32_u64(uint64_t crc, uint64_t v)
     return crc;
 }
 
+const uint32_t crc32_tbl[] = {
```
jserv, line 8489: Add `static` qualifier and enforce `sse2neon_` prefix.

jserv:
The current `_sse2neon_crc32_tbl` uses 256 * 4 bytes, totaling 1 KiB. Is there a way to reduce its memory usage?

Cuda-Chen:
> Consider the following changes:
> 
> ```diff
> --- a/sse2neon.h
> +++ b/sse2neon.h
> @@ -8501,10 +8501,9 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
>  #else
>      crc ^= v;
>      for (int bit = 0; bit < 8; bit++) {
> +        crc >>= 1;
>          if (crc & 1)
> -            crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
> -        else
> -            crc = (crc >> 1);
> +            crc ^= UINT32_C(0x82f63b78);
>      }
>  #endif
>      return crc;
> ```

This operation does binary division (we would like to check if the LSB is 1, then right-shift 1-bit then XOR, else we just right-shift 1-bit ).
As a result, first right-shifting 1-bit then checking whether the LSB is 1 produces the wrong result.

jserv:
Reference implementation in Linux kernel:
* [arch/arm/crypto/crc32-ce-glue.c](https://github.com/torvalds/linux/blob/master/arch/arm/crypto/crc32-ce-glue.c)
* [arch/arm/crypto/crc32-ce-core.S](https://github.com/torvalds/linux/blob/master/arch/arm/crypto/crc32-ce-core.S)

Accelerated CRC32(C) using ARM CRC, NEON and Crypto Extensions instructions.

sse2neon.h
```diff
@@ -8546,7 +8535,11 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
     crc = __crc32cb(crc, v);
 #else
     crc ^= v;
-    crc = (crc >> 8) ^ _sse2neon_crc32_tbl[crc & 0xFF];
+
+    // Apply half-byte comparision algorithm for the best ratio between
```
jserv, line 8511: Since you already shrink the size of the table, move the table inside the function body.

Cuda-Chen: Done.

sse2neon.h
```diff
@@ -8500,12 +8500,21 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
...
-    for (int bit = 0; bit < 8; bit++) {
-        if (crc & 1)
-            crc = (crc >> 1) ^ UINT32_C(0x82f63b78);
-        else
-            crc = (crc >> 1);
-    }
+
+    // Adapted from: https://create.stephan-brumme.com/crc32/
+    // Apply half-byte comparision algorithm for the best ratio between
+    // performance and look-up table spaces.
```
jserv, line 8506: Simply refer it as "lookup table."

jserv:
Reference paper titled [Fast CRC Computation for Generic Polynomials Using PCLMULQDQ Instruction](https://www.lambda-v.com/links/programming/files/intel_white_papers/fast-crc-computation-generic-polynomials-pclmulqdq-paper.pdf).

sse2neon.h
```diff
@@ -8500,12 +8500,47 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
...
+#if defined(__ARM_FEATURE_CRYPTO)
+    // Adapted from: https://mary.rs/lab/crc32/
+    // Barrent reduction
+    uint64x2_t orig =
+        vcombine_u64(vcreate_u64((uint64_t) (crc) << 24), vcreate_u64(0x0));
+    uint64x2_t tmp = orig;
+
+    // Polynomial P(x) of CRC32C
+    uint64_t p = 0x105EC76F1;
+    // Barrett Reduction constant (u64`) = u` = (x**64 / P(x))` = 0x11f91caf6 in
```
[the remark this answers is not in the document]

sse2neon.h
```diff
@@ -8500,12 +8500,47 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t crc, uint8_t v)
...
+    tmp =
+        vandq_u64(tmp, vcombine_u64(vcreate_u64(0xFFFFFFFF), vcreate_u64(0x0)));
+    // Multiply by P(x) (shifted left by 1 for alignment reasons)
+    tmp = _sse2neon_vmull_p64(vget_low_u64(tmp), vcreate_u64(p));
+    // Subtract original from result
+    tmp = veorq_u64(tmp, orig);
+
+    // Extract the 'lower' (in bit-reflected sense) 32 bits
+    crc = vgetq_lane_u32(vreinterpretq_u32_u64(tmp), 1);
+#else
```
jserv, line 8528: Add comment "Fall back to the generic table lookup approach"

jserv:
> before converting this PR to a normal PR, I would like to ask whether it is necessary to implement runtime dispatching in this PR?

Compile-time options for diverse Arm extensions would be great.

Cuda-Chen:
> > before converting this PR to a normal PR, I would like to ask whether it is necessary to implement runtime dispatching in this PR?
> 
> Compile-time options for diverse Arm extensions would be great.

For my experience, I will use `-mcpu` to determine the instruction set and the features (e.g., crypto and crc).

Anyway, I think I will make this PR ready for merge.

PR #630, opened 2024-04-18
Improve unsupported target message

### LINKED ISSUE #629: Get error "Unsupported target. Must be either ARMv7-A+NEON or ARMv8-A." on armv7 system
I am trying to compile a project using `sse2neon.h` inside a docker container from the base image `balenalib/raspberrypi3-debian:bookworm`, using QEMU for armv7 virtualization.

Here is some info as printed by the `meson` build system inside this container
```
C compiler for the host machine: cc (gcc 12.2.0 "cc (Raspbian 12.2.0-14+rpi1) 12.2.0")
C linker for the host machine: cc ld.bfd 2.40
Host machine cpu family: arm
Host machine cpu: armv7l
```

The build fails with "Unsupported target. Must be either ARMv7-A+NEON or ARMv8-A." in the file that the header is included but this system is clearly armv7.

I am not sure if this is something that must be fixed in this library or the compiler toolchain/distribution I am using, but I am making this issue report anyways because I believe this check can be removed or updated on the `sse2neon` side.

ankith26:
closes #629

Slightly improve the error message to suggest users a way to fix the issue.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -241,7 +241,9 @@ FORCE_INLINE void _sse2neon_smp_mb(void)
 #pragma GCC push_options
 #endif
 #else
-#error "Unsupported target. Must be either ARMv7-A+NEON or ARMv8-A."
+#error \
+    "Unsupported target. Must be either ARMv7-A+NEON or ARMv8-A \
+(you could try setting target explicitly with -march or -mcpu)"
 #endif
 #endif
 


jserv (changes requested):
Run `clang-format -i` to ensure the consistent style.

ankith26:
I have reformatted the line and force pushed the amended commit

jserv:
Thank @ankith26 for contributing!

PR #637, opened 2024-06-20
Fix undefined `_mm_malloc` and `_mm_free` when building with LLVM/MinGW.

bruvzg:
MinGW define  `_mm_malloc` and `_mm_free` only for x86, but not for ARM.

https://github.com/mingw-w64/mingw-w64/blob/7c9cfe6708cafc83c14a2654308b2db62b126eae/mingw-w64-headers/crt/malloc.h#L82-L85

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -120,10 +120,8 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-#if defined(_WIN32)
-/* Definitions for _mm_{malloc,free} are provided by <malloc.h>
- * from both MinGW-w64 and MSVC.
- */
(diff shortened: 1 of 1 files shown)


bruvzg:
Hm, seems like `20220906` do have it defined (I was using `20240619`), or it might be something specific about Godot/Embee build environment.

jserv:
> Hm, seems like `20220906` (used in CI) do have it defined (I was using `20240619`), or it might be something specific about Godot/Embee build environment.

@bruvzg, Do you plan to check the recent configurations and see if the above can be resolved?

jserv:
superseded by #657

PR #638, opened 2024-06-25
fix: Fix strict-aliasing errors

### LINKED ISSUE #635: Test failures with newer versions of clang/gcc using O1 and O2
After having a little more of a dig in relation to #633 - it appears that tests are failing on all platforms, including linux, when using newer versions of clang (tested against LLVM 14 through to LLVM 18), and using higher optimization levels.

GCC also has a subset of these failures.

This is reproducible on Linux (WSL - below) and Windows (#633).

A log (thanks to @pbo-linaro) is as follows:
```
$ clang++ --version
Debian clang version 14.0.6
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

$ clang++ tests/*.cpp -I . -O0 -o test
$ ./test  |& grep -i failed
Failed: 0

$ clang++ tests/*.cpp -I . -O1 -o test
$ ./test  |& grep -i failed
Test mm_set_rounding_mode           failed
Test mm_setcsr                      failed
Test mm_storeu_si64                 failed
Test mm_set_denormals_zero_mode     failed
Failed:  4

$ clang++ tests/*.cpp -I . -O2 -o test
$ ./test  |& grep -i failed
Test mm_set_rounding_mode           failed
Test mm_setcsr                      failed
Test mm_storeu_si64                 failed
Test mm_cvtpd_epi32                 failed
Test mm_cvtpd_pi32                  failed
Test mm_cvttpd_epi32                failed
Test mm_cvttpd_pi32                 failed
Test mm_set_denormals_zero_mode     failed
Failed:  8
```

Nothing suspect is returned by `fsanitize=address` or `fsanitize=undefined`.

At this point it's a little beyond my level of expertise with SIMD/NEON, so it would be appreciated if someone a little more versed could have a look.

howjmay:
Casting vector registers to a pointer or accessing the address of a
variable and then casting it to a different pointer type would
violate strict aliasing rules, potentially causing strict aliasing
errors.

`sse2neon_recast_u64_f64` and `sse2neon_recast_f64_s64` are introduced
to solve this error.

closes #635 #639

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -130,6 +130,20 @@
 
 #include <stdint.h>
 #include <stdlib.h>
+#include <string.h>
+
+FORCE_INLINE double sse2neon_recast_u64_f64(uint64_t u64)
+{
+    double f64;
+    memcpy(&f64, &u64, sizeof(uint64_t));
+    return f64;
+}
+FORCE_INLINE int64_t sse2neon_recast_f64_s64(double f64)
+{
+    int64_t i64;
+    memcpy(&i64, &f64, sizeof(uint64_t));
+    return i64;
+}
 
 #if defined(_WIN32)
 /* Definitions for _mm_{malloc,free} are provided by <malloc.h>
@@ -2981,11 +2995,17 @@ FORCE_INLINE __m128d _mm_add_pd(__m128d a, __m128d b)
     return vreinterpretq_m128d_f64(
         vaddq_f64(vreinterpretq_f64_m128d(a), vreinterpretq_f64_m128d(b)));
 #else
-    double *da = (double *) &a;
-    double *db = (double *) &b;
+    double a0 =
+        sse2neon_recast_u64_f64(vgetq_lane_u64(vreinterpretq_u64_m128d(a), 0));
+    double a1 =
+        sse2neon_recast_u64_f64(vgetq_lane_u64(vreinterpretq_u64_m128d(a), 1));
+    double b0 =
+        sse2neon_recast_u64_f64(vgetq_lane_u64(vreinterpretq_u64_m128d(b), 0));
+    double b1 =
+        sse2neon_recast_u64_f64(vgetq_lane_u64(vreinterpretq_u64_m128d(b), 1));
     double c[2];
-    c[0] = da[0] + db[0];
-    c[1] = da[1] + db[1];
+    c[0] = a0 + b0;
+    c[1] = a1 + b1;
     return vld1q_f32((float32_t *) c);
 #endif
 }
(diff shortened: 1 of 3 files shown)


howjmay:
`_mm_dp_pd` and `_mm_dp_ps` still have errors

aqrit:
Type punning via union is illegal in C++.
 C++20 added `std::bit_cast`.
Without `bit_cast` one could use `memcpy` for this, however that is still somewhat undefined?

Type punning via union is "okay" in C. Though it makes some assumptions about the floating point format and underlying integer types.

howjmay:
@aqrit 
Do you think I should change to the `memcpy` instead of using union?

aqrit:
AFAIK, `memcpy` is the only "safe" way to do this in C++ before C++20. 

```
#ifdef __cplusplus
# if __cplusplus >= 202002L
    // use bitcast
# else
    // use memcpy
# endif
#else // file is being compiled as "C"
    // use union
#endif
```

howjmay:
@aqrit how do you think about the new refactoring? I am using memcpy to recast variables now

howjmay:
@jserv If this implementation is ok, I will remove the optimization option and squash the commits

jserv:
> I still need some time to investigate the cause that makes `_mm_dp_pd` and `_mm_dp_ps` failed.

The above seems to be working. Is my understanding correct?

howjmay:
@jserv Yes. Here I have fixed `_mm_dp_pd` and `_mm_dp_ps`. The whole test cases were optimized in a wrong way, so we need to disable optimization for these two test cases

jserv:
> If this PR is ok, I will remove the optimization option and turn this PR into ready for review

Except for the `recast_f64` and `recast_i64` functions (the naming should be amended for consistency), I think the proposed changes are fine."

howjmay:
> > If this PR is ok, I will remove the optimization option and turn this PR into ready for review
> 
> Except for the `recast_f64` and `recast_i64` functions (the naming should be amended for consistency), I think the proposed changes are fine."

How about rename them as `sse2neon_recast_f64` and `sse2neon_recast_i64`?

jserv:
> How about renaming them as `sse2neon_recast_u64_f64` and `sse2neon_recast_f64_s64`?

It sounds pretty well.

jserv (changes requested):
Since this is not a trivial change, the git commit messages should be amended to be more informative. You might consider to split the proposed changes into several commits.

howjmay:
@jserv I have added the commit message. 
However, when it comes to splitting this PR into several small commits, I think I can only separate it into a commit that introducing `OPTNONE` for some tests/intrinsics and another one handling the strict-aliasing errors. 
How do you think about it?

jserv:
Is it time to close #635 and #639 ? If so, refine the git commit messages accordingly.

howjmay:
I think it is ok now

PR #640, opened 2024-06-26
fix: Disable optimization to avoid pontential errors

howjmay:
Some function may be ignored or not well executed if optimization is applied. A classic example is `test test_mm_set_rounding_mode`.

_mm_round_ps is computed in compiling time, in which case, rounding mode changing is ignored.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -106,6 +106,17 @@
 #pragma message("Macro name collisions may happen with unsupported compilers.")
 #endif
 
+
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma push_macro("FORCE_INLINE_OPTNONE")
+#define FORCE_INLINE_OPTNONE static inline __attribute__((optimize("O0")))
+#elif defined(__clang__)
+#pragma push_macro("FORCE_INLINE_OPTNONE")
+#define FORCE_INLINE_OPTNONE static inline __attribute__((optnone))
+#else
+#define FORCE_INLINE_OPTNONE FORCE_INLINE
+#endif
+
 #if !defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10
 #warning "GCC versions earlier than 10 are not supported."
 #endif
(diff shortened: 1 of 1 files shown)


howjmay:
relates to #635

sse2neon.h
```diff
@@ -106,6 +106,14 @@
 #pragma message("Macro name collisions may happen with unsupported compilers.")
 #endif
 
+#if defined(__GNUC__) && !defined(__clang__)
+#define FORCE_INLINE_NO_OPTIMIZE static inline __attribute__((optimize("O0")))
```
jserv, line 110: Let's decouple `__attribute__((optimize("O0")))` or `__attribute__((optnone))` out of `FORCE_INLINE`. That is, there should be a new macro `SSE2NEON_OPTNONE` to suppress the optimizations for certain functions.

howjmay: `always_inline` conflicts with `SSE2NEON_OPTNONE`, so the suppressed funcs will be like
```c
static inline SSE2NEON_OPTNONE __m128i _mm_cvtpd_epi32(__m128d a)
```

Not sure is it better to merge all the attributes together?

howjmay:
Another question. 
Should we add another option for developers to disable forced not optimized attributes here?

sse2neon.h
jserv: > `always_inline` conflicts with `SSE2NEON_OPTNONE`, so the suppressed funcs will be like

Got it. We then need new macro `FORCE_INLINE_OPTNONE` , which should be specified via `##pragma push_macro`. `OPTNONE` is shorter than `NO_OPTIMIZE`.

jserv:
> Another question. Should we add another option for developers to disable forced not optimized attributes here?

I found no reason to disable forced not optimized attributes. Can you enumerate the considerations?

howjmay:
The error we met this time happened when the rounding mode is changed, and the input is given in compiled time.
Not sure if there will be the use case that the input is given and the rounding mode is unchanged, then the original implementation will be faster

PR #642, opened 2024-07-18
Allow optimization and use fesetround(), fegetround()

### LINKED ISSUE #648: Optimizations disabled in `_MM_SET_ROUNDING_MODE`
`_MM_SET_ROUNDING_MODE` is declared as `FORCE_INLINE_OPTNONE`, which lowers performance. Unless I'm missing something, wouldn't it be better to define the `r` union as `volatile` instead?

howjmay:
### feat: Allow optimization option in sse2neon.h    
 Revert the previous restriction that some of the functions are forced
to not be optimized when compiling-time optimization options were
given.    
Now it is users' responsibility to ensure the behavior after
optimization.
Shifting the responsibility to the users enables sse2neon the run in
the optimized state in general, but not restricted by some specific
scenarios.

### feat: Use fesetround() and fegetround()    

Setting/getting rounding mode directly through fpcr with volatile
keyword could be unstable.
Therefore we use the C99 fesetround()/fegetround() here to ensure
the behavior.

closes #648

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,10 @@ $ make \
 
 Check the details via [Test Suite for SSE2NEON](tests/README.md).
 
+### Optimization
+
+Misbehavior may exist when turning on optimization options. Developers should be aware of possible errors.
+
 ## Adoptions
 Here is a partial list of open source projects that have adopted `sse2neon` for Arm/Aarch64 support.
 * [Aaru Data Preservation Suite](https://www.aaru.app/) is a fully-featured software package to preserve all storage media from the very old to the cutting edge, as well as to give detailed information about any supported image file (whether from Aaru or not) and to extract the files from those images.
--- a/README.md
+++ b/README.md
@@ -157,7 +157,12 @@ Check the details via [Test Suite for SSE2NEON](tests/README.md).
 
 ### Optimization
 
-Misbehavior may exist when turning on optimization options. Developers should be aware of possible errors.
+The SSE2NEON project is designed with performance-sensitive scenarios in mind, and as such, optimization options (e.g. `O1`, `O2`) can lead to misbehavior under specific circumstances. For example, frequent changes to the rounding mode or repeated calls to `_MM_SET_DENORMALS_ZERO_MODE()` may introduce unintended behavior.
+
+Enforcing no optimizations for specific intrinsics could solve these boundary cases but may negatively impact general performance. Therefore, we have decided to prioritize performance and shift the responsibility for handling such edge cases to developers.
+
+It is important to be aware of these potential pitfalls when enabling optimizations and ensure that your code accounts for these scenarios if necessary.
+
 
 ## Adoptions
 Here is a partial list of open source projects that have adopted `sse2neon` for Arm/Aarch64 support.
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -106,20 +106,14 @@
 #pragma message("Macro name collisions may happen with unsupported compilers.")
 #endif
 
-
-#if defined(__GNUC__) && !defined(__clang__)
-#pragma push_macro("FORCE_INLINE_OPTNONE")
-#define FORCE_INLINE_OPTNONE static inline __attribute__((optimize("O0")))
-#elif defined(__clang__)
-#pragma push_macro("FORCE_INLINE_OPTNONE")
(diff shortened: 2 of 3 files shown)


sse2neon.h
```diff
@@ -1846,25 +1840,17 @@ FORCE_INLINE unsigned int _sse2neon_mm_get_flush_zero_mode(void)
...
+    case FE_TONEAREST:
+        return _MM_ROUND_NEAREST;
+    case FE_DOWNWARD:
+        return _MM_ROUND_DOWN;
+    case FE_UPWARD:
+        return _MM_ROUND_UP;
+    case FE_TOWARDZERO:
+        return _MM_ROUND_TOWARD_ZERO;
+    default:  // FIXME
+        return _MM_ROUND_TOWARD_ZERO;
```
howjmay, line 1853: @jserv I am not sure whether we have a better value to return in the error case, so I temporarily return `_MM_ROUND_TOWARD_ZERO ` here.

sse2neon.h
```diff
@@ -2458,46 +2444,25 @@ FORCE_INLINE __m128 _mm_set_ps1(float _w)
...
+        rounding = FE_UPWARD;
         break;
-    default:  //_MM_ROUND_NEAREST
-        r.field.bit22 = 0;
-        r.field.bit23 = 0;
+    case _MM_ROUND_TOWARD_ZERO:
+        rounding = FE_TOWARDZERO;
+        break;
+    default:  // FIXME
+        rounding = FE_TOWARDZERO;
```
howjmay, line 2463: same here

sse2neon.h
jserv: > I am not sure whether we have a better value to return in the error case, so I temporarily return `_MM_ROUND_TOWARD_ZERO ` here.

It is reasonable, and you should explain more in comments.

howjmay: I have added comments for explaining the error case

jserv:
I would like to ask @jmather-sesi and @alexorlov124 for reviewing.

jserv (changes requested):
Remove the label `refactor` in the git commit messages. The changes were not associated with refactoring. Instead, the commits did change the behavior with compiler optimizations.

howjmay:
I changed to `feat` labels now

README.md
```diff
@@ -155,6 +155,10 @@ $ make \
 
 Check the details via [Test Suite for SSE2NEON](tests/README.md).
 
+### Optimization
+
+Misbehavior may exist when turning on optimization options. Developers should be aware of possible errors.
```
jserv, line 160: This message was a bit misleading. You should summarize the issues reported by the developers as known considerations.

howjmay: More content has been added to explain the context

sse2neon.h
```diff
@@ -106,28 +106,22 @@
...
-#else
-#define FORCE_INLINE_OPTNONE FORCE_INLINE
-#endif
-
 #if !defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10
 #warning "GCC versions earlier than 10 are not supported."
 #endif
 
+#ifdef __OPTIMIZE__
+#warning "Optimization may cause potential errors in sse2neon. see #648"
```
jserv, line 114: Instead of listing #648, would it make sense to refer to the documentation instead?

howjmay: amended

sse2neon.h
```diff
@@ -106,28 +106,23 @@
...
-#define FORCE_INLINE_OPTNONE FORCE_INLINE
-#endif
-
 #if !defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10
 #warning "GCC versions earlier than 10 are not supported."
 #endif
 
+#ifdef __OPTIMIZE__
+#warning \
+    "Optimization may cause potential errors in sse2neon. See `Optimization` section in root README."
```
jserv, line 115: Rephrase as following:
> Report any potential compiler optimization issues when using SSE2NEON.
> See the 'Optimization' section at https://github.com/DLTcollab/sse2neon.

No [backtick](https://en.wikipedia.org/wiki/Backtick) here.

jserv:
Thank @howjmay for contributing!

howjmay:
thanks for reviewing

PR #643, opened 2024-07-18
fix test_mm_dp_pd test

alexorlov124:
When I ran make check with -O2 passed to ARCH_CFLAGS, the test_mm_dp_pd failed. After debugging, I discovered that a buffer initialized with floats was being cast to a buffer of doubles. The root cause of the test failure was that the exact match of the dot product obtained with intrinsics and calculated with C++ doubles showed differences. This occurred because the numbers were too large, and the dot product had rounding errors.

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -5877,7 +5877,7 @@ result_t test_mm_shuffle_epi32(const SSE2NEONTestImpl &impl, uint32_t iter)
     int32_t _d[4];
 
 #define TEST_IMPL(IDX)              \
-    _d[0] = _a[((IDX) &0x3)];       \
+    _d[0] = _a[((IDX) & 0x3)];      \
     _d[1] = _a[((IDX >> 2) & 0x3)]; \
     _d[2] = _a[((IDX >> 4) & 0x3)]; \
     _d[3] = _a[((IDX >> 6) & 0x3)]; \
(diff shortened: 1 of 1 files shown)


howjmay:
I think you use the different version of clang-format. It should be `clang-format-12`. But you can refer CI directly to format the code

alexorlov124:
> I think you use the different version of clang-format. It should be `clang-format-12`. But you can refer CI directly to format the code

Thank you, I installed clang-formatter 12 and now format is fixed

howjmay:
Btw may I ask which compiler and platform you used to trigger the error?

alexorlov124:
> Btw may I ask which compiler and platform you used to trigger the error?

I used gcc 10.2 and 11.4, Graviron3 and Amazon Linux 2023

howjmay:
Sorry it seems not to pass the CI. May you help me to enable `-O2` in the CI and check it ?

https://github.com/howjmay/sse2neon/actions/runs/10019289097

alexorlov124:
> Sorry it seems not to pass the CI. May you help me to enable `-O2` in the CI and check it ?
> 
> https://github.com/howjmay/sse2neon/actions/runs/10019289097

I updated .github\workflows\main.yml with -O2 flag not sure we should merge it in master but we can make sure tests are green

alexorlov124:
Here my pull request with optimization enabled https://github.com/DLTcollab/sse2neon/pull/644

jserv (changes requested):
Squash the git commits and refine the commit messages.
See https://cbea.ms/git-commit/

howjmay:
> Here my pull request with optimization enabled #644

Thank you for the test. However, it seems the test didn't pass. Not sure if I did something wrong. 

I think this PR should be ok to merge after squash. But I just feel strange. It seems this PR can't solve the error when O2 is given

jserv:
Thank @alexorlov124 for contributing!

howjmay:
I will investigate what cause O2 fail in #642

alexorlov124:
Welcome,
I could participate in investigating other test fails, Right now I need to
finish a project in my work.

Thank you!

чт, 25 июл. 2024 г. в 16:34, Jim Huang ***@***.***>:

PR #646, opened 2024-08-16
Fix performance regression after OPTNONE changes

sergeyvfx:
The OPTNONE changes in #638 introduced 25% performance regression in Cycles render engine running on Apple Silicon:

  https://projects.blender.org/blender/blender/issues/126408

The reason for this is because as per Clang documentation optnone is incompatible with inline, so none of the functions that is marked as optnone is inlined. The biggest bottleneck for Cycles after that change is _mm_mul_ps.

This change makes it so _mm_mul_ps is inlined and is no longer marked as optnone. This solves the immediate performance regression, and the correctness is verified using the sse2neon's test suit on M2 Ultra and M3 Max, with various optimization levels (default, -O2, -O3).

Additionally, adding -Wstrict-aliasing flag does not introduce new warnings.


jserv:
Thank @sergeyvfx for contributing!

PR #650, opened 2024-10-08
CI: Bump dependency

jserv:


howjmay (approved):
Maybe we can bump the version of clang-format too?

jserv:
> Maybe we can bump the version of clang-format too?

The `clang-format` should be upgraded along with Ubuntu Linux systems. I plan to switch to Ubuntu 24.04-LTS.

PR #652, opened 2024-11-04
Add _MM_SHUFFLE2() macro for shuffle parameter for _mm_shuffle_pd()

rouault:
This macro is defined in emmintrin.h and used sometimes by application. Cf https://clang.llvm.org/doxygen/emmintrin_8h.html#a3aeb3a59d7de7f9ca49c1153baa9da58 for its definition


jserv:
Thank @rouault for contributing!

PR #653, opened 2024-11-08
fewer intrinsics in _mm_movemask_ps

shinjiogaki:


aqrit:
an alternative is to extract bits with using `umulh`:
```
uint64_t x = vget_lane_s64(vreinterpret_s64_s16(vqmovn_s32(vreinterpretq_s32_m128i(a))),0);
const uint64_t mask = 0x8000800080008000ULL;
const uint64_t magic = 0x0002000400080010ULL;
return (uint8_t)(((__uint128_t)(x & mask) * (__uint128_t)magic) >> 64);
```

I don't know which is faster.

howjmay:
@shinjiogaki are you interested in giving a try to see whether this bring performance enhancement?

aqrit:
I haven't tested it but... I'm pretty sure the following would be the best `__aarch64__` implementation:
```
uint64_t x = vget_lane_s64(vreinterpret_s64_s16(vshrn_n_s32(vreinterpretq_s32_m128i(a), 15)), 0);
return (x * 0x0F001F003F008000) >> 60;
```

magic number was adapted from https://zeux.io/2022/09/02/vpexpandb-neon-z3/

PR #654, opened 2024-11-10
README.md: mention GDAL

rouault:
We are happy users of sse2neon with https://github.com/OSGeo/gdal/pull/11202 and https://github.com/OSGeo/gdal/pull/11239


jserv:
Thank @rouault for sharing this successful story!

PR #657, opened 2024-12-25
Fix undefined _mm_{malloc,free} with LLVM/MinGW

jserv:


invertego:
If an alignment of 1 is passed to _mm_malloc(), this implementation will allocate with malloc() and then _mm_free() will free with _aligned_free(). The latter function is only designed to free memory allocated with an _aligned function.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-free

I think if you're going to use _aligned_malloc() at all, it needs to be done unconditionally, not just for certain alignments.

PR #659, opened 2025-01-06
Fix mismatched _mm_{malloc,free} with LLVM/MinGW

### LINKED ISSUE #658: Windows implementation of _mm_alloc()/_mm_free() can use mismatched allocators
Reproducing https://github.com/DLTcollab/sse2neon/pull/657#issuecomment-2564587426 here:

If an alignment of 1 is passed to _mm_malloc(), this implementation will allocate with malloc() and then _mm_free() will free with _aligned_free(). The latter function is only designed to free memory allocated with an _aligned function.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-free

I think if you're going to use _aligned_malloc() at all, it needs to be done unconditionally, not just for certain alignments.

invertego:
Regular malloc() should not be used for 1-byte aligned allocations, as such allocations cannot be freed correctly by _aligned_free().

Fixes #658

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -1977,20 +1977,18 @@ FORCE_INLINE __m128i _mm_loadu_si64(const void *p)
 #if !defined(SSE2NEON_ALLOC_DEFINED)
 FORCE_INLINE void *_mm_malloc(size_t size, size_t align)
 {
+#if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


jserv:

PR #661, opened 2025-02-02
Enable `-Wconversion`

marcin-serwin:
When trying to compile [torzu](https://notabug.org/litucks/torzu) I got a bunch of errors originating from this library regarding the implicit int conversion. I patched it around by disabling `-Wconversion` but I figured it would be nice to address them here. The first commit enables this flag in the Makefile and fixes the resulting warnings. I've also made some minor improvements to the code I was touching.

I've also encountered `-Wshadow` errors because torzu is using `s64`, `f64`, and `u64` as type aliases and this library is using them as variable names. While these variable names are perfectly reasonable, I hope it's okay that I changed them in the second commit to avoid this. I've seen similar type aliases used elsewhere so keeping these variable names will probably trip up someone else sooner or later.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -128,17 +128,17 @@
 #include <stdlib.h>
 #include <string.h>
 
-FORCE_INLINE double sse2neon_recast_u64_f64(uint64_t u64)
+FORCE_INLINE double sse2neon_recast_u64_f64(uint64_t val)
 {
-    double f64;
-    memcpy(&f64, &u64, sizeof(uint64_t));
-    return f64;
+    double tmp;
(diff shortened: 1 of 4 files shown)


sse2neon.h
```diff
@@ -128,17 +128,17 @@
 #include <stdlib.h>
 #include <string.h>
 
-FORCE_INLINE double sse2neon_recast_u64_f64(uint64_t u64)
```
howjmay, line 131: not sure about the reason change this

jserv:
Thank @marcin-serwin for contributing!

PR #662, opened 2025-02-04
fix missing parenthesis

coco875:


jserv:
Thank @coco875 for contributing!

PR #664, opened 2025-02-24
Fix GitHub Action status badge

Mes0903:
There was an typo in the URL, which cause the svg file can't display correctly:

image


jserv:
Thank @Mes0903 for contributing!

PR #665, opened 2025-03-25
Support Windows Arm64EC

mcfi:
This change adds support for Windows Arm64EC build through proper ifdefs.

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -148,6 +148,9 @@ FORCE_INLINE int64_t sse2neon_recast_f64_s64(double val)
 
 /* If using MSVC */
 #ifdef _MSC_VER
+#if defined(_M_ARM64EC)
+#define _DISABLE_SOFTINTRIN_ 1
+#endif
 #include <intrin.h>
 #if SSE2NEON_INCLUDE_WINDOWS_H
 #include <processthreadsapi.h>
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Indent the source files via clang-format-18.

mcfi:
Thanks. To clarify, do you mean the clang-format command in clang 18? What's the exact clang version does the official build use to check the format?

jserv:
> Thanks. To clarify, do you mean the clang-format command in clang 18? What's the exact clang version does the official build use to check the format?

The automated continuous integration process incorporated clang-format version `18.1.3` into the Ubuntu Linux 24.04 development infrastructure.

mcfi:
Done, this.

jserv:
The CI reported [build failure on Arm Cortex-A32 hosts](https://github.com/DLTcollab/sse2neon/actions/runs/14073176076/job/39411708541), with flags `-mcpu=cortex-a32 -mfpu=neon-fp-armv8`. The following changes should be refined:
```c
/* On ARMv7, some registers, such as PMUSERENR and PMCCNTR, are read-only
 * or even not accessible in user mode.
 * To write or access to these registers in user mode,
 * we have to perform syscall instead.
 */
#if (!defined(__aarch64__) && !defined(_M_ARM64) && !defined(_M_ARM64EC)) && \
    (__ARM_ARCH == 7)
#include <sys/time.h>
#endif
```

[Cortex-A32](https://developer.arm.com/Processors/Cortex-A32) processor provides the 32-bit Armv8-A features, meaning that its `__ARM_ARCH` is not 7.

mcfi:
Thanks, @jserv . I don't recall why I added that __ARM_ARCH == 7 check. Perhaps AI did since I was using VSCode with Copilot..

jserv (changes requested):
Squash the git commits and draft commit messages that explain the benefit or use scenario of Arm64EC awareness in the project.

jserv:
Thank @mcfi for contributing!

PR #667, opened 2025-04-07
Update README.md with Windows Arm64EC instructions

mcfi:
Windows headers for Arm64EC includes SSE headers by default. To exclude those headers, developers need to define a macro `_DISABLE_SOFTINTRIN_` either through MSVC command line option or `#include` before any Windows headers.


jserv:
Thank @mcfi for contributing!

PR #671, opened 2025-06-17
Migrate the code to use C++ style cast operators

marcin-serwin:
This patch enables old-style-cast warnings and fixes reported problems. If the header is used as a C library the old casts are used.


jserv:
Thank you for this non-trivial change. However, I am hesitant to merge it because it modifies how C-style casting works and introduces new rules that must be followed. My concern is that this could create potential issues and make it more difficult for contributors to comply with the updated requirements.

eeshvardasikcm:
This is one move towards language agnostic

jserv:
@cubic-dev-ai review this PR

jserv:
@marcin-serwin, Can you refine based on the above review messages?

PR #678, opened 2025-09-18
Fix problem with _MM_GET_ROUNDING_MODE on certain platforms

mg-mburetorp:
fegetround() returns a bitmask and depending on the platform there may be bits used to indicate more than just the current rounding mode. In those cases _MM_GET_ROUNDING_MODE will always take the default switch case and incorrectly truncate even if _MM_ROUND_NEAREST was set.


jserv:
@cubic-dev-ai review this PR

PR #679, opened 2025-09-18
Fix bug with _mm_round and _MM_FROUND_NO_EXC

mg-mburetorp:
Fix _mm_round_pd / _mm_round_ps ignoring rounding argument when _MM_FROUND_NO_EXC was not set. If _MM_FROUND_NO_EXC was set the rounding argument was respected, however if was not set it would always use _MM_FROUND_CUR_DIRECTION rounding and incorrectly ignore the argument.

I don't know if _MM_FROUND_NO_EXC is supposed to or even can be emulated, but regardless, changing the rounding mode depending on this flag is a bug.


jserv (changes requested):
Enforce the rules of [How to Write a Git Commit Message](https://cbea.ms/git-commit/).

jserv:
> I don't know if _MM_FROUND_NO_EXC is supposed to or even can be emulated, but regardless, changing the rounding mode depending on this flag is a bug.

You can update the test suite as well.

PR #708, opened 2025-12-10
Optimize _mm_crc32_u16 on Armv8-A 32-bit platform

Cuda-Chen:
Optimize `_mm_crc32_u16()` on Armv8-A 32-bit platform using Barrett reduction shortcut.

---
## Summary by cubic
Optimized _mm_crc32_u16 on Armv8-A 32-bit by using a NEON Barrett reduction path when the crypto extension is available, reducing work vs the byte-wise fallback. Also unified the CRC32 path with a shared macro for consistency with _mm_crc32_u8.

- **Refactors**
  - Added SSE2NEON_CRC32C_BASE macro (supports 8- and 16-bit updates via NEON pmull).
  - Applied the macro to _mm_crc32_u16 and _mm_crc32_u8 under __ARM_FEATURE_CRYPTO.
  - Simplified the generic fallback: single xor and half-byte lookup table retained.

<sup>Written for commit d37a86040616e604c239894946785be5cb2450e3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -9311,6 +9311,50 @@ FORCE_INLINE __m128i _mm_cmpgt_epi64(__m128i a, __m128i b)
 #endif
 }
 
+/* A function-like macro to generate CRC-32C calculation using Barrett
+ * reduction.
+ *
+ * The input parameters depict as follows:
+ * - 'crc' means initial value or CRC.
+ * - 'v' means the element of input message.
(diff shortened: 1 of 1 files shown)


jserv:
> I think there are some rooms for optimization but it will take more time (e.g., alignment on different input size). I would like to request to merge this after I confirm the rooms for optimization.

Create GitHub issue(s) for further optimizations.

PR #759, opened 2026-02-16
mm_cmpistrm: fix one test that wasn't being run

mr-c:
The test data is indeed unit and not bit based.

The previous bad test wasn't being run as "SWORD_RANGES_UNIT" is specified on line 12481, and so the test using "SWORD_RANGES_BIT" wasn't triggered.

---
## Summary by cubic
Fixes a misconfigured mm_cmpistrm SWORD range test so it runs in the correct mode. Replaces IMM_SWORD_RANGES_BIT with IMM_SWORD_RANGES_UNIT to match the unit-based data.

- **Bug Fixes**
  - Switched the test immediate to IMM_SWORD_RANGES_UNIT in tests/impl.cpp.
  - The suite runs SWORD_RANGES_UNIT, so the BIT-based test never executed; this now validates the intended case.

<sup>Written for commit 6cbcc1aea7331525278e9c20bde32eb71a879717. Summary will update on new commits.</sup>

### DIFF
--- a/tests/impl.cpp
+++ b/tests/impl.cpp
@@ -12436,7 +12436,7 @@ static test_mm_cmpistrm_sword_data_t
         {
             {-39, -10, 17, 89, 998, 1000, 1234, 4566},
             {-40, -52, -39, -29, 100, 1024, 4565, 4600},
-            IMM_SWORD_RANGES_BIT,
+            IMM_SWORD_RANGES_UNIT,
             {0, 0, -1, -1, 0, 0, -1, 0},
         },
         {


mr-c:
Thank you all for the excellent hand written test cases! We are copying some of them into SIMDe 🙏🏻

jserv:
> Thank you all for the excellent hand written test cases! We are copying some of them into SIMDe 🙏🏻

I am pleased to hear that SSE2NEON has been helpful to the development of SIMDe.

With the release of SSE2NEON v1.9.x, I believe that all known translations from SSE intrinsics to their Armv7-A and Armv8-A counterparts have been fully implemented and validated. The current codebase reflects a comprehensive and carefully tested mapping across the supported intrinsic set.

It would be excellent to see selected improvements from SSE2NEON cherry-picked into SIMDe, further strengthening it as a unified and robust SIMD translation solution across architectures.

Issue #537, opened 2022-10-08
Scalable Vector Extension (SVE) support


jserv:
Quote from [Arm website](https://developer.arm.com/Architectures/Scalable%20Vector%20Extensions):
> Scalable Vector Extension (SVE) is a vector extension the A64 instruction set of the Armv8-A architecture. Armv9-A builds on SVE with the SVE2 extension. Unlike other SIMD architectures, SVE and SVE2 do not define the size of the vector registers, but constrains it to a range of possible values, from a minimum of 128 bits up to a maximum of 2048 in 128-bit wide units. Therefore, any CPU vendor can implement the extension by choosing the vector register size that better suits the workloads the CPU is targeting. The design of SVE and SVE2 guarantees that the same program can run on different implementations of the instruction set architecture without the need to recompile the code.

[genarchbench](https://github.com/LorienLV/genarchbench) comes with partial [sse2sve](https://github.com/LorienLV/genarchbench/blob/master/benchmarks/bsw/src/sse2sve.h) header, which can be regarded as a starting point for SSE2NEON to support SVE.

jserv:
## High-level view: where SVE/SVE2 can help SSE2NEON

From an SSE2NEON perspective, SVE/SVE2 is most attractive for intrinsics that are currently expensive to emulate with NEON, especially when operations involve:

* Horizontal reductions (sum / min / max)
* Dot products and multiply–add patterns
* Sum of absolute differences (SAD) and absolute-diff–plus–accumulate
* Saturating arithmetic and widening/narrowing operations
* Certain shuffle / table-lookup patterns
* Higher-level string scanning or token search behavior

In contrast, intrinsics that already map 1:1 to simple NEON operations (add, sub, mul, bitwise ops) usually do not gain much from SVE unless the surrounding algorithm is redesigned to actually exploit scalable vector length.

## 1. Horizontal reductions and dot-product-like patterns

SSE often implements horizontal reductions and dot products with multiple instructions:
* Multiply, then add in stages
* Use `shuffle` and `add` to accumulate within a 128-bit lane

In NEON, SSE2NEON does essentially the same thing using:
* Pairwise add (`vpadd*`)
* Manual `vget_*`/`vcombine_*` plus scalar adds

SVE/SVE2 provides direct reduction operations, for example:
* Sum of all elements: `svaddv_*`
* Min/max of all elements: `svminv_*`, `svmaxv_*`

[...]

PR #773, opened 2026-08-20
Add opt-in SVE2 fast path for _SIDD_CMP_EQUAL_ANY

alanhc:
Partially addresses #537.

SVE2 `MATCH` (FEAT_SVE2) compares each element of its first operand against
every element in the corresponding 128-bit segment of its second, setting the
destination predicate accordingly. That is exactly the `PCMPISTRM`/`PCMPESTRM`
EQUAL_ANY relation, which the current NEON path computes by materialising a
16x16 comparison matrix and reducing it row by row.

## Change

`sse2neon.h`, 122 insertions, no deletions:

- `SSE2NEON_ENABLE_SVE`, default 0, documented alongside the existing
  `SSE2NEON_PRECISE_*` tunables.
- `SSE2NEON_HAS_SVE2`, which additionally requires `SSE2NEON_ARCH_AARCH64` and
  `__ARM_FEATURE_SVE2`. When any condition fails the NEON path is used, so the
  tunable is safe to define unconditionally.
- SVE2 implementations of `_sse2neon_cmp_byte_equal_any` and
  `_sse2neon_cmp_word_equal_any`. Their signatures and their slots in
  `_sse2neon_cmpfunc_table` are unchanged, so `SSE2NEON_COMP_AGG` and all 14
  public `_mm_cmpistr*`/`_mm_cmpestr*` entry points are untouched.

`Makefile`, 48 insertions, 12 deletions: an `SVE=1` flag following the shape of
the existing `SANITIZE` and `STRICT_ALIASING` blocks, and a `check-sve` target.

`MATCH` is defined for 8- and 16-bit elements only, which covers exactly the
byte and word forms required here.

## Length handling

`MATCH` always scans the full 128-bit segment of its second operand, so lanes
of `a` at and beyond `la` must not contribute. They are filled with `a[0]`:

    za_valid[i] = a[i]  for i < la
                = a[0]  for la <= i < 16

so `MATCH(b[j])` is `(exists i < la: b[j] == a[i]) or (b[j] == a[0])`. Since
`la >= 1` at that point, `a[0]` lies within the valid range and the second
disjunct is subsumed by the first. `la == 0` returns 0 before the comparison,
as the NEON path does.

`SSE2NEON_CMPESTRX_LEN_PAIR` takes the absolute value and clamps to `bound`,
so `la` and `lb` reach these functions in `[0, 16]` for bytes and `[0, 8]` for
words. No further clamping is required.

## Vector length

`__m128` is a fixed 128-bit contract, so the code uses fixed-width predicates
(`svptrue_pat_b8(SV_VL16)` and friends) rather than vector-length-agnostic
loops. It is nonetheless written with sizeless SVE types only and remains
correct at any hardware vector length.

`-msve-vector-bits` is deliberately not used. It is a global ABI option that a
header cannot impose on its users, and it renders the code undefined on any
machine whose vector length differs from the value it names.

The NEON-to-SVE move goes through memory because `svset_neonq`
(`arm_neon_sve_bridge.h`) requires GCC 14 or Clang 16, above this project's
stated minimum. Using the bridge where available is a possible follow-up.

## Instruction counts

aarch64-linux-gnu-gcc 13.3, `-O2`:

| | NEON | SVE2 |
|---|---|---|
| `_sse2neon_cmp_byte_equal_any` | 150 | 30 |
| `_sse2neon_cmp_word_equal_any` | 78 | 24 |

## Measurements

Two SoC generations, six microarchitectures, both with SVE2 at 128-bit vectors.
The NEON and SVE2 paths are linked into one binary as two translation units and
measured back to back, so they see identical thermal and frequency conditions.
Per call, 16-byte vectors, minimum of 15 trials of 200 repetitions, pinned with
`taskset`. NEON -> SVE2:

Google Tensor G3, Android 15:

| | Cortex-X3 2.9 GHz | Cortex-A715 2.4 GHz | Cortex-A510 1.7 GHz |
|---|---|---|---|
| `_mm_cmpistri` byte | 11.43 -> 3.73 ns (3.07x) | 22.98 -> 6.75 ns (3.40x) | 71.93 -> 49.30 ns (1.46x) |
| `_mm_cmpistrm` byte | 11.67 -> 3.94 ns (2.96x) | 23.39 -> 7.41 ns (3.16x) | 76.65 -> 55.23 ns (1.39x) |
| `_mm_cmpistri` word | 5.57 -> 3.11 ns (1.79x) | 11.45 -> 5.87 ns (1.95x) | 48.74 -> 45.72 ns (1.07x) |

Google Tensor G5, Android 16:

| | Cortex-X4 3.8 GHz | Cortex-A725 3.1 GHz | Cortex-A520 2.2 GHz |
|---|---|---|---|
| `_mm_cmpistri` byte | 9.70 -> 3.09 ns (3.14x) | 17.87 -> 4.64 ns (3.85x) | 57.32 -> 37.44 ns (1.53x) |
| `_mm_cmpistrm` byte | 10.07 -> 3.28 ns (3.07x) | 18.12 -> 4.86 ns (3.73x) | 60.93 -> 41.47 ns (1.47x) |
| `_mm_cmpistri` word | 5.00 -> 2.57 ns (1.94x) | 8.82 -> 4.10 ns (2.15x) | 40.93 -> 36.97 ns (1.11x) |

The gain is below the instruction-count ratio because the memory round trip and
the predicate-to-bitmask conversion are latency rather than throughput.

The in-order little cores gain least on both generations and by nearly the same
margin, 1.46x on Cortex-A510 and 1.53x on Cortex-A520, which suggests the limit
is in-order execution itself rather than any particular vector-unit
arrangement. The best ratio comes from the mid cores rather than the largest
one, presumably because the NEON path's 150 instructions cost relatively more
where execution resources are narrower.

## Why not movemask

movemask is the more obvious SVE target and was evaluated first. Base SVE and
SVE2 provide no predicate-to-general-register bitmask instruction; `PMOV` is
FEAT_SVE2p1 (Armv9.4-A). The horizontal reduction and the vector-to-general-
register transfer therefore survive translation, and the result is slower than
the existing NEON code.

Steady-state loop body, work instructions, loop-invariant setup hoisted:

| | `_mm_movemask_epi8` | `_mm_movemask_ps` | `_mm_movemask_pd` |
|---|---|---|---|
| NEON | 8 | 4 | 4 |
| SVE2 | 9 | 5 | 5 |

Measured on the same devices, per call, NEON -> SVE2. SVE2 is slower on every
core tested:

| | Cortex-X3 | Cortex-A715 | Cortex-A510 |
|---|---|---|---|
| `_mm_movemask_epi8` | 0.93 -> 1.80 ns | 1.97 -> 2.68 ns | 17.19 -> 19.00 ns |

| | Cortex-X4 | Cortex-A725 | Cortex-A520 |
|---|---|---|---|
| `_mm_movemask_epi8` | 0.84 -> 1.49 ns | 1.53 -> 1.97 ns | 12.47 -> 16.47 ns |

## Testing

No new test framework. `tests/impl.h` already lists all 14
`cmpistr*`/`cmpestr*` entry points and `tests/impl.cpp` references
`_SIDD_CMP_EQUAL_ANY` 30 times; `check-sve` reuses that harness.

```
make CROSS_COMPILE=aarch64-linux-gnu- check
    525 / 49 / 131 / 20 passed, 0 failed

make CROSS_COMPILE=aarch64-linux-gnu- check-sve
    vector length 16, 32 and 64 bytes: 525 / 49 / 131 / 20 passed, 0 failed each
```

On both devices, the same suites built with `SVE=1 FEATURE=sve2`: 725 tests,
0 failures on the Tensor G3 under Android 15 and again on the Tensor G5 under
Android 16. `objdump` confirms 6 `MATCH` instructions in the executed
`tests/main`.

Further checks:

- Preprocessor output with `SSE2NEON_ENABLE_SVE` unset is identical to current
  `master`; the `sse2neon.h` diff is 122 insertions and no deletions.
- GCC 13 and Clang 19, C and C++, project warning flags: no warnings from the
  new code. The Clang build also passes 525/525 on the device.
- `make check-macros`: 0 errors, the same 5 pre-existing warnings as `master`.
- `STRICT_ALIASING=1` with SVE2 enabled: passes, no new warnings from
  `sse2neon.h`.
- `make indent` (clang-format 22) touches only the new code and is idempotent.

## Not covered

- `SANITIZE=undefined` could not be run here: linking libubsan statically with
  the GNU cross toolchain fails on `__dynamic_cast`, identically on unmodified
  `master`. CI runs sanitizers on native Arm runners and is unaffected.
- `perf-tier.md` is not regenerated. The report generated from this branch is
  byte-identical to the one generated from `master`, so the change does not
  affect the tier analysis. The committed file is already out of sync on
  `master` in this environment by 289 lines, and regenerating it would add an
  unrelated diff produced by a possibly different libclang. It can be included
  if preferred.
- Neoverse is untested. The measurements cover six phone microarchitectures
  across two generations: Cortex-X3, A715, A510, X4, A725 and A520.
- MSVC defines no `__ARM_FEATURE_SVE2`, so it takes the NEON path by
  construction rather than by test.

## Question

Is a per-intrinsic, opt-in progression the direction you want for #537? The
alternative reading of that issue, porting genarchbench's `sse2sve.h` wholesale,
is a considerably larger change whose provenance and licensing would need
settling first. No code from it was used here.

If this direction is acceptable, the natural next step is a dedicated
`_mm_cmpistri` path: `MATCH` followed by `BRKB` and `CNTP` yields the index in
three instructions, since `CNTP` produces it directly and `BRKB` gives the
"no match returns 16" behaviour without forming a bitmask. That requires
modifying `SSE2NEON_COMP_AGG` and is deliberately excluded here.

### DIFF
--- a/.github/workflows/benchmark-linux-a64.yml
+++ b/.github/workflows/benchmark-linux-a64.yml
@@ -56,3 +56,92 @@ jobs:
           alert-threshold: '200%'
           comment-on-alert: true
           fail-on-alert: true
+
+  # NEON against SVE2 for the one relation SVE2 MATCH computes directly.
+  # Both numbers come from the same source built twice on the same runner, so
+  # the comparison is not across machines or across thermal states.
+  cmpistr-sve2:
+    name: Compare NEON and SVE2 for EQUAL_ANY on Linux A64
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          # scipy and numpy are what benchmark/tools/compare.py runs its
+          # U-test with; without them it refuses to report.
+          sudo apt-get install -y g++ cmake python3-numpy python3-scipy
+
+      - name: Require SVE2 on the runner
+        run: |
+          grep -m1 '^Features' /proc/cpuinfo
+          if ! grep -qw sve2 /proc/cpuinfo; then
+            echo "ERROR: runner reports no SVE2; both builds would be NEON"
+            exit 1
+          fi
+
+      - name: Cache Benchmark library
+        uses: actions/cache@v4
+        with:
+          path: benchmark
+          key: ${{ runner.os }}-googlebenchmark-a64-v1.8.3
+
+      - name: Build Google Benchmark
+        run: |
+          if [ ! -d "benchmark/build" ]; then
+            git clone --branch v1.8.3 --depth 1 https://github.com/google/benchmark.git
+            cd benchmark
+            cmake -E make_directory "build"
+            cmake -E chdir "build" cmake -DCMAKE_BUILD_TYPE=Release -DBENCHMARK_ENABLE_TESTING=OFF -DBENCHMARK_ENABLE_LTO=true ..
+            cmake --build "build" --config Release
+          fi
+
+      # clean between the two, because the flags change and the executable's
+      # timestamp does not.
+      - name: Benchmark the NEON path
+        env:
+          LD_LIBRARY_PATH: ./benchmark/build/src
+        run: |
+          make clean
+          make bench-cmpistr CXX=g++ \
+            BENCHMARK_CXXFLAGS="-I./benchmark/include" \
+            BENCHMARK_LDFLAGS="-L./benchmark/build/src -lbenchmark -lpthread -lrt" \
+            BENCH_ARGS="--benchmark_format=json --benchmark_out=$PWD/neon.json"
+
+      - name: Benchmark the SVE2 path
+        env:
+          LD_LIBRARY_PATH: ./benchmark/build/src
+        run: |
+          make clean
+          make bench-cmpistr SVE=1 FEATURE=sve2 CXX=g++ \
+            BENCHMARK_CXXFLAGS="-I./benchmark/include" \
+            BENCHMARK_LDFLAGS="-L./benchmark/build/src -lbenchmark -lpthread -lrt" \
+            BENCH_ARGS="--benchmark_format=json --benchmark_out=$PWD/sve2.json"
+          count=$(objdump -d tests/bench_cmpistr | grep -cE '\bmatch[[:space:]]+p[0-9]' || true)
+          echo "MATCH instructions in tests/bench_cmpistr: $count"
+          if [ "$count" -eq 0 ]; then
+            echo "ERROR: no MATCH instruction; this run measured the NEON path twice"
+            exit 1
+          fi
+
+      - name: Report the comparison
+        run: |
+          python3 benchmark/tools/compare.py --no-color benchmarks neon.json sve2.json | tee compare.txt
+          {
+            echo '### `_SIDD_CMP_EQUAL_ANY`: NEON vs SVE2'
+            echo
+            echo 'Same source, same runner, built twice. Negative means SVE2 is faster.'
+            echo
+            echo '```'
+            cat compare.txt
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload benchmark JSON
+        uses: actions/upload-artifact@v5
+        with:
+          name: cmpistr-neon-vs-sve2
+          path: |
+            neon.json
+            sve2.json
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,6 +106,69 @@ jobs:
             make FEATURE=${{ matrix.feature }} check
           fi
 
+  # SVE2 on the Neoverse N2 behind GitHub's arm64 runners. The paths guarded by
+  # SSE2NEON_ENABLE_SVE are compiled by nothing above this job, so without it
+  # the whole opt-in feature is untested.
+  host-arm-sve2:
+    runs-on: ubuntu-24.04-arm
+    strategy:
+      matrix:
+        cxx_compiler: [g++, clang++-18]
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v6
+      - name: install clang
+        if: contains(matrix.cxx_compiler, 'clang')
+        run: |
+          # clang-18 is in 'universe' repository on Ubuntu 24.04
+          sudo add-apt-repository -y -n universe
+          sudo apt-get -o Acquire::Retries=5 update -q -y
+          sudo apt-get -o Acquire::Retries=5 install -q -y --no-install-recommends "${CXX_COMPILER/clang++/clang}"
+        env:
+          CXX_COMPILER: ${{ matrix.cxx_compiler }}
+      # A runner without SVE2 builds the NEON paths and passes, which is the
+      # one result this job must never report as success.
+      - name: require SVE2 on the runner
+        run: |
+          grep -m1 '^Features' /proc/cpuinfo
+          if ! grep -qw sve2 /proc/cpuinfo; then
+            echo "ERROR: runner reports no SVE2; this job would test the NEON paths instead"
+            exit 1
+          fi
+      - name: build and test
+        env:
+          CXX: ${{ matrix.cxx_compiler }}
+        run: make SVE=1 FEATURE=sve2 check
+      # MATCH writes a predicate register, which no NEON encoding does, so its
+      # presence separates a real SVE2 build from one where the macro was
+      # defined and the paths gated out anyway.
+      - name: require MATCH in the test binary
+        run: |
+          count=$(objdump -d tests/main | grep -cE '\bmatch[[:space:]]+p[0-9]' || true)
+          echo "MATCH instructions in tests/main: $count"
+          if [ "$count" -eq 0 ]; then
+            echo "ERROR: no MATCH instruction; the SVE2 path was not compiled in"
+            exit 1
+          fi
+
+  # The vector-length sweep. GitHub's arm64 runners are 128-bit, so QEMU is the
+  # only way to run this code at 256 and 512 bits, which is what the sizeless
+  # SVE types in sse2neon.h claim to support. EXEC_WRAPPER is set only when
+  # CROSS_COMPILE is, so this cross-compiles from x86 rather than running on the
+  # arm runner.
+  host-arm-sve2-vl:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v6
+      - name: install cross toolchain and QEMU
+        run: |
+          sudo apt-get -o Acquire::Retries=5 update -q -y
+          sudo apt-get -o Acquire::Retries=5 install -q -y --no-install-recommends \
+            gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
+      - name: run the suite at every vector length
+        run: make CROSS_COMPILE=aarch64-linux-gnu- check-sve
+
   # ARMv7 requires QEMU emulation - minimize job count
   host-arm-emulated:
     runs-on: ubuntu-24.04
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ ifdef ARCH_CFLAGS_IS_SET
 endif
 
 COMMA := ,
+EMPTY :=
+SPACE := $(EMPTY) $(EMPTY)
 
 FEATURE ?=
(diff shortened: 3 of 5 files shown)


alanhc:
Thanks. Three of the four are real and are fixed in the updated `Makefile` commit. The fourth rests on a version claim that does not hold, and the correction changes the conclusion.

**`Makefile:75` — `SVE=0` was not honoured.** Correct. The guard now tests for the exact value:

```make
SVE ?=
SVE_FLAGS =
ifeq ($(SVE),1)
SVE_FLAGS = -DSSE2NEON_ENABLE_SVE=1
else ifneq ($(SVE),)
ifneq ($(SVE),0)
$(error Unsupported SVE value '$(SVE)'; use SVE=1 to enable, SVE=0 or unset to disable)
endif
endif
```

`SVE=0` and an unset `SVE` both keep the NEON paths. Any other value is rejected outright rather than silently enabling the macro, since a silent `SVE=sve2` is the same class of surprise as the one reported.

**`Makefile:107` — the fuzzer did not reach the SVE paths.** Correct, for two reasons: `FUZZ_CXXFLAGS` never carried `SVE_FLAGS`, and on AArch64 `FUZZ_ARCH_FLAGS` replaces `ARCH_CFLAGS` wholesale, which also discarded `FEATURE`. Both are folded back in, so `make SVE=1 FEATURE=sve2 fuzz` now builds with `-march=armv8-a+fp+simd+crypto+crc+sve2 -DSSE2NEON_ENABLE_SVE=1`. `COMMA` moves out of the `FEATURE` conditional so both sites can use it.

**`Makefile:238` — `check-sve` destroyed the build before failing.** Correct. The target now rejects non-AArch64 targets first, in the `@echo ERROR` / `@exit 1` shape already used by `generate-golden`:

```
$ make processor=x86_64 check-sve
ERROR: check-sve requires an AArch64 target, but the target is 'x86_64'
Use: make CROSS_COMPILE=aarch64-linux-gnu- check-sve
```

No `clean` runs.

**`sse2neon.h:9712` — the NEON-to-SVE bridge.** Right in principle, and named as a follow-up in the PR description. The stated availability is wrong, though:

| | first release |
|---|---|
| GCC, bridge intrinsics (`c535360788e1`, 2023-12-13) | **14** |
| GCC, `__ARM_NEON_SVE_BRIDGE` (`43530bc40b1d`, 2024-06-06) | **14.2**, after 14.1 on 2024-05-07 |
| Clang, `arm_neon_sve_bridge.h` actually installed (`8cb9e3c3ce1e`, 2022-11-03) | **16** |
| Clang, `__ARM_NEON_SVE_BRIDGE` (`0adab6b1af5b`, 2024-12-04) | **20** |

GCC 13 branched in April 2023 and predates the intrinsics by eight months, so it does not have them. Clang carried the header in-tree from December 2021 but did not install it until the 2022 commit above, so Clang 14 and 15 cannot include it either.

Clang 19 does have the header, but it is also the case that breaks the proposed gate: `__ARM_NEON_SVE_BRIDGE` is not defined before Clang 20, so `#ifdef __ARM_NEON_SVE_BRIDGE` would silently fall back to the memory path there. A gate that works would need GCC >= 14.2 and Clang >= 20 for the feature macro, or explicit version comparisons for GCC >= 14 and Clang >= 16.

README lists GCC 10+ and Clang 11+ as the minimum supported compilers, so either form adds a version-conditional path that most supported compilers would not take. That is why the memory round trip is deliberate here and the bridge is left as a separate change.

alanhc:
Correction to my own comment above, before anyone spends time on it.

I claimed `__ARM_NEON_SVE_BRIDGE` is not defined before Clang 20. That is wrong. `0adab6b1af5b` is titled "Define `__ARM_NEON_SVE_BRIDGE` **unconditionally**" — it removed a condition rather than introducing the macro. Before it, the macro was already defined whenever NEON and SVE were both enabled, which is exactly the configuration an SVE build of this header uses. Checked on Clang 19:

```
-march=armv8-a+simd  ->  not defined
-march=armv8-a+sve   ->  1
-march=armv8-a+sve2  ->  1
```

So `#ifdef __ARM_NEON_SVE_BRIDGE` would fire correctly there rather than silently falling back. I read the commit date and inferred the release instead of reading the title.

The conclusion is unchanged, and it never depended on that detail. GCC 13 genuinely lacks the header, which I verified locally, and README lists GCC 10+ and Clang 11+ as supported, so a bridge path still needs a version-conditional that most supported compilers would not take. That is why the memory round trip stays here and the bridge remains a follow-up.

jserv:
Can you integrate SVE2-aware targets in GitHub Actions, so that we can determine the effectiveness?

alanhc:
Added in three commits: `f080d4a` an EQUAL_ANY benchmark, `abe7324` the SVE2 CI jobs, `338661b` the NEON/SVE2 comparison.

The runners make this measurable rather than emulated. GitHub's `ubuntu-24.04-arm` images are Cobalt 100, Neoverse N2, and report SVE2:

```
Features : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid
asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm uscat
ilrcpc flagm sb paca pacg dcpodp sve2 sveaes svebitperm svesha3 svesm4 flagm2
frint svei8mm svebf16 i8mm bf16
```

`host-arm-sve2` builds and runs the suite with `SVE=1 FEATURE=sve2` under both g++ and clang++-18. It asserts twice rather than once, because the failure that matters here is not a test failure: a runner without SVE2 compiles the NEON paths and reports green for code it never ran. So the job refuses to start unless `/proc/cpuinfo` lists `sve2`, and afterwards requires a MATCH instruction in `tests/main`. MATCH writes a predicate register, which no NEON encoding does. g++ leaves 6 and clang-18 leaves 2, so the assertion is doing work across both.

`host-arm-sve2-vl` runs `check-sve` at 16, 32 and 64 byte vector lengths. The runners are 128-bit, so this is the only place the sizeless SVE types see another length. `EXEC_WRAPPER` is set only when `CROSS_COMPILE` is, so it cross-compiles from x86 under qemu-user instead of running on the arm runner.

`tests/bench_cmpistr.cpp` follows `tests/bench_movemask.cpp`: throughput over 1024 haystacks split by where the first set member falls, a dependent chain for latency, and strcspn for the in-context case. Both element sizes, and a set null-terminated at five characters, which is both what a delimiter set looks like and the case the SVE2 path pads. The comparison job builds that one source twice on one runner and reports `benchmark/tools/compare.py` into the step summary, with the raw JSON as an artifact. `make clean` runs between the builds because the flags change and the executable's timestamp does not, and the SVE2 build is required to contain MATCH so the job cannot measure the NEON path twice and report a difference of zero.

Negative is SVE2 faster:

```
BM_Throughput_Index_NoMatch      -0.6501       16 ->    6 ns
BM_Throughput_Index_MatchLate    -0.6501       16 ->    6
BM_Throughput_Index_MatchEarly   -0.6502       16 ->    6
BM_Throughput_Index_Random       -0.6501       16 ->    6
BM_Throughput_Mask_Random        -0.6493       17 ->    6
BM_Throughput_Index_Word_Random  -0.4372        8 ->    5
BM_Latency_Index                 -0.4365       23 ->   13
BM_Latency_Mask                  -0.4590       21 ->   11
BM_Strcspn_FoundAt2048           -0.6772     2058 ->  664
BM_Strcspn_NotFound              -0.6803     4066 -> 1300
BM_Strcspn_FoundAt0              -0.6566       20 ->    7
OVERALL_GEOMEAN                  -0.6098
```

2.86x on the byte form, 1.78x on the word form, 2.56x geometric mean. That is a third microarchitecture, measured by a machine nobody here owns, landing in the same range as the six Cortex cores in the description: 3.07x on Cortex-X3 and 3.40x on Cortex-A715 for the byte form, 1.79x and 1.95x for the word form. The two in-order little cores remain the outliers at 1.4-1.5x.

Cost: 22s for the g++ job, 30s for clang, 145s for the three-length sweep.

Two limits. The benchmark measures a 128-bit implementation only, since that is what the runners are; the emulated lengths establish correctness and their timings mean nothing. And the numbers cover `_SIDD_CMP_EQUAL_ANY` through `_mm_cmpistri` and `_mm_cmpistrm`, which is the scope of this branch.

PR #774, opened 2026-08-21
Drop the <fenv.h> dependency to stop leaking FE_* into includers

### LINKED ISSUE #767: FE_XXX macro redefinition warnings when compiling with streflop
Hi folks, when we try and use both sse2neon and [streflop](https://github.com/abma/streflop/tree/master) together in over at https://github.com/beyond-all-reason/RecoilEngine/, streflop emits a ton of warnings:

```
/build/src/rts/lib/streflop/FPUSettings.h:105:2: warning: #warning STREFLOP: FE_XXX flags were already
  defined and will be redefined! Check you do not use the system libm. [-Wcpp]
    105 | #warning STREFLOP: FE_XXX flags were already defined and will be redefined! Check you do not use
  the system libm.
        |  ^~~~~~~
```

From what I can tell this is because sse2neon pulls in `#include <fenv.h>` (which defines FE_INVALID, FE_TONEAREST, … as macros) which then leaks out to any places that include sse2neon. I don't think these are necessary outside of the sse2neon code itself?

Those conflict with streflop which _intentionally_ defines its own FE_XXX macros (the whole point of streflop is to replace the system libm with bit-for-bit reproducible math across platforms).

My workaround is to [undef them](https://github.com/beyond-all-reason/RecoilEngine/pull/2999/changes#diff-c7f31fd1772104591b9044fb77abd5ff8532e8c4e71fb51fae9c813b09878139R6-R18) from sse2neon in the build.

### Questions
1. is undef'ing them from sse2neon the right solution?
2. If so, should these be undef'd in this repo?

alanhc:
Fixes #767.

sse2neon.h includes <fenv.h>, so every translation unit that includes sse2neon.h also gets FE_INVALID, FE_TONEAREST and the rest. That breaks projects which deliberately define their own: RecoilEngine reports streflop emitting "FE_XXX flags were already defined and will be redefined" for every file, because streflop replaces the system FE_* macros on purpose.

The include exists only for _MM_GET_ROUNDING_MODE and _MM_SET_ROUNDING_MODE, which is more indirection than the job needs. The rounding mode lives in FPCR.RMode, bits [23:22], and this header already has everything required to reach it: _sse2neon_get_fpcr()/_sse2neon_set_fpcr() and the fpcr_bitfield struct, whose bit22 and bit23 members are exactly that field. The sibling _MM_GET_FLUSH_ZERO_MODE and _MM_SET_FLUSH_ZERO_MODE already manipulate FPCR this way, so the two functions now follow the same shape. On AArch32 they use the same vmrs/vmsr FPSCR access as those siblings.

Mapping, per the Arm ARM: 0b00 nearest, 0b01 toward +infinity, 0b10 toward -infinity, 0b11 toward zero. Invalid input to _MM_SET_ROUNDING_MODE still falls back to truncation, matching the previous behaviour.

Besides fixing the reported problem this removes a libm dependency and replaces a libc call with a single MRS.

Verified equivalent rather than assumed: a test that sets each mode through _MM_SET_ROUNDING_MODE and then reads fegetround() finds them in agreement for all four modes, on QEMU and on a Tensor G5 device, and 1.0f/3.0f differs between the round-down and round-up settings, so the FPCR write demonstrably takes effect.

No new tests. The existing harness already covers this well: test_mm_set_rounding_mode checks that _mm_round_ps(_MM_FROUND_CUR_DIRECTION) agrees with the explicit-mode form in each of the four modes, and nine _mm_cvt* tests set rounding modes and validate their results. make CROSS_COMPILE=aarch64-linux-gnu- check passes 525/49/131/20 with zero failures, and again with STRICT_ALIASING=1. The AArch32 path was compile-checked and confirmed to emit vmrs/vmsr.

---
## Summary by cubic
Removes the `<fenv.h>` dependency from `sse2neon.h` by reading/writing FPCR/FPSCR directly for rounding mode. This stops leaking `FE_*` into includers and fixes redefinition conflicts (Fixes #767) without changing observable behavior.

- `_MM_GET_ROUNDING_MODE`/`_MM_SET_ROUNDING_MODE` now map FPCR.RMode bits [23:22] (00 nearest, 01 up, 10 down, 11 toward zero); invalid input still truncates.
- Unifies `_sse2neon_get_fpcr`/`_sse2neon_set_fpcr` across AArch64/AArch32 via a new `_sse2neon_fpcr_t`, removing per-callsite `#if` and using `mrs/msr` or `vmrs/vmsr` as appropriate.
- Updates flush-zero and denormals-zero helpers to use the unified accessors; fixes an AArch32 write that previously passed a union instead of its value.
- Drops `fegetround`/`fesetround` and any `libm`/`libc` calls; projects redefining `FE_*` (e.g., streflop) no longer see warnings.
- No migration required; existing tests already cover rounding-mode behavior.

<sup>Written for commit 9197ec24f051736d740c3579746be65695b18260. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/DLTcollab/sse2neon/pull/774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sse2neon.h
+++ b/sse2neon.h
@@ -413,7 +413,6 @@
 #endif
 #endif /* SSE2NEON_ARM64EC */
 
-#include <fenv.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
@@ -2708,20 +2707,33 @@ FORCE_INLINE unsigned int _sse2neon_mm_get_flush_zero_mode(void)
 // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_MM_GET_ROUNDING_MODE
 FORCE_INLINE unsigned int _MM_GET_ROUNDING_MODE(void)
 {
-    const int mask = FE_TONEAREST | FE_DOWNWARD | FE_UPWARD | FE_TOWARDZERO;
-    switch (fegetround() & mask) {
-    case FE_TONEAREST:
+    // Read FPCR/FPSCR directly rather than going through <fenv.h>, so that the
+    // header does not leak the FE_* macros into everything that includes it.
+    union {
+        fpcr_bitfield field;
+#if SSE2NEON_ARCH_AARCH64
+        uint64_t value;
+#else
+        uint32_t value;
+#endif
+    } r;
+
+#if SSE2NEON_ARCH_AARCH64
+    r.value = _sse2neon_get_fpcr();
+#else
+    __asm__ __volatile__("vmrs %0, FPSCR" : "=r"(r.value)); /* read */
+#endif
+
+    // FPCR.RMode occupies bits [23:22]: 0b00 nearest, 0b01 toward +infinity,
+    // 0b10 toward -infinity, 0b11 toward zero.
+    switch ((r.field.bit23 << 1) | r.field.bit22) {
+    case 0:
         return _MM_ROUND_NEAREST;
-    case FE_DOWNWARD:
-        return _MM_ROUND_DOWN;
-    case FE_UPWARD:
+    case 1:
         return _MM_ROUND_UP;
-    case FE_TOWARDZERO:
-        return _MM_ROUND_TOWARD_ZERO;
+    case 2:
+        return _MM_ROUND_DOWN;
     default:
-        // fegetround() must return _MM_ROUND_NEAREST, _MM_ROUND_DOWN,
-        // _MM_ROUND_UP, _MM_ROUND_TOWARD_ZERO on success. all the other error
-        // cases we treat them as FE_TOWARDZERO (truncate).
         return _MM_ROUND_TOWARD_ZERO;
     }
 }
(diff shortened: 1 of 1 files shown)


sse2neon.h
```diff
@@ -3342,26 +3354,51 @@ FORCE_INLINE __m128 _mm_set_ps1(float _w)
...
+        r.field.bit23 = 1;
+        break;
     }
-    fesetround(rounding);
+
+#if SSE2NEON_ARCH_AARCH64
+    _sse2neon_set_fpcr(r.value);
+#else
+    __asm__ __volatile__("vmsr FPSCR, %0" ::"r"(r)); /* write */
+#endif
```
jserv: How about unifying `_sse2neon_set_fpcr` for both cases?

alanhc: Done in 9197ec2. It turned out the same `#if` was repeated at nine call sites across six functions, so I pushed the split down into the accessors and added a `_sse2neon_fpcr_t` typedef for the register width. That also collapsed the six copies of the `union { fpcr_bitfield field; <width> value; } r;` declaration, which each carried the same `#if` inside — net 41 lines removed.

I touched all six functions rather than only the two this PR added. Leaving `_MM_GET/SET_FLUSH_ZERO_MODE` and the denormals pair in the old shape would have made the file inconsistent with itself, but it does widen a bug fix into a small refactor.

The AArch32 write sites were passing the union rather than its value:

```c
__asm__ __volatile__("vmsr FPSCR, %0" ::"r"(r)); /* r is the union */
```

Going through the accessor passes `r.value`, so those three are gone. An AArch32 build now emits six `vmrs` and three `vmsr`, matching the six reads and three writes, and objdump shows the write reaching `vmsr` through a plain register (`bfi r1, r0, #24, #1` then `vmsr fpscr, r1`). Same 34 pre-existing warnings as master, none added.

One caveat on the green CI: it was for 31c24f7. This commit changes the accessors themselves, including the `_ReadStatusReg(ARM64_FPCR)` path the MSVC jobs exercise, so those results do not carry over and the workflows need another approval to be meaningful.

jserv:
I defer to @Cuda-Chen for confirmation.

jserv:
Thank @alanhc for contributing!

PR #775, opened 2026-09-13
Restore hardware AES path for MSVC ARM64

### LINKED ISSUE #760: The change for _M_ARM64EC has broken _M_ARM64
line 10270 reads "#if !defined(__ARM_FEATURE_CRYPTO) || SSE2NEON_ARM64EC || defined(_M_ARM64EC)" which in my MSVC resolves to true because __ARM_FEATURE_CRYPTO is not defined. It used to be "#if !defined(__ARM_FEATURE_CRYPTO) && (!defined(_M_ARM64) || defined(__clang__))" which resolves to false. 

I now have it as "#if !defined(__ARM_FEATURE_CRYPTO) && ((!defined(_M_ARM64) || defined(__clang__)) || SSE2NEON_ARM64EC || defined(_M_ARM64EC))" which works for me, but I'm not sure it's right for anyone else.

yuefdev:
Fixes #760.

## Problem

On MSVC targeting native ARM64, the AES intrinsics (`_mm_aesenc_si128`, `_mm_aeskeygenassist_si128`, ...) have fallen back to the table-based software implementation instead of using AESE/AESD/AESMC.

MSVC never defines `__ARM_FEATURE_CRYPTO`. b390e8a (MSVC support) handled this in the guard of the software fallback:

```c
#if !defined(__ARM_FEATURE_CRYPTO) && (!defined(_M_ARM64) || defined(__clang__))
```

70eb0de (ARM64EC CI coverage) rewrote that guard as:

```c
#if !defined(__ARM_FEATURE_CRYPTO) || SSE2NEON_ARM64EC || defined(_M_ARM64EC)
```

The ARM64EC part is intended, but the `_M_ARM64 && !__clang__` exception was dropped along with it. For MSVC ARM64 the guard is now always true, so:

- the hardware AES path is never compiled on MSVC ARM64, although `_sse2neon_vmull_p64` still assumes crypto on the same target ("every single ARM64 processor supported by WoA has crypto extensions");
- the MSVC-specific `__n128` branch in `_mm_aeskeygenassist_si128` is dead code.

## Fix

Restore the exception and keep ARM64EC on the software path:

```c
#if (!defined(__ARM_FEATURE_CRYPTO) &&              \
     (!defined(_M_ARM64) || defined(__clang__))) || \
    SSE2NEON_ARM64EC || defined(_M_ARM64EC)
```

| Target | Before | After |
|---|---|---|
| GCC/Clang AArch64 with `+crypto` | HW | HW |
| GCC/Clang without crypto | SW | SW |
| MSVC ARM64 | **SW** (regression) | HW |
| clang-cl ARM64 without crypto | SW | SW |
| MSVC ARM64EC | SW | SW |

## Testing

- `clang-format` 20.1.8 produces no diff.
- CI run [34761633726](https://github.com/DLTcollab/sse2neon/actions/runs/34761633726): all 28 jobs passed. `test-win-msvc` (MSVC ARM64 Release under wine-arm64) reports `Passed: 525, Failed: 0`, including all six AES tests.
- Heads-up: the MSVC `__n128` branch of `_mm_aeskeygenassist_si128` was rewritten (brace-init → field assignment) after it stopped being compiled on MSVC ARM64. That run was the first time it had been compiled and tested since, and it passes.
- I don't have Windows-on-Arm hardware, so I haven't run it natively.
- After that run I amended the commit to reword the header comment (per the review bot). It's a comment-only change.

AI disclosure: I tracked down the regression and wrote this patch with help from Claude Code. I reviewed the change myself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jserv:
Thank @yuefdev for contributing!

